PR #38, opened 2017-07-27
Improve controller security

tompizmor:
This PR will make the controller image to run as non-root user. The file system will also be read-only.

```
$ kubectl exec -it -n kube-system sealed-secrets-controller-222321944-q4npb -- sh
/ $ id
uid=1001 gid=0(root)
/ $ touch /tmp/test
touch: /tmp/test: Read-only file system
```

### DIFF
--- a/controller-norbac.jsonnet
+++ b/controller-norbac.jsonnet
@@ -32,6 +32,8 @@ local controllerContainer =
   container.command(["controller"]) +
   container.livenessProbe(controllerProbe) +
   container.readinessProbe(controllerProbe) +
+  container.mixin.securityContext.readOnlyRootFilesystem(true) +
+  container.mixin.securityContext.runAsUser(1001) +
   container.helpers.namedPort("http", controllerPort);
 
 local labels = {name: "sealed-secrets-controller"};
--- a/controller-norbac.jsonnet
+++ b/controller-norbac.jsonnet
@@ -32,8 +32,10 @@ local controllerContainer =
   container.command(["controller"]) +
   container.livenessProbe(controllerProbe) +
   container.readinessProbe(controllerProbe) +
+  container.securityContext(k.core.v1.podSecurityContext.default()) +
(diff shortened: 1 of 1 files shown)


anguslees:
Travis is telling us:
> Error: Error reading controller.jsonnet: RUNTIME ERROR: Field does not exist: runAsUser

controller-norbac.jsonnet
```diff
@@ -32,6 +32,8 @@ local controllerContainer =
   container.command(["controller"]) +
   container.livenessProbe(controllerProbe) +
   container.readinessProbe(controllerProbe) +
+  container.mixin.securityContext.readOnlyRootFilesystem(true) +
+  container.mixin.securityContext.runAsUser(1001) +
```
anguslees: Urgh. ksonnet-lib beta.1 has a bunch of bugs.  You've typed two lines and you've encountered two of the bugs ;)

- the base securityContext object isn't created implicitly by `container.default()`
- it seems it expects `runAsNonRoot` to take an object, not a simple integer.

Here are the workarounds:
```javascript
  container.securityContext(k.core.v1.podSecurityContext.default()) +
  container.mixin.securityContext.readOnlyRootFilesystem(true) +
  container.mixin.securityContext.runAsNonRoot(true) +
  {securityContext+: {runAsUser: 1001}} +
```

(I've also included `container.mixin.securityContext.runAsNonRoot(true)` because we may as well assert that runAsUser is doing what we think)

(I'll move to ksonnet-lib beta.2 RSN, but separately to this PR)

tompizmor: Do you know why I couldn't reproduce this locally? This is the command I run to get the .ymls

```bash
$ kubecfg show -o yaml controller-norbac.jsonnet -J /Users/tompizmor/Desktop/github-repos/ksonnet-lib-beta.1
```

tompizmor: BTW, good point including the `runAsNonRoot(true)`. I didn't know it exists

PR #41, opened 2017-08-10
Add oidc and gcp auth providers

bakins:
Register the auth providers for oidc and gcp. This allows `kubeseal` to be used with clusters that use those authentication methods.


cmd/kubeseal/main.go
```diff
@@ -22,6 +22,10 @@ import (
 
 	// Register v1.Secret type
 	_ "k8s.io/client-go/pkg/api/install"
+
+	// Register Auth providers
+	_ "k8s.io/client-go/plugin/pkg/client/auth/gcp"
```
anguslees: Instead of listing them each individually, just import `k8s.io/client-go/plugin/pkg/client/auth`

bakins: I didn't even notice that. I'll change to use that. Thanks for the quick feedback!

bakins:
@anguslees I made the change. I can squash the commits - not sure if you do that on merge or not.

PR #57, opened 2017-09-21
Add an openapi schema declaration to the CRD

anguslees:
This prepares us for the associated alpha feature in k8s 1.8.  See
https://github.com/kubernetes/kubernetes.github.io/pull/5290 for docs.


mmikulicic:
nice!

PR #58, opened 2017-09-26
Release download link change

migmartri:
Following the installation readme I noticed that the provided URL was returning 404 even after the GOOS GOARCH expansion.

This patch fixes the link to what I think is the proper release URL.


anguslees:
I made this change intentionally (in fb7628db56118b946244fca4d7fce2a561b91dd5), intending for this to make it easier to download the binary without renaming it.  When I went to cut the release, I found this didn't work as I'd hoped and needed to be reverted :(

I ended up fixing this with my own PR #60, which I made by mechanically reverting the relevant pieces of that earlier change.  Thanks.

PR #61, opened 2017-10-27
Allow setting custom label for encrypt/decrypt

bakins:
This would allow a secret to be used in multiple namespaces.  We spin up test environments in different generated namespaces and this allows us to reuse the same secret without having to re-seal each time by setting an annotation on the source secret.

I can add docs if you think this is useful.

### DIFF
--- a/apis/v1alpha1/sealedsecret.go
+++ b/apis/v1alpha1/sealedsecret.go
@@ -81,8 +81,8 @@ func (sl *SealedSecretList) GetListMeta() metav1.List {
 }
 
 func labelFor(o metav1.Object) ([]byte, bool) {
-	label, ok := o.GetAnnotations()[SealedSecretLabelAnnotation]
-	if ok && label != "" {
+	label := o.GetAnnotations()[SealedSecretLabelAnnotation]
+	if label != "" {
 		return []byte(label), true
 	}
 	label = fmt.Sprintf("%s/%s", o.GetNamespace(), o.GetName())
--- a/apis/v1alpha1/sealedsecret.go
+++ b/apis/v1alpha1/sealedsecret.go
@@ -25,9 +25,9 @@ const (
 	// SealedSecretPlural is the collection plural used with SealedSecret API
 	SealedSecretPlural = "sealedsecrets"
 
-	// SealedSecretLabelAnnotation is the name for the annotation for
-	// overriding the label used for encryption and decryption.
-	SealedSecretLabelAnnotation = "sealedsecrets.bitnami.com/label"
+	// SealedSecretClusterWideAnnotation is the name for the annotation for
+	// setting the secret to be availible cluster wide.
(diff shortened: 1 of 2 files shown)


apis/v1alpha1/sealedsecret.go
```diff
@@ -76,9 +80,13 @@ func (sl *SealedSecretList) GetListMeta() metav1.List {
 	return &sl.Metadata
 }
 
-func labelFor(o metav1.Object) []byte {
-	label := fmt.Sprintf("%s/%s", o.GetNamespace(), o.GetName())
-	return []byte(label)
+func labelFor(o metav1.Object) ([]byte, bool) {
+	label, ok := o.GetAnnotations()[SealedSecretLabelAnnotation]
+	if ok && label != "" {
```
anguslees: May as well drop the `ok` here, since we only want to know if it is a non-empty value:

```golang
label := o.GetAnnotations()[SealedSecretLabelAnnotation]
if label != "" {
  ...
}
```

bakins:
@anguslees "The feature implemented in this PR basically means that a SealedSecret (created with this new annotation) can be decrypted by any user of your cluster (with RBAC access to upload a SealedSecret), regardless of the namespace/name they're using within that cluster." - correct, this is my use case and my intention.  As we need the same secret in multiple namespaces, this is more straightforward than having the same secret encrypted for each namespace. Some of these namespaces are created dynamically for tests, etc.

This is "opt-in" - the secret must be created with the annotation. By default, behavior is the same as before.  This allows us to use this when needed, but most secrets are done the default way.

The label is just text - one could use the same value for all sealed secrets or use some descriptive text.

bakins:
@anguslees Are there any more changes you would like?

anguslees:
@bakins: thanks for your patience...

> Given that the label is visible on the unencrypted SealedSecret (with this PR) and independent of the decryption context, there isn't much point asking the user to provide a value at encryption-time - and indeed they might get the mistaken impression that this value somehow acts like a secret password to "unlock" the secret (which it isn't since it is publicly known!). I suggest that we should instead just change this new annotation to a simple boolean indicating a "cluster-wide" SealedSecret (and when true, we use a known label (ie: empty string) and skip the namespace/name check - basically exactly what this PR does now).

Specifically:

Change the annotation to `sealedsecrets.bitnami.com/cluster-wide=true`.

When sealing, if the annotation (on the Secret) is "true":
- use empty-string for the RSA-OAEP label
- propagate the annotation to the SealedSecret

When unsealing, if the annotation (on the SealedSecret) is "true"
- use empty-string for the RSA-OAEP label (this will verify that it was _sealed_ with an empty label)
- (the annotation should already be on the Secret, leave it there for informational purposes)

For bonus points, we could add a flag to `kubeseal` to set this annotation - that way we can keep the simple `kubectl create secret ... | kubeseal --cluster-wide` usage.  I'm ok if you want to leave that (and docs) to a later PR.

bakins:
@anguslees Changes made. kubeseal changes should probably be a separate PR.

PR #67, opened 2017-11-30
Encrypt secrets individually

### LINKED ISSUE #63: Encrypt each secret separately
Just found this great project today! It looks pretty much exactly what we were considering save for a few details. Would like to contribute some extras, just had a few questions first.

We were likely going to utilize [sops](https://github.com/mozilla/sops) for the main portion of it with a controller or initializer wrapped around it. In terms of differences, there's a few things I'd be interested to gain. Mostly encrypting each secret separately to make diffs meaningful and conflicts easier to resolve. KMS support and CLI editing for those with access to the key might be good too although likely separate issues.

sullerandras:
Resolves #63 
This is not backwards compatible. The old secrets will have to be re-encrypted.

Encrypting a secret like this:
```json
{
    "kind": "Secret",
    "apiVersion": "v1",
    "metadata": {
        "name": "mysecret2",
        "creationTimestamp": null
    },
    "data": {
        "bar": "cXdl",
        "foo": "dGhpc2lzbXlzZWNyZXQy\n"
    }
}
```

Will look like this:
```json
{
  "kind": "SealedSecret",
  "apiVersion": "bitnami.com/v1alpha1",
  "metadata": {
    "name": "mysecret2",
    "namespace": "default",
    "creationTimestamp": null
  },
  "spec": {
    "data": {
      "bar": "AgBpKIK/fq7bhbUPhYhelMksEdr8hjeUjTh6MGbp3rox/pJmZ+Zsu5vCMrvth+wHt9N2vcqSnmin1dFRqeMFRCOiMvJbpWxoCqhaXR66GtGIS8lMeCZ7t0QLp5n4JkW3UDQz7qLy5InQIe9prmKTVntYicNeap4Dtx5qw275t3jW9bLbjjFn0BXxqrlkYeK9iJC1fRPvnDtOIBTmwv5MfDdZTPI5nfj1j5ot6TnQALtzZM0iogvgcnwNXIDO90feaCkyoMaG44/EcjrsEAepBrwj8tT/QwZylaqtlK1J3NDVK03IcCx7FV92zUy2BrXyDycd0KGin6wjDqkO5u9gtQTURKmM5uMQgfwH5u7P+0ZjhU3yZQs/EvMvCCDr7voTRw4bPKZevTAEt9D12U3dDgviIgahAxVyXWUsrxQXcpzPHXiw47zWZAALh4gl+RqTwN4UMQjSvw5PwIv+PQ6Sx4T1Sirxh1N1Xby+xY+hBgZXyHtl7URTURmFvYF9NE52dzC4dURAMKgsN6p94u2uiin+rngN6Um+bSwh+d72K4aIgJaD3Fbq5L3HGXrktvesmQr8QtJGZEbVf7Z2+J9p/bDUSOQ/H08zWc5yJLCUeBo6QXnSW7hseKVNcLz+tJfWOp8xskkadOUu8OJqkUDacs5CbPHo3+nhGYttJfdaTPcPP9ZsDk3OaG4nEfco9NjoC71rWjE=",
      "foo": "AgABvIHwOTs7j+aNfqC08PypNOOmigoVhxpxNwbV8wap7VajbYyG3oqdECrWKEsad1L4w9/qdA9umX9pzVVSs+mW0BWJNRNiCbQ/ekEG/K8cEeorDmhCy252YjH1tE7nXTX/OxYTowXX45BLHQyKaKnN7lVjK24QfnSnQb4ITD1RQ7yGMuVkFnBL20JwTViImJfCoJnKfqnmMQRsDMYCBblmwBzq8LT5wdvQMwZSRGAUbKK/QU2A1BvRPUQRC6j73OI0jXQT+17HIQ9Sfpyr4oZte/JjGyozckpR2057fk8uErU3mLdH+qF2qtGx0siht2LUzxsEDs2Z752RSgw3QwAWCSlH12rIZ3WM+pMi5Cj3Q/RJgaULgwGIX58AiLbeNQcxbHWjbDgTIti/qYh3UVCATsOkM2NYa+8hloKoM0TpNMTbqXPJ38imPMdgq9LenKKBLIhkrqJSJ8TpZFKfrsjLGb1Ey+PEUBH0Z9vluJKVPrK+alc6o6hP0z0Z7EgDpIN6qmgs5BoTGshZ5WN4DjsEobBENtUL4uE+96//O06UtYJLehTAv80InhLkSib+em+V/bIS5uiKjcHXySEUNayzLyIo5s5x41OIym1qwWp4g+wDg3LRsS4fGMKECLUjN+h8FQpfqN5WTmBSwY9sk9spRkF/2nTMXU01g+agYhqmtscpKAhGs5xDM2DLI0RHkD1NsqI5//QLcAbey5VaOV0="
    }
  }
}
```

### DIFF
--- a/apis/v1alpha1/sealedsecret.go
+++ b/apis/v1alpha1/sealedsecret.go
@@ -13,6 +13,7 @@ import (
 	"io"
 
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
 	"k8s.io/apimachinery/pkg/runtime/schema"
 	runtimeserializer "k8s.io/apimachinery/pkg/runtime/serializer"
 	"k8s.io/client-go/pkg/api/v1"
@@ -36,7 +37,8 @@ var ErrTooShort = errors.New("SealedSecret data is too short")
 
 // SealedSecretSpec is the specification of a SealedSecret
 type SealedSecretSpec struct {
-	Data map[string][]byte `json:"data"`
+	Data          []byte            `json:"data"`
+	EncryptedData map[string][]byte `json:"encryptedData"`
 }
 
 // SealedSecret is the K8s representation of a "sealed Secret" - a
@@ -164,8 +166,54 @@ func hybridDecrypt(rnd io.Reader, privKey *rsa.PrivateKey, ciphertext, label []b
 	return plaintext, nil
 }
 
+// NewSealedSecretV1 creates a new SealedSecret object wrapping the
+// provided secret. This encrypts all the secrets into a single encrypted
+// blob and stores it in the `Data` attribute. Keeping this for backward
+// compatibility.
+func NewSealedSecretV1(codecs runtimeserializer.CodecFactory, pubKey *rsa.PublicKey, secret *v1.Secret) (*SealedSecret, error) {
+	info, ok := runtime.SerializerInfoForMediaType(codecs.SupportedMediaTypes(), runtime.ContentTypeJSON)
+	if !ok {
+		return nil, fmt.Errorf("binary can't serialize JSON")
+	}
+
+	if secret.GetNamespace() == "" {
+		return nil, fmt.Errorf("Secret must declare a namespace")
+	}
+
+	codec := codecs.EncoderForVersion(info.Serializer, v1.SchemeGroupVersion)
+	plaintext, err := runtime.Encode(codec, secret)
+	if err != nil {
+		return nil, err
+	}
+
+	// RSA-OAEP will fail to decrypt unless the same label is used
+	// during decryption.
+	label, clusterWide := labelFor(secret)
+
+	ciphertext, err := hybridEncrypt(rand.Reader, pubKey, plaintext, label)
+	if err != nil {
+		return nil, err
+	}
+
+	s := &SealedSecret{
+		Metadata: metav1.ObjectMeta{
+			Name:      secret.GetName(),
+			Namespace: secret.GetNamespace(),
+		},
+		Spec: SealedSecretSpec{
+			Data: ciphertext,
+		},
+	}
+
+	if clusterWide {
+		s.Metadata.Annotations = map[string]string{SealedSecretClusterWideAnnotation: "true"}
+	}
+	return s, nil
+}
+
 // NewSealedSecret creates a new SealedSecret object wrapping the
-// provided secret.
+// provided secret. This encrypts only the values of each secrets
+// individually, so secrets can be updated one by one.
 func NewSealedSecret(codecs runtimeserializer.CodecFactory, pubKey *rsa.PublicKey, secret *v1.Secret) (*SealedSecret, error) {
 	if secret.GetNamespace() == "" {
 		return nil, fmt.Errorf("Secret must declare a namespace")
@@ -177,7 +225,7 @@ func NewSealedSecret(codecs runtimeserializer.CodecFactory, pubKey *rsa.PublicKe
 			Namespace: secret.GetNamespace(),
 		},
 		Spec: SealedSecretSpec{
-			Data: map[string][]byte{},
+			EncryptedData: map[string][]byte{},
 		},
 	}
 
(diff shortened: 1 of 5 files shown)


bakins:
Could we use a label or annotation to denote what for the secret is? Perhaps use a different field than data? bump the API version?

We are currently using sealed-secrets heavily and it'd be painful to have to re-encrypt everything at the same time we did an upgrade.

anguslees:
@bakins: I hear you, and I'm also keen to make sure we can provide a smooth transition somehow.

We can't use the normal api version machinery unfortunately, since CRDs don't support more than one version at once - but I have a few other approaches in mind.

@sullerandras: The change looks great (thanks!).  I won't have time to give it a proper look for another ~2 weeks (after kubecon US) - but please don't misinterpret the delay as a lack of enthusiasm on my part :)

c-knowles:
@anguslees yes the CRD limitation is the one we hit and thought it better to get the conversation started on how to do this.

Currently one has to have access to the whole secret to encrypt a new version, so I think the way to provide backwards compatibility is to support both individual secrets as well as joint for a few releases. We could name it `encryptedData` or similar?

@sullerandras we should also update the `Details` section of the readme to be clear on the differences.

bakins:
It would be a hack, but could we try to unmarshal data as a map and if that fails unmarshal as a byte array? This would allow both types to be supported for a while. We could then remove support for the "old" way.

I haven't tried it, but perhaps directly using https://golang.org/pkg/encoding/json/#RawMessage ?

anguslees:
@bakins: That was actually my vague plan.

We can also migrate from the old form to the new form easily.  Use the introspection hack to detect the old byte-array form and re-encrypt it as the new form.  I don't want to do this, since that will probably confuse git-driven workflows - so open to other migration aids.  (Could make `kubeseal` convert an old-form SealedSecret, but that also requires `kubeseal` to have access to the private key which is scary - may as well just fetch the unencrypted Secret at that point and do a "regular" re-encrypt with `kubeseal`)

anguslees:
So what do we want to do for migration?

Options (off the top of my head):

1. Intuit old/new scheme from looking at the json structure and "just work".
2. Introduce a new field that adds our own schema version (since CRDs can't use apiVersion effectively).  Everyone will need to give `version: 2` (or similar) forever going forward.
3. Introduce a new `encryptedData` field with the new individual-keys behaviour, but doesn't overlap (at the schema level) with `data`.  Reject SealedSecrets that define both `encryptedData` _and_ `data`.  Everyone will use `encryptedData` going forward.
4. Automatically decrypt/re-encrypt SealedSecrets into the new form in the controller.  This will likely upset workflows that expect to read back something that looks like what they uploaded (kubectl apply, kubecfg, etc).

Votes?  Other suggestions?

I'm also implicitly assuming here that we're all ok with dropping support for the existing all-in-one scheme - the only open question is what the new form looks like, and how the transition is going to work.  If we want to keep support for _both_ all-in-one and individual-values schemes going forward, then that knocks out a few options from the list above.

anguslees:
Oh, and at _some_ point we should move off `bitnami.com/v1alpha1` probably to `bitnami.com/v1` or some other apiGroup.  That's going to be clumsy and disruptive, because CRDs don't support multiple versions for the same underlying resource.  We could possibly make that change at the same time as the schema transition - even though the combined result is going to cause a greater disturbance.

sullerandras:
@anguslees I tried to implement option 3, so now it should be backward compatible. Please have a look. I didn't update the readme as it was not talking about the low level implementation details.

sealedsecret-crd.jsonnet
```diff
@@ -30,6 +30,14 @@ local crd = {
                 type: "string",
                 pattern: "^[^A-Za-z0-9+/=]*$", // base64
               },
+              encryptedData: {
+                patternProperties: {
+                  "^.*$": {
+                    pattern: "^[^A-Za-z0-9+/=]*$", // base64
```
anguslees: This needs to be `"^[A-Za-z0-9+/=]*$"`  (see #71)

apis/v1alpha1/sealedsecret.go
```diff
@@ -37,7 +37,8 @@ var ErrTooShort = errors.New("SealedSecret data is too short")
 
 // SealedSecretSpec is the specification of a SealedSecret
 type SealedSecretSpec struct {
-	Data []byte `json:"data"`
+	Data          []byte            `json:"data"`
```
anguslees: We should add a "deprecated" comment above this, or similar.

sullerandras:
Closing in favor of https://github.com/bitnami/sealed-secrets/pull/73

PR #70, opened 2017-12-15
Big code refactor: client-go v5, and code-generator

### LINKED ISSUE #69: Upgrade k8s library?
Currently, we have a custom command (like kubeseal) that has a lot of copy pasted code from this repo but we'd like to just use this as a library. Unfortunately, I quickly ran into issues as we need a newer version of the k8s client. k8s client seems to change package layout every release. Any interest in upgrading the k8s client used by sealed-secrets? I'm happy to do the work to make it happen.

anguslees:
Huge code update, which should be an effective no-op.

- client-go release-5
- apimachinery, api, etc release-1.8
- Use code-generator to generate "standard" pkg/{apis,client}
- Change controller to use new (generated) informer
- Moved the small amount of actual crypto code into pkg/crypto

Fixes #69


anguslees:
I'm going to merge this.   Apologies for anyone who has outstanding PRs, since you're almost certainly going to have rebase conflicts to deal with :(

PR #74, opened 2018-01-10
move to bitnami-labs

sebgoa:
some of the files seem to be auto-generated 
I could not build because of the wrong import.

not sure what is the best/cleanest way to proceed.


anguslees:
`make generate` should regenerate the auto-generated files.  The generator tools are a bit fussy in the $GOROOT layout they expect, so you might need to fiddle a bit - in light of that, I'm happy just accepting a "manual" search/replace to get the build working again, and then fix the autogeneration pipeline again later.

What wouldn't build with the current PR?  travis-ci should be set up again now, so if you rebase/repush, you should get some travis logs that we can look at.

anguslees:
Superseded by #75

PR #81, opened 2018-02-20
Set controller command as image entrypoint

wmedlar:
This simple PR modifies the Dockerfile to use the `ENTRYPOINT` instruction instead of `CMD`, as the container's primary purpose is to run the controller. Using an entrypoint is more intuitive to an end-user as they can simply set runtime flags in the pod spec's args.


anguslees:
I have no strong opinion on this either way.  Setting ENTRYPOINT means it's harder to exec a debugging shell or otherwise use the container differently - otoh I agree it saves you from having to specify `container.command` in k8s.

PR #84, opened 2018-03-16
Omit empty spec.data, and improve integration test coverage

anguslees:
Improve integration test coverage to exercise both negative
paths (various types of invalid SealedSecrets), and new per-key
encryption.

Add `omitempty` to spec.data, fixing an issue when Spec.Data was
`nil` (yay for tests).

Also: bump version of kubecfg used to generate manifest YAML.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ os:
 
 env:
   global:
+    - PATH=$PATH:$GOPATH/bin
     - CONTROLLER_IMAGE_NAME=quay.io/bitnami/sealed-secrets-controller
     - CONTROLLER_IMAGE=${CONTROLLER_IMAGE_NAME}:${TRAVIS_TAG:-build-$TRAVIS_BUILD_ID}
     - MINIKUBE_WANTUPDATENOTIFICATION=false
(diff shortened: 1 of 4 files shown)


anguslees:
For posterity, the error fixed by omitempty was validation failing on `spec.data` = json `null`, and looked like:
```
SealedSecret.bitnami.com "not-testsecret" is invalid: []: Invalid value: map[string]interface {}{"kind":"SealedSecret", "apiVersion":"bitnami.com/v1alpha1", "metadata":map[string]interface {}{"namespace":"createvzpx2", "creationTimestamp":"2018-03-16T07:04:55Z", "annotations":map[string]interface {}{"sealedsecrets.bitnami.com/cluster-wide":"true"}, "uid":"54947da0-28e8-11e8-9faf-525400a05f29", "selfLink":"", "clusterName":"", "name":"not-testsecret"}, "spec":map[string]interface {}{"data":interface {}(nil), "encryptedData":map[string]interface {}{"foo":"AgAA..."}}}: validation failure list:        
          spec.data in body must be of type string: "null"
```

PR #88, opened 2018-04-19
Make sure secret type does not get dropped after sealing and unsealing

### LINKED ISSUE #86: Support docker-registry secret type
I'm trying to seal a docker registry secret, but when the controller unseals the secret it is always of `type: Opaque`.

Sample:
```
$ kubectl create secret docker-registry my-registry \
  --docker-server=myserver.example.com \
  --docker-username=user \
  --docker-password=password \
  --docker-email=fake@example.com \
  --output json --dry-run | kubeseal > sealed.json
$ kubectl apply -f sealed.json
$ kubectl get secret my-registry -o yaml
```
```yaml
apiVersion: v1
data:
  .dockerconfigjson: [snip]
kind: Secret
metadata:
  creationTimestamp: 2018-04-09T20:52:03Z
  name: my-registry
  namespace: default
  ownerReferences:
    controller: true
    kind: SealedSecret
    name: my-registry
    uid: db17d8cf-3c37-11e8-98fb-0a58ac1f010d
  resourceVersion: "3425531"
  selfLink: /api/v1/namespaces/default/secrets/my-registry
  uid: db1c68ac-3c37-11e8-98fb-0a58ac1f010d
type: Opaque
```

Expected result: The secret's `type` is `kubernetes.io/dockerconfigjson`.

Actual result: The secret type is `Opaque`.

### LINKED ISSUE #87: `type` property is removed between SealedSecret and Secret
I have a SealedSecret like this
```
kind: SealedSecret
metadata:
  creationTimestamp: null
  name: ceph-admin
  namespace: ns
  labels:
    component: ceph-secrets
type:
  kubernetes.io/rbd
spec:
  encryptedData:
    key: 1234
```

But the `type: kubernetes.io/rbd` argument is not passed forward to the `Secret` object

linoplt:
This addresses #86 and #87, where the secret type is not persisted throughout the sealing/unsealing process. Secret type used to be persisted before v0.6.0 when the whole secret was being sealed, but is now ignored and lost since only secret.Data key-value pairs are being sealed.

Fixes #86
Fixes #87

### DIFF
--- a/pkg/apis/sealed-secrets/v1alpha1/types.go
+++ b/pkg/apis/sealed-secrets/v1alpha1/types.go
@@ -38,6 +38,7 @@ type SealedSecret struct {
 
 	Spec SealedSecretSpec `json:"spec"`
 
+	// +optional
 	Type apiv1.SecretType `json:"type,omitempty" protobuf:"bytes,3,opt,name=type,casttype=SecretType"`
 }
 
--- a/pkg/apis/sealed-secrets/v1alpha1/types.go
+++ b/pkg/apis/sealed-secrets/v1alpha1/types.go
@@ -1,6 +1,7 @@
 package v1alpha1
 
 import (
+	apiv1  "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 )
 
@@ -36,6 +37,8 @@ type SealedSecret struct {
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
 	Spec SealedSecretSpec `json:"spec"`
+
+	Type apiv1.SecretType `json:"type,omitempty" protobuf:"bytes,3,opt,name=type,casttype=SecretType"`
 }
 
 // +k8s:deepcopy-gen:interfaces=k8s.io/apimachinery/pkg/runtime.Object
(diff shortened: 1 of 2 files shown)


linoplt:
Obviously this still needs some integration tests.

pkg/apis/sealed-secrets/v1alpha1/types.go
```diff
@@ -36,6 +37,8 @@ type SealedSecret struct {
 	metav1.ObjectMeta `json:"metadata,omitempty"`
 
 	Spec SealedSecretSpec `json:"spec"`
+
```
anguslees: Add a `// +optional` comment telling the autogen machinery that this is optional, just like: https://github.com/kubernetes/api/blob/99b35971cc9d07896d4428dc09ad3efacef8b21f/core/v1/types.go#L4613

anguslees:
(Thanks for the fix, @linoplt !)

mgk:
Thanks for the fix! It is working well for us with secrets of type `docker-registry` and `tls`.

We've been pointing at latest for now, but would like to peg to a version number. Is a release on the way?

anguslees:
> Is a release on the way?

I want to get #92 merged (in some form) first, since that moves the Type field added in this PR to a more future-proof schema.  .. and better to collapse both those changes in a single release cycle, rather than break CRD compatibility in the middle.

Otherwise yes, consider me nagged to hurry up and finish #92 :P

Issue #55, opened 2017-09-18
Implement kubeseal version


arapulido:
```kubeseal version```should return the version of the kubeseal cli tool

PR #94, opened 2018-05-06
Add version flag to kubeseal

c-knowles:
Fixes https://github.com/bitnami-labs/sealed-secrets/issues/55.

Had to add `kubeseal` make target to PHONY list to have it work well.

Might want to add to controller as well.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -62,7 +62,7 @@ install:
     fi
   - |
     if ! which kubecfg; then
-      wget -O $GOPATH/bin/kubecfg https://github.com/ksonnet/kubecfg/releases/download/v0.7.1/kubecfg-$(go env GOOS)-$(go env GOARCH)
+      wget -O $GOPATH/bin/kubecfg https://github.com/ksonnet/kubecfg/releases/download/v0.7.2/kubecfg-$(go env GOOS)-$(go env GOARCH)
       chmod +x $GOPATH/bin/kubecfg
     fi
   - git clone --depth=1 https://github.com/ksonnet/ksonnet-lib.git
--- a/Makefile
+++ b/Makefile
@@ -43,7 +43,7 @@ kubeseal: $(GO_FILES)
 	$(GO) build -o $@ $(GO_FLAGS) -ldflags "$(GO_LD_FLAGS)" ./cmd/kubeseal
 
 %-static: $(GO_FILES)
-	CGO_ENABLED=0 $(GO) build -o $@ -installsuffix cgo $(GO_FLAGS) ./cmd/$*
+	CGO_ENABLED=0 $(GO) build -o $@ -installsuffix cgo $(GO_FLAGS) -ldflags "$(GO_LD_FLAGS)" ./cmd/$*
 
 docker/controller: controller-static
 	cp $< $@
--- a/Makefile
+++ b/Makefile
@@ -37,7 +37,7 @@ generate: $(GO_FILES)
 	$(GO) generate $(GO_PACKAGES)
 
 controller: $(GO_FILES)
-	$(GO) build -o $@ $(GO_FLAGS) ./cmd/controller
+	$(GO) build -o $@ $(GO_FLAGS) -ldflags "$(GO_LD_FLAGS)" ./cmd/controller
 
 kubeseal: $(GO_FILES)
 	$(GO) build -o $@ $(GO_FLAGS) -ldflags "$(GO_LD_FLAGS)" ./cmd/kubeseal
@@ -64,9 +64,9 @@ controller-norbac.yaml: controller-norbac.jsonnet controller.image
 test:
 	$(GO) test $(GO_FLAGS) $(GO_PACKAGES)
 
-integrationtest: kubeseal
+integrationtest: kubeseal controller
 	# Assumes a k8s cluster exists, with controller already installed
-	$(GINKGO) -tags 'integration' integration -- -kubeconfig $(KUBECONFIG) -kubeseal-bin $(abspath $<)
+	$(GINKGO) -tags 'integration' integration -- -kubeconfig $(KUBECONFIG) -kubeseal-bin $(abspath $<) -controller-bin $(abspath $(word 2,$^))
 
 vet:
 	# known issue:
@@ -82,4 +82,4 @@ clean:
 	$(RM) controller*.yaml
 	$(RM) docker/controller
 
-.PHONY: all kubeseal test clean vet fmt
+.PHONY: all kubeseal controller test clean vet fmt
--- a/Makefile
+++ b/Makefile
@@ -30,7 +30,7 @@ ifneq ($(DIRTY),)
 VERSION := $(VERSION)+dirty
 endif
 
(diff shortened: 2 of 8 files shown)


c-knowles:
The build is failing due to the same errors as master branch.

integration/kubeseal_test.go
```diff
@@ -209,3 +209,26 @@ var _ = Describe("kubeseal --fetch-cert", func() {
 			Should(Equal(certs))
 	})
 })
+
+var _ = Describe("kubeseal --version", func() {
+	var c corev1.CoreV1Interface
```
anguslees: I think `c` is unused here?

integration/kubeseal_test.go
```diff
@@ -209,3 +209,26 @@ var _ = Describe("kubeseal --fetch-cert", func() {
...
+	var input io.Reader
+	var output *bytes.Buffer
+	var args []string
+
+	BeforeEach(func() {
+		c = corev1.NewForConfigOrDie(clusterConfigOrDie())
+
+		args = append(args, "--version")
```
anguslees: I get bitten by this ginkgo misfeature all the time:  you should set `args` (not append) in the "base" `BeforeEach` (here!) to avoid accumulating `args` between otherwise independent tests.  If you've copied this from somewhere else, then it means I've made that mistake there too (again).

You won't see it in this specific example because you've only got a single `It`, but add another one and print out `args` ...

cmd/kubeseal/main.go
```diff
@@ -211,6 +215,11 @@ func main() {
 	flag.Parse()
 	goflag.CommandLine.Parse([]string{})
 
+	if *printVersion {
+		fmt.Printf("%s\n", VERSION)
```
anguslees: `fmt.Printf("kubeseal version: %s\n", VERSION)`

Just so that the output "scales" sensibly when we want to report the versions of other components (client-go?) too one day.

Makefile
```diff
@@ -14,6 +14,24 @@ KUBECONFIG ?= $(HOME)/.kube/config
 GO_PACKAGES = ./cmd/... ./pkg/...
 GO_FILES := $(shell find $(shell $(GO) list -f '{{.Dir}}' $(GO_PACKAGES)) -name \*.go)
 
+COMMIT = $(shell git rev-parse HEAD)
```
anguslees: These commands force the user to build from a git clone, rather than (eg) a source-only tarball or git export.  I'm kind of ok with that (in particular tracking dirty state is useful), but an alternative would be to just make the build environment (eg: travis) set VERSION.

Happy to follow your preference here.

.travis.yml
```diff
@@ -62,7 +62,7 @@ install:
     fi
   - |
     if ! which kubecfg; then
-      wget -O $GOPATH/bin/kubecfg https://github.com/ksonnet/kubecfg/releases/download/v0.7.1/kubecfg-$(go env GOOS)-$(go env GOARCH)
+      wget -O $GOPATH/bin/kubecfg https://github.com/ksonnet/kubecfg/releases/download/v0.7.2/kubecfg-$(go env GOOS)-$(go env GOARCH)
```
anguslees: Oh wow.

So there never was an osx build of kubecfg 0.7.1 - the osx build failure was the reason 0.7.2 was released.

So what I'm most concerned about here is that we (/I) made the change to use 0.7.1, it passed travis tests, and got merged without ever actually executing this line of code on osx :/

I think this means we should download on every build, or (better) download to a versioned filename (and hence keep the cache) but then always symlink/copy that over to the unversioned executable name.

I'm happy to put in a PR to clean that up tomorrow if you want - just a heads up that it will conflict with this patch.

Makefile
```diff
@@ -14,6 +14,24 @@ KUBECONFIG ?= $(HOME)/.kube/config
...
+else
+VERSION := $(TAG)
+endif
+
+# Check for changed files
+ifneq ($(DIRTY),)
+VERSION := $(VERSION)+dirty
+endif
+
+GO_LD_FLAGS = -ldflags "-X main.VERSION=$(VERSION)"
```
anguslees: From experience, this variable is impossible to extend, because `go` chokes on multiple `-ldflags`.  I suggest instead moving the actual `-ldflags` flag up one level:

```Makefile
GO_LD_FLAGS = -X main.VERSION=$(VERSION)

kubeseal: $(GO_FILES)
	$(GO) build -o $@ $(GO_FLAGS) -ldflags "$(GO_LD_FLAGS)" ./cmd/kubeseal
```

Makefile
c-knowles: This is slightly more automatic but we could allow override by env var. I was looking at https://github.com/go-semantic-release/semantic-release/, maybe it’s better to use it although a lot more moving pieces. WDYT?

anguslees: I don't really care - provided the git version tag ends up in the right places _somehow_ :P
Let's stick with what you've got here (using git commands).

.travis.yml
anguslees: (fixed in #97)

cmd/kubeseal/main.go
c-knowles: Yeah, I guess let's assume it's just for display for now rather than automated use.

integration/kubeseal_test.go
c-knowles: Ah, hadn't realised it's not running fmt/vet over this.

integration/kubeseal_test.go
c-knowles: Corrected it. I copied it from just above, there 4 or so others in the same file - should we correct them here?

c-knowles:
@anguslees done the first pass at that. Just progressing on adding the same to the controller.

Makefile
```diff
@@ -29,18 +29,18 @@ ifneq ($(DIRTY),)
...
 controller: $(GO_FILES)
-	$(GO) build -o $@ $(GO_FLAGS) ./cmd/controller
+	$(GO) build -o $@ $(GO_FLAGS) -ldflags "$(GO_LD_FLAGS)" ./cmd/controller
 
 kubeseal: $(GO_FILES)
-	$(GO) build -o $@ $(GO_FLAGS) $(GO_LD_FLAGS) ./cmd/kubeseal
+	$(GO) build -o $@ $(GO_FLAGS) -ldflags "$(GO_LD_FLAGS)" ./cmd/kubeseal
 
 %-static: $(GO_FILES)
 	CGO_ENABLED=0 $(GO) build -o $@ -installsuffix cgo $(GO_FLAGS) ./cmd/$*
```
anguslees: Need ldflags here too..  (This Makefile could do with some refactoring..)
In many ways, this is the only controller target that matters, since it's the one that goes into the docker image.

(:thinking: Perhaps this indicates the integration test should instead exec the controller binary in the k8s container, or verify the correctly versioned log entry exists?)

c-knowles: @anguslees hmmm yes I see. It seems like I've tested the non static version and then assumed it's always the same which it's not. I suppose we could add some docker run commands into the integration suite, going to take a look at that.

c-knowles:
I think I almost got a basic docker setup working.

Issue #90, opened 2018-04-24
Move out of kube-system namespace


BrianChristie:
When running on GKE, nothing other than core Kubernetes components can run in the `kube-system` namespace (GKE throws errors).

Most likely it should be possible to run in any namespace. It looks like kubeseal has a --namespace argument, but from brief testing it looks like this is used for the namespace of the secret, rather than for finding the sealed-secrets-controller?

anguslees:
Oh, I didn't realise GKE enforced this - good to know.

The controller should work in another namespace just fine, but you will need to modify `controller.yaml` to change the `metadata.namespace` values.

You are correct that `kubeseal --namespace` arg governs the namespace of the Secret/SealedSecret.  **The arg you want is `kubeseal --controller-namespace`** to find the controller and fetch the pubkey at runtime.

Note the recommended workflow is to run `kubeseal --fetch-cert` once, store the public key somewhere offline (trusted, but not necessarily hidden - I put it in git alongside my sealedsecrets, for example), and then use `kubeseal --cert` when encrypting.  This means you don't need access to the cluster (or RBAC privs, etc) when sealing, and also won't be affected by changes to the namespace and other operational details of the sealed-secrets controller.

Given what you say about GKE, we should a) document more clearly how to run in another namespace and b) consider just moving to another namespace by default, as you suggest.  Leaving this bug open to track those things.

vitorarins:
Hey @BrianChristie, is this still true for GKE?
I just ran `kubectl -n kube-system run --image=busybox` and it worked. My cluster is at version `1.12.6-gke.10` for master and nodes.

mkmik:
I also tried on GKE and it seems to work just fine (1.13+).

Are there other K8s distributions that would not allow admins to install stuff in `kube-system`?
Are there other good reasons for us to change the default ns (e.g. `sealed-secrets`)?

I intuitively feel that installing random stuff in `kube-system` makes me feel a bit dirty inside,
but on the other hand it's the most fool-proof way of ensuring that only admins can get Sealed Secrets private key (assuming that unprivileged users would be prevented from accessing kube-system).

Would it be trivial to craft some RBAC rules that we can bake in our YAML manifest for the `sealed-secret` namespace and which would work on any possible cluster out there? (or at least not silently leak private keys)

CC: @jjo

mkmik:
[...]

PR #99, opened 2018-05-09
Support GKE / change the default controller namespace

c-knowles:
Separating it also means we have a little bit more clarity over access to the encryption private key and it's entirely separate from any system components which might have admin permissions (say, the kubernetes dashboard).

Fixes https://github.com/bitnami-labs/sealed-secrets/issues/90.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,11 @@ release.
 **See additional TPR->CRD migration section below if updating an
 existing Sealed Secrets installation from Kubernetes <= 1.7**
 
+**See additional namespace migration section below if updating from a 
+version prior to 0.8.0, the controller was previously installed into
+`kube-system` namespace by default but to support managed services like
+GKE the default was changed to `sealed-secrets`.**
+
 ```sh
 $ release=$(curl --silent "https://api.github.com/repos/bitnami-labs/sealed-secrets/releases/latest" | sed -n 's/.*"tag_name": *"\([^"]*\)".*/\1/p')
 
@@ -34,14 +39,14 @@ $ kubectl create -f https://github.com/bitnami-labs/sealed-secrets/releases/down
 # Install SealedSecret CRD (for k8s >= 1.7)
 $ kubectl create -f https://github.com/bitnami-labs/sealed-secrets/releases/download/$release/sealedsecret-crd.yaml
 
-# Install server-side controller into kube-system namespace (by default)
+# Install server-side controller into `sealed-secrets` namespace (by default)
 $ kubectl create -f https://github.com/bitnami-labs/sealed-secrets/releases/download/$release/controller.yaml
 ```
 
 After the `SealedSecret` resource is created with either
 `sealedsecret-tpr.yaml` or `sealedsecret-crd.yaml` (depending on
 Kubernetes version), `controller.yaml` will install the controller
-into `kube-system` namespace, create a service account and necessary
+into `sealed-secrets` namespace, create a service account and necessary
 RBAC roles.
 
 After a few moments, the controller will start, generate a key pair,
@@ -162,6 +167,52 @@ the generic documentation for more information on the process.
    $ kubectl scale --replicas=1 -n kube-system deployment/sealed-secrets-controller
    ```
 
+### Migration from SealedSecret <0.8.0 to 0.8.0+
+
+In versions prior to 0.8.0 the controller was installed into
+`kube-system` namespace by default but to support managed services like
+GKE the default was changed to `sealed-secrets`.
+[Issue ref](https://github.com/bitnami-labs/sealed-secrets/issues/90)
+
+To migrate, follow this procedure:
+
+1. Backup your existing encryption private key:
+   ```sh
+   $ kubectl get secret -n sealed-secrets sealed-secrets-key -o yaml >master.key
+   ```
+
+2. Prepare to load your encryption private key into the new namespace:
+   ```sh
+   $ sed -i 's/kube-system/sealed-secrets/g' master.key
+   ```
+   Or on Mac OS:
+   ```sh
+   $ sed -i '' 's/kube-system/sealed-secrets/g' master.key
+   ```
+
+3. Install the latest sealed-secrets controller into the `sealed-secrets` namespace (the new default) and wait a minute or so for it to boot
+   ```sh
+   $ release=$(curl --silent "https://api.github.com/repos/bitnami-labs/sealed-secrets/releases/latest" | sed -n 's/.*"tag_name": *"\([^"]*\)".*/\1/p')
+   $ kubectl create -f https://github.com/bitnami-labs/sealed-secrets/releases/download/$release/controller.yaml
+   ```
+
+4. Scale down the new controller and then override the secret it just created with your existing one you just prepared in steps 1/2:
+   ```sh
+   $ kubectl scale deployment sealed-secrets-controller -n sealed-secrets --replicas=0
+   $ kubectl replace secret -n sealed-secrets sealed-secrets-key master.key
+   ```
+
+5. Delete the old controller and encryption private key from the `kube-system` namespace:
+   ```sh
+   $ kubectl delete deployment sealed-secrets-controller -n kube-system
+   $ kubectl delete secret sealed-secrets-key -n kube-system
+   ```
+
+6. Finally, scale back up the new controller:
+   ```sh
+   $ kubectl scale deployment sealed-secrets-controller -n sealed-secrets --replicas=1
+   ```
+
 ## Usage
 
 ```sh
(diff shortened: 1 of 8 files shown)


README.md
```diff
@@ -162,6 +167,52 @@ the generic documentation for more information on the process.
...
+   $ kubectl get secret -n sealed-secrets sealed-secrets-key -o yaml >master.key
+   ```
+
+2. Prepare to load your encryption private key into the new namespace:
+   ```sh
+   $ sed -i 's/kube-system/sealed-secrets/g' master.key
+   ```
+   Or on Mac OS:
+   ```sh
+   $ sed -i '' 's/kube-system/sealed-secrets/g' master.key
```
anguslees: Really, osx doesn't have a "normal" `sed`?  Perhaps we can do `sed <master.key >moved.key` or something that works on both platforms..

README.md
```diff
@@ -162,6 +167,52 @@ the generic documentation for more information on the process.
...
+   ```
+
```
anguslees: I think we should create the secret here, before installing the new controller.  That way we don't need to restart the controller post-install.

Something like:
```markdown
3. Load the existing master secret key into the new namespace:
    ```sh
    $ kubectl create -f moved.key
    ```
```

What I _actually_ think we should do is combine these 3 steps into one, since I think it's easier to see what's going on as a single step (please disagree if you don't think so):

```markdown
1. Copy your existing master key to the new namespace:
    ```sh
    $ kubectl create namespace sealed-secrets
    $ kubectl get secret -n sealed-secrets sealed-secrets-key -o yaml |
       sed 's/kube-system/sealed-secrets/' |
       kubectl create -f -
    ```
2. Install the new controller.
3. Delete the old kube-system controller.
4. Verify unsealing new/modified keys is working as expected, then delete the old kube-system master key.
```

anguslees:
Travis is failing because we need to create the new namespace.

Need to add something like the following, at ~L27 in controller-norbac.jsonnet:
```javascript
   ns: kube.Namespace($.namespace.metadata.namespace),
```

README.md
c-knowles: I've gone ahead and updated this, can always re-order 3+4 if there's any possibility for race conditions.

c-knowles:
Looks like I need a way to force namespace to the top of the YAML file with kubecfg.

anguslees:
That's very odd:
```
namespace "sealed-secrets" created
 ...
Error from server (NotFound): error when creating "controller.yaml": namespaces "sealed-secrets" not found
```

README.md
anguslees: We already have the possibility of 2 controllers running simultaneously for a brief period during a "normal" controller upgrade (same namespace/Deployment upgrade).  This should be fine (at worst we decrypt a SealedSecret twice).

c-knowles:
Does kubectl wait for the namespace to be available for use?

anguslees:
Wat :man_facepalming:  I turned on extra kubectl debugging in #100.  I learned:
- `kubectl` delays the error responses and prints them after trying to create everything - it does not abort at the first error.  So the ordering of output messages/errors is misleading.
- `kubectl` does no sorting of input, it just tries them in order (I knew this).
- `kubecfg` does *not* sort the `show` output in dependency order.  I distinctly remember adding this code a long time ago, and just assumed that this was how `kubecfg` behaved - but I can't find any mention of it now in the kubecfg git log.  This is the real error here :(

I will fix this in kubecfg, but I'd like to not block this PR on waiting for a kubecfg release.
Our options are:
1. Use `kubecfg update controller.yaml` rather than `kubectl create -f`.  This is tempting, but I would like to test our published install instructions here, so we can uncover problems (like this one!)
2. Run an explicit separate `kubectl create namespace sealed-secrets` command first.
3. Change the way we generate the `controller-*.yaml` files so that we place the Namespace before any namespaced resources, without relying on `kubecfg` to do this for us.  This probably means (temporarily) removing the Namespace from jsonnet, and adding it using a shell cat or similar in the make rule.
4. Wait until a fixed kubecfg is released that sorts the namespace early in YAML output.

Whatever we do, we should update the install instructions similarly - since this CI failure is showing us what users will experience when they try the same.  I think this means I'd prefer (3) (generate the right yaml somehow), since that contains the hackery within that small build step and doesn't affect users.

c-knowles:
@anguslees how about https://github.com/bitnami-labs/sealed-secrets/pull/99/commits/e36869b7f252ba3f40abc294cc86d4b705f8554e?

anguslees:
lgtm, but CI is still choking with:
```
Error from server (AlreadyExists): error when creating "controller.yaml": namespaces "sealed-secrets" already exists
```

I don't see anywhere obvious where we're double-creating the namespace, and I've run out of debugging time for today :(

Wow, this looked like such a straightforward change :(  Fwiw, I suspect we should change our install instructions / CI script to say `kubectl apply -f` rather than `create` - and this would also paper over issues like the above.  I would rather understand the issue better first, but that's an option ...

lewisdawson:
Is this still being worked on and considered for merging into a release?

c-knowles:
@lewisdawson I looked into this for a while but couldn't spot the problem. I suspect parts of the CI setup need to be reworked slightly. If you have any ideas how to fix that part I think the rest is mergeable.

jramos:
This fixes CI for me:

```
diff --git a/.travis.yml b/.travis.yml
index 2480d3e..c6ca177 100644
--- a/.travis.yml
+++ b/.travis.yml
@@ -93,7 +93,7 @@ script:
       minikube update-context
       minikube status
       while ! kubectl cluster-info; do sleep 3; done
-      kubectl create -f $INT_SSC_CONF
+      kubectl apply -f $INT_SSC_CONF
       kubectl rollout status deployment/sealed-secrets-controller -n sealed-secrets -w
       make integrationtest CONTROLLER_IMAGE=$CONTROLLER_IMAGE
     fi
```

c-knowles:
@anguslees are you happy for us to change the create to apply? You mention above it's only papering over the problem. I tend to agree but I'd prefer to have GKE support sooner.

c-knowles:
@anguslees I've updated this PR so the CI is working again, let me know if you want any other changes or you'd like me to squash/rebase it etc.

ricardompcarvalho:
Hello @c-knowles @anguslees ,
I'm trying to migrate sealed secrets from namespace kube-system to namespace sealed-secrets but gives me error!
I'm following the readme step where says "Migration from SealedSecret <0.8.0 to 0.8.0+", but in step number 3 gives me that for example : "_sealed-secrets-controller" already exists_".
Can someone help me, please?

Ricardo Carvalho

c-knowles:
@ricardompcarvalho perhaps try reversing steps 3 and 4 to delete the old controller first, as long as you’ve done the backup in step 1 it should not be dangerous. If that works better I can update the PR before it gets merged.

ricardompcarvalho:
@c-knowles thanks for answering, but i dont understand exactly what you say about "revert"!!!
Revert does not work because it is installed in "kube-system" and when doing the command kubectl create it tries to create in "kube-system" and not in "sealed-secrets". Another question I have is: do I have to create the "sealed-secrets" namespace or does it automatically create when we do kubectl

c-knowles:
@ricardompcarvalho reverse rather than revert, to clarify I mean run step 4 first and then 3 after. To try to answer your other questions:

- Unless you are using a CLI binary built from this branch which is not merged or released yet then the default namespace will still be kube-system. If you want to do the same using a stable release you will need to add the additional flag I updated the default of in this PR (`controller-namespace`).
- Generally kubectl will not auto-create any objects, the namespace has an explicit create step documented in this PR.

ricardompcarvalho:
@c-knowles thanks once more.
I'm working on stable Jenkins allocated in GKE and i want to add sealed-secrets to my work, i have some questions if you allow to ask:

- I must first install sealed-secrets on kube-system namespace, only then do the migration?

- If i'm working on a different CLI don't work?

kskewes:
@ricardompcarvalho - we installed directly into `sealed-secrets` namespace in each of our clusters.
Just update all the yaml files with whatever namespace you like before applying them and you'll be good to go. You can test it with minikube or similar.

mkmik:
I'm not against the spirit of this PR, but it has become stale and we need more discussion in #90 

It seems that nowadays you can use `kube-system` on GKE so it seems to be less blocking.
That said, a lot of people I've talked to feel a bit uneasy polluting the `kube-system` namespace.

On the other hand that's usually the most foolproof way of ensuring that your RBAC rules do protect your private key properly. Anyway let's keep the discussion in #90 and possibly in #233.

c-knowles:
For what it’s worth what is here does work, I’ve used it. No need to install elsewhere and migrate on the above question, the new default here is a namespace specific to sealed-secrets which ensure RBAC etc is more easily managed.

mkmik:
@c-knowles sorry could you please rephrase? My non-native brain still does a poor job parsing some English sentences.

c-knowles:
Sure, the main things I wanted to convey
- PR was tested installed from scratch plus the migration path
- There was a question above about installing to kube-system first and then migrating but there is no need for this
- Separation eases RBAC management since it’s now separated from kube-system

Happy to close it off as you did, I just wanted to update as hadn’t had any action on it and no longer actively using this project.

mkmik:
Thanks. Yeah, as I said I'd personally lean towards defaulting to a dedicated namespace in principle but as this PR no longer applies cleanly on master I'd like to first understand the implications before putting some work to adapt against HEAD.

For example, one pet peeve of mine is that it should be possible to use sealed secrets in both "system" mode and in "bring your own controller" mode and as part of that journey we might have the sealed secrets controller create a new sealing keys per namespace. If we have that, the question of whether a `sealed-secrets` namespace is "protected enough" in the average user's RBAC config becomes moot.

PR #121, opened 2018-10-24
Add the ability to check whether a sealed secret is valid

### LINKED ISSUE #118: Allow user to check if a sealed secret can be unsealed by controller
theres no way of knowing whether a cluster can unseal a sealed-secret. 

could have something like this:
``` kubeseal verify <secretcontent.yaml>``` with an exit code 
ie: 
if unsealable - exit code 0 else not

this is kind of related to #72 but more on a cli level

wjam:
Make it easier to verify whether a sealed secret is valid, without having to rely on reading the logs of the controller. Example: `cat manifest.yml | kubeseal --validate`
Fixes #118

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -18,4 +18,7 @@
 /controller.image
 /*-static
 /controller.yaml
+/sealedsecret-crd.yaml
 /docker/controller
+*.iml
+.idea
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -3,6 +3,7 @@ package main
 import (
 	"crypto/rsa"
 	"fmt"
+	"k8s.io/apimachinery/pkg/runtime"
 	"log"
 	"time"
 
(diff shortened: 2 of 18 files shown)


lavoiedn:
Hi there @anguslees !

I'll chime in for the motivation behind such a validation mechanism, as issue https://github.com/bitnami-labs/sealed-secrets/issues/118 was filed by someone from my team!

At our company, we use a gitops flow for all of our kubernetes resources, having a repository to reflect the state of each of our clusters. Having `SealedSecret` allows us to keep them in source revision system.

We currently use Github webhooks to initiate verification/deployments, running yaml verifications with a dry-run, vvalidating presence of images tags, etc.

The issue we have is that, if a secret is improperly encrypted, the deployment flow will work, but secrets will not be updated. The only indication of this will be in the sealed-secret-controller's logs.

Therefore, we were looking for a way to run a validation check to ensure that a secret was properly encrypted for a given sealed-secret-controller/cluster without overwriting the existing secret.

wjam:
Pushed change to fix the review comments.

The use case we have for this is similar to @lavoiedn - there's a Git repository containing ConfigMaps & SealedSecrets and being able to add a step to verify that the secret could be unsealed before it gets applied would help avoid simple mistakes such as not having the same name/namespace in the initial secret as with the committed SealedSecret.

anguslees:
Thanks for the additional context @lavoiedn @wjam - I agree some sort of pre-submit check requires an approach like this one (or exporting the private key elsewhere - ouch).

bors r+

BizAhmad:
@anguslees would you be able to tell roughly when the next version including these changes will be released?
thanks!

PR #130, opened 2018-11-23
Retain existing OwnerReferences

### LINKED ISSUE #127: Controller drops foreign OwnerReferences
When reconciling a `SealedSecret` where the corresponding `Secret` already exists in the cluster, the controller will replace that `Secret` with a fresh copy, overwriting any existing `ObjectMeta`.

This drops all existing `OwnerReferences` on the `Secret` and causes other controllers in the same cluster to misbehave where they rely on these `OwnerReferences` to be present.
Our [Wave Controler](https://github.com/pusher/wave) uses `OwnerReferences` to tie `Secrets` to `Deployments` that use them and fails to correctly track `Secrets` that are also owned by a `SealedSecret`.

The problematic code is at https://github.com/bitnami-labs/sealed-secrets/blob/2e3902bcd6bdc870894f5599b87b6b3583e2bfc5/cmd/controller/controller.go#L192-L194 where `Update()` is called without merging with existing `OwnerReferences` first.

I'm currently working on a PR to fix this.

gargath:
This PR ensures that existing `OwnerReferences` on Secrets managed by the controller are retained when a Secret is reconciled.
When an existing Secret is found, the controller will now merge existing OwnerReferences with its own, ensuring that other controllers can use OwnerReferences safely to track these Secrets.

Fixes #127

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -190,7 +190,14 @@ func (c *Controller) unseal(key string) error {
 
 	_, err = c.sclient.Secrets(ssecret.GetObjectMeta().GetNamespace()).Create(secret)
 	if err != nil && errors.IsAlreadyExists(err) {
+		// Fetch existing owner references
+		existingSecret, err := c.sclient.Secrets(ssecret.GetObjectMeta().GetNamespace()).Get(secret.GetObjectMeta().GetName(), metav1.GetOptions{})
+		if err != nil {
+			return err
+		}
+		secret.SetOwnerReferences(append(existingSecret.GetOwnerReferences(), secret.GetOwnerReferences()...))
 		_, err = c.sclient.Secrets(ssecret.GetObjectMeta().GetNamespace()).Update(secret)
+
 	}
 	return err
 }
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -191,7 +191,11 @@ func (c *Controller) unseal(key string) error {
 	}
 
 	_, err = c.sclient.Secrets(ssecret.GetObjectMeta().GetNamespace()).Create(secret)
-	if err != nil && !errors.IsAlreadyExists(err) {
+	if err == nil {
+		// Secret successfully created
+		return nil
+	}
+	if !errors.IsAlreadyExists(err) {
 		// Error wasn't already exists so is real error
 		return err
 	}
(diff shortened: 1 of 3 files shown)


gargath:
Looking into these failling integration tests.

gargath:
Missing RBAC "get" for secret. Fixed now.

gargath:
FWIW, we're now running this latest build in production and it has resolved the problem for us.

gargath:
@anguslees, any chance to get this merged soon?

cmd/controller/controller.go
```diff
@@ -189,12 +191,50 @@ func (c *Controller) unseal(key string) error {
 	}
 
 	_, err = c.sclient.Secrets(ssecret.GetObjectMeta().GetNamespace()).Create(secret)
-	if err != nil && errors.IsAlreadyExists(err) {
-		_, err = c.sclient.Secrets(ssecret.GetObjectMeta().GetNamespace()).Update(secret)
+	if err != nil && !errors.IsAlreadyExists(err) {
+		// Error wasn't already exists so is real error
+		return err
 	}
+
```
anguslees: On a fresh create, I think you're now doing a Create followed by an immediate Update.

I think we need a `if err == nil { return nil }` about here (or equivalent change to the above predicate) to avoid the unnecessary update.

cmd/controller/controller.go
```diff
@@ -189,12 +191,50 @@ func (c *Controller) unseal(key string) error {
...
+	_, err = c.sclient.Secrets(ssecret.GetObjectMeta().GetNamespace()).Update(updatedSecret)
 	return err
 }
 
+func (c *Controller) updateSecret(newSecret *apiv1.Secret) (*apiv1.Secret, error) {
+	existingSecret, err := c.sclient.Secrets(newSecret.GetObjectMeta().GetNamespace()).Get(newSecret.GetObjectMeta().GetName(), metav1.GetOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("failed to read existing secret: %s", err)
+	}
+	existingSecret.Data = newSecret.Data
```
anguslees: `existingSecret` may have come from a shared cache above, so you need to copy it before modifying it:

`existingSecret = existingSecret.DeepCopy()`

anguslees:
otherwise lgtm, thanks!

(apologies for not responding earlier - kubecon, holidays, and other excuses)

cmd/controller/controller.go
gargath: That should do it

cmd/controller/controller.go
gargath: Good catch. Fixed.

gargath:
@anguslees No worries. Thanks for taking a look. I think I have addressed your concerns. Could you re-review?

gargath:
@anguslees Could you give this another once-over? Keen to getting it merged.

Issue #92, opened 2018-04-25
Missing annotations & labels


jalberto:
When creating a new sealed-secret based on a existing secret.yaml labels and annotations are lost.

This is very important, as some labels and annotations are related to helm lifecycle.

jalberto:
is this intended behaviour?

anguslees:
No, it's another thing that got lost in the 0.6.0 change to per-key encryption.  If you use 0.5.x kubeseal (perhaps even with a 0.6 controller), then you should have annotations+labels preserved.

I think the desired semantics with per-key encryption, is that the labels+annotations will end up in plaintext (no encryption), and in a new `spec.template` subfield of the SealedSecret.  Would this be ok for your usecase (in particular, that the labels/annotations are not hidden by encryption)?

jalberto:
@anguslees yes, that will be perfect, labels/annotations should not keep any sensible data.

In other hand, "another thing that got lost in the 0.6.0 change" I am worried to use kubeseal in production if there is risk of accidentally loosing features like this from one stable version to the next.

c-knowles:
@anguslees we'd be happy to fix this. I can't see much difference between combined encryption and per-key in [the PR for that](https://github.com/bitnami-labs/sealed-secrets/pull/73/files). I went through and didn't spot much about the labels/annotations from the [original import](https://github.com/bitnami-labs/sealed-secrets/commit/150b3ac09efa411e4ea9d381cd0ad97a2543a939) so probably missed something. If you could point us in the right direction we can submit a PR.

anguslees:
> In other hand, "another thing that got lost in the 0.6.0 change" I am worried to use kubeseal in production if there is risk of accidentally loosing features like this from one stable version to the next.

(Fwiw, the "other thing" was the Secret.Type value - ie: secrets of types other than "Opaque")

I'm afraid I can't give a useful response to non-specific future concerns.  As with everything else, a conservative deployment involves qualifying new releases of your infrastructure against your requirements.  If you can't/don't want to do that, then you can also just stick with a fixed (unchanging) version.  Again, as with everything else, the safest option for continuing upgrades is to become actively involved in upstream development and ensure that your particular use case is well represented in tests and design discussions.

anguslees:
@c-knowles thanks for the offer.  I think I've got most of this implemented already, however (see below).
The "real" fix is adding tests that better exercises these corner-cases with the new per-key scheme.

> If you could point us in the right direction we can submit a PR

[...]

PR #135, opened 2018-12-13
Add a note about the secret type bug

johnraz:
I just got bitten by the bug in #86 and #92 and lost quite a bit of time finding out what was happening.
This will hopefully help other people figure it out faster until the proper fix is released.

### DIFF
--- a/README.md
+++ b/README.md
@@ -168,6 +168,8 @@ the generic documentation for more information on the process.
 
 ## Usage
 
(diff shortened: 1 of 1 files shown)


victornoel:
@johnraz @anguslees quick question: the changelog from [the releases page](https://github.com/bitnami-labs/sealed-secrets/releases) says that the per-key encryption was introduced in 0.7.0 and in 0.7.0, so shouldn't 0.6.0 enough to circumvent this bug? Or am I missing something?

Thanks :)

johnraz:
@victornoel I can’t remember if it tested 0.6.0 or not but I guess you could quickly do it and find out ;-)

victornoel:
@johnraz @anguslees I can confirm that it works with 0.6.0

Issue #137, opened 2019-01-16
Rotating master key


peterbosalliandercom:
For security reasons we would like to rotate the master key frequently because the private key could be copied from the cluster and reused in another cluster to decrypt sealed secrets which are in a public repo. 

Is this possible?

mkmik:
The bulk of the feature has been implemented in #143 but there are still some rough edges:

@grossvogel wrote in #95
> I apologize if this is off topic, but I've been looking at this awesome tool and wondering specifically about updating existing secrets. Can somebody tell me whether the work planned for this feature would allow e.g. adding new data to a sealed secret whose encryption key has been rotated out?

> My impression is that (a) each sealed secret currently has a single applicable encryption key and (b) older secrets aren't automatically re-encrypted upon key rotation and (c) only the latest key is currently accessible to the client... but maybe / hopefully I'm missing something?

@grossvogel, the new `--rotate` flag allows users to update their local checked-in copy of the sealed secret by decrypting an existing secret.

This feature is still underdocumented and probably there are many rough edges. I'll work on this as soon as I cut a release with the rest of the long overdue but more stable features.

mkmik:
Status update: the bulk of this feature is going to be released in v0.8.0 but not turned on by default.
Full rollout blocked on #185 and possibly other issues; I still didn't have enough time to fully appreciate the implications of this.

PR #143, opened 2019-02-22
Key rotation

anzboi:
This is a set of proposed changes to add key rotation to sealed secrets. Key rotation refers to regular generation of a new private/public key pair. This adds a layer of security to sealed secrets but makes it harder to work with when you want to keep secrets up to date. Changes proposed here ensure seamless integration with current sealed secrets usage.

Proposed changes include:
* Regular private/public key pair generation on a configurable period. Pairs are referenced by a unique name identifier. Default generation period is 30 days.
* Retrieval of the latest public key via the /cert.pem endpoint
* A /rotate endpoint to rotate a `SealedSecret`. This refers to decrypting and old `SealedSecret`, re-encrypting it with the latest key (or any selected key specified by name), and returning the new `SealedSecret`. Endpoint can be accessed via the `--rotate` flag with kubeseal.
* Ability to blacklist a key. This works by modifying the `sealedsecrets.bitnami.com/sealed-secrets-key` label from `active` to anything else. After a controller restart this effectively deletes the key from sealed secrets, but leaves it available to k8s admins.
* Ability to force generating a new key. This is done by sending `SIGUSR 1` signal to the controller.
* Updates to README.md to address the new changes.

Changes do NOT break backward compatibility as long as you label the current `sealed-secrets-key` with the label `sealedsecrets.bitnami.com/sealed-secrets-key=active`. You will need to start the controller after adding this label.

Addresses #137


anzboi:
Notes on the current state of this PR:
* This is still a work in progress, all feedback welcome.
* Missing new integration tests to test the new features (unit tests are written).
* The sealed secrets admin interface is implemented by adding a listener on a second port of the controller which the `ssadmin` command hits. This port is currently not configurable. It is also currently implemented using `go`s native `rpc` library.

README.md
```diff
@@ -232,6 +247,55 @@ namespace/name is used as the OAEP input parameter, ensuring that the
...
+
+### Blacklisting
+Sealed secrets provides a blacklisting feature to warn users of secrets
+that may have been encrypted with a compromised private key. The
+controller maintains a list of key names corresponding to keys that are
+suspected to be compromised. This list is also stored as a secret in
+the controllers namespace and maintained similarly to the key list
+itself.
+
+The `SealedSecret`s controller will refuse to work with blacklisted
```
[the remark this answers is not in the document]
anzboi: I originally thought to keep them around in the case the original secrets are lost somehow. A blacklisted private key can still be used to un-encrypt it if need be. I'm not sure how useful that actually is though.

cmd/controller/funcs.go
```diff
@@ -0,0 +1,121 @@
...
+}
+
+func createKeyGenJob(client kubernetes.Interface,
+	keyRegistry *KeyRegistry,
+	namespace, listname string,
+	keySize int,
+	nameGen keyNameGen,
+) func() error {
+	return func() error {
+		newKeyName, err := nameGen()
```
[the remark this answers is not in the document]
anzboi: Thanks, I wasn't aware of the namegen feature.

cmd/controller/server.go
```diff
@@ -57,11 +88,36 @@ func httpserver(cp certProvider, sc secretChecker) {
...
 			w.WriteHeader(http.StatusConflict)
 		}
-
 	})))
 
+	mux.HandleFunc("/v1/rotate", func(w http.ResponseWriter, r *http.Request) {
+		content, err := ioutil.ReadAll(r.Body)
+
+		if err != nil {
+			log.Printf("Error handling /v1/rotato request: %v", err)
```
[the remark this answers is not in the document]
anzboi: Thanks

pkg/apis/sealed-secrets/v1alpha1/types.go
```diff
@@ -26,8 +26,9 @@ const (
 // SealedSecretSpec is the specification of a SealedSecret
 type SealedSecretSpec struct {
 	// Data is deprecated and will be removed eventually. Use per-value EncryptedData instead.
-	Data          []byte            `json:"data,omitempty"`
-	EncryptedData map[string][]byte `json:"encryptedData"`
+	Data              []byte            `json:"data,omitempty"`
+	EncryptedData     map[string][]byte `json:"encryptedData"`
+	EncryptionKeyName string            `json:"keyname"`
```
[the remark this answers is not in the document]
anzboi: I think I'm inclined to go with implicit as well, mainly because it's not a breaking change.

anzboi:
Getting back from busy work on other projects. A few general notes
* With regards to blacklisting, I keep the blacklisted private keys around in case users still want to use them to decrypt secrets with them. This is what requires us to keep the blacklist around, as opposed to using annotations.
* I think the features I provided allow users to do too much with keynames. Is it preferable to hide keynames completely?
* I like removing `ssadmin`.

cmd/controller/server.go
```diff
@@ -77,6 +133,20 @@ func httpserver(cp certProvider, sc secretChecker) {
 		}
 	})
 
+	mux.HandleFunc("/v1/keyname", func(w http.ResponseWriter, r *http.Request) {
```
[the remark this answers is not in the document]
anzboi: I agree, Probably better to hide keynames as much as possible to encourage use of latest key always.

README.md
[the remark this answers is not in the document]
anguslees: I could imagine for key-rotation that it would be useful to "deprecate" a key - still use it for decryption but try to prevent it being used for new encryption (hide it from --fetch-cert output, etc).

I think I want a clear workflow use-case here to make sure the extra code is actually helping someone.  Until then, I suggest we remove the blacklist code/feature until someone asks for it (we can go from nothing->something more easily than we can go something->something_else once it's a public released feature).

anguslees:
> With regards to blacklisting, I keep the blacklisted private keys around in case users still want to use them to decrypt secrets with them. This is what requires us to keep the blacklist around, as opposed to using annotations.

I think there are two subthreads to the blacklist conversation:

- Do we want a blacklist? What semantics are most useful?

  I think we've discussed:
  1. No blacklist.  Deleting a key means deleting the key Secret.  No middle-ground.
  2. Blacklist means pretend-it's-deleted.  Allows for easy "undelete" (but not much else?).
  3. Blacklist means prevent-for-new-encryption, but is still used for decrypting existing SealedSecrets.  Allows for a staged deprecation.  Unsure how this differs from just changing the "default" key.

  What's the workflow we want to implement?  (and is that important/clear enough to implement now or should we wait and get more feedback on the design?)  I'm not trying to skew this either way, I honestly don't know the answer to these questions :P

- How do we store the blacklist in k8s resources?

  Assuming we want a blacklist, I still suggest we store the blacklist as an annotation on the Secret objects.  It's obvious, the lifecycle of blacklist entries then automatically matches the keys themselves, avoids managing a new ConfigMap, etc.
  The advantage of a separate blacklist is that it can have separate RBAC policy, and therefore different people can manage the blacklist without actually having read (/write) access to the real Secrets. The blacklist entries in a separate blacklist can also outlive the actual Secrets - unsure if this is a pro or con ;)

  I don't think storing the blacklist "decentralised" in annotations or "centralised" in a dedicated ConfigMap affects our answers to the earlier "what semantics do we want?" question.

> I think the features I provided allow users to do too much with keynames. Is it preferable to hide keynames completely?

Yes, I agree we should hide the keynames unless we actually need them for something (and can't just use the k8s metadata.name).

anzboi:
All issues addressed up to this point.

**Note**: The controller currently does not react to labelling a key as compromised. The only way to have it respond is to restart the controller pod. I want to use an informer similar to the sealed secret informer.

cmd/controller/keyregistry.go
```diff
@@ -0,0 +1,86 @@
...
+}
+
+func (kr *KeyRegistry) registerNewKey(keyName string, privKey *rsa.PrivateKey, cert *x509.Certificate) {
+	kr.keys[keyName] = privKey
+	kr.certs[keyName] = cert
+	kr.currentKeyName = keyName
+}
+
+func (kr *KeyRegistry) latestKeyName() string {
+	return kr.currentKeyName
```
anguslees: I think this is "last registered", not necessarily "latest".  Perhaps when initialising the registry from existing keys, we should find the newest by `metadata.creationTimestamp`?

cmd/controller/keyregistry_test.go
```diff
@@ -0,0 +1,103 @@
...
+
+	"k8s.io/client-go/kubernetes/fake"
+)
+
+func writeTestKeylistSecret(client *fake.Clientset) error {
+	privkey, cert, err := generatePrivateKeyAndCert(1024)
+	if err != nil {
+		return err
+	}
+	if err := writeKeyRegistry(client, privkey, cert, "namespace", "listname"); err != nil {
```
anguslees: These `writeKeyRegistry` tests are obsolete and should be removed.

(See also travis failures: https://travis-ci.org/bitnami-labs/sealed-secrets/jobs/521534393)

cmd/controller/keys.go
```diff
@@ -0,0 +1,126 @@
...
+	for _, cert := range certs {
+		certbytes = append(certbytes, certUtil.EncodeCertPEM(cert)...)
+	}
+
+	secret := v1.Secret{
+		ObjectMeta: metav1.ObjectMeta{
+			Namespace:    namespace,
+			GenerateName: prefix,
+			Labels: map[string]string{
+				label: "active", // Value is not currently used to find keys
```
anguslees: Is this comment still correct?

cmd/controller/keys.go
```diff
@@ -0,0 +1,126 @@
...
+	"math/big"
+	"time"
+
+	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	certUtil "k8s.io/client-go/util/cert"
+)
+
+const SealedSecretsKeyLabel = "sealed-secrets-key"
```
anguslees: This should be namespaced to avoid collisions with local values / other controllers: `sealedsecrets.bitnami.com/sealed-secrets-key`

cmd/controller/main.go
```diff
@@ -2,42 +2,41 @@ package main
...
 	ssinformers "github.com/bitnami-labs/sealed-secrets/pkg/client/informers/externalversions"
 )
 
 var (
-	keyName      = flag.String("key-name", "sealed-secrets-key", "Name of Secret containing public/private key.")
-	keySize      = flag.Int("key-size", 4096, "Size of encryption key.")
-	validFor     = flag.Duration("key-ttl", 10*365*24*time.Hour, "Duration that certificate is valid for.")
-	myCN         = flag.String("my-cn", "", "CN to use in generated certificate.")
-	printVersion = flag.Bool("version", false, "Print version information and exit")
+	keyLabelName    = flag.String("key-label", "sealed-secrets-key", "Label used to identify public/private key pairs in k8s.")
```
anguslees: I _think_ `keyLabelName` is ignored by the code, which is good since I was going to suggest we remove it and just use a hard-coded value :stuck_out_tongue: 
I think we should remove this and just use a hardcoded value.

If we want to support multiple sealed-secrets controllers in the same k8s namespace, then I think we should just have a flag for arbitrary additional label selectors.

cmd/controller/main.go
```diff
@@ -2,42 +2,41 @@ package main
...
+	validFor        = flag.Duration("key-ttl", 10*365*24*time.Hour, "Duration that certificate is valid for.")
+	myCN            = flag.String("my-cn", "", "CN to use in generated certificate.")
+	printVersion    = flag.Bool("version", false, "Print version information and exit")
+	keyRotatePeriod = flag.Duration("rotate-period", 30*24*time.Hour, "New key generation period")
 
 	// VERSION set from Makefile
 	VERSION = "UNKNOWN"
+
+	// Selector used to find existing public/private key pairs on startup
+	keySelector = SealedSecretsKeyLabel + "=" + "active"
```
anguslees: If you want to be fancy, this could be:
```go
import "k8s.io/apimachinery/pkg/fields"

keySelector = fields.OneTermEqualSelector(SealedSecretsKeyLabel, "active")
```

And then call `.String()` on it when building the actual k8s label query.  This would be particularly useful if including another selector (perhaps from a command line flag), since there are functions in that package for building And queries, etc.

cmd/controller/main.go
```diff
@@ -53,108 +52,39 @@ type controller struct {
...
-func readKey(client kubernetes.Interface, namespace, keyName string) (*rsa.PrivateKey, []*x509.Certificate, error) {
-	secret, err := client.Core().Secrets(namespace).Get(keyName, metav1.GetOptions{})
-	if err != nil {
-		return nil, nil, err
-	}
-
-	key, err := certUtil.ParsePrivateKeyPEM(secret.Data[v1.TLSPrivateKeyKey])
-	if err != nil {
-		return nil, nil, err
+func initNames(prefix, label *string) (string, string, error) {
```
anguslees: I'm confused by this function ;)
The name is very generic, it takes references to values (which it modifies), and then it also returns values? 

Afaics, all it _actually_ does is check the characters on each value, and maybe append a `-` to one of them.  I think I would prefer we just fix the default values (to include the `-` suffix or not), and then let the k8s server return an error if the user gave us something with the wrong characters.

cmd/controller/server.go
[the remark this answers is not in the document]
anguslees: I think we still need to remove this `/v1/keyname` endpoint and the keyname arg to `/v1/cert.pem`

cmd/kubeseal/main.go
```diff
@@ -2,16 +2,18 @@ package main
...
 	goflag "flag"
 	"fmt"
 	"io"
 	"io/ioutil"
-	"k8s.io/apimachinery/pkg/util/net"
 	"net/http"
 	"os"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/util/net"
```
anguslees: Remove the whitespace after this import (should be combined with the rest of the 3rd-party imports)

cmd/kubeseal/main.go
```diff
@@ -36,7 +38,8 @@ var (
 	controllerNs   = flag.String("controller-namespace", metav1.NamespaceSystem, "Namespace of sealed-secrets controller.")
 	controllerName = flag.String("controller-name", "sealed-secrets-controller", "Name of sealed-secrets controller.")
 	outputFormat   = flag.String("format", "json", "Output format for sealed secret. Either json or yaml")
-	dumpCert       = flag.Bool("fetch-cert", false, "Write certificate to stdout.  Useful for later use with --cert")
+	rotate         = flag.Bool("rotate", false, "Rotate the given key to use the latest private key in the cluster.")
```
anguslees:
```suggestion
	rotate         = flag.Bool("rotate", false, "Re-encrypt the given sealed-secret using the latest controller key.")
```

cmd/kubeseal/main.go
```diff
@@ -236,17 +252,75 @@ func validateSealedSecret(in io.Reader, namespace, name string) error {
...
+	restClient, err := corev1.NewForConfig(conf)
+	if err != nil {
+		return err
+	}
+	req := restClient.RESTClient().Post().
+		Namespace(namespace).
+		Resource("services").
+		SubResource("proxy").
+		Name(net.JoinSchemeNamePort("http", name, "")).
+		Suffix("/v1/keygen")
```
anguslees: Did we remove this endpoint?

pkg/client/listers/.gitignore
```diff
@@ -0,0 +1,24 @@
+# Binaries for programs and plugins
```
anguslees: This file should be removed from the PR

README.md
```diff
@@ -118,93 +118,34 @@ This controller adds a new `SealedSecret` custom resource.  The
...
+The sealed secrets controller will...
+1. Search for these keys and add them to its local store if they are
+labelled as active.
+2. Create a new key
+3. Start the key rotation cycle
+
+#### Key rotation
+
+Keys are automatically rotated. This can be configured on startup with
+the `--rotate-preiod=<value>` flag. `value` is taken as milliseconds with
```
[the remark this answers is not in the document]
anzboi: Of course, whoops.

anzboi:
Thanks for the numerous spelling/grammar checks in the README.

cmd/kubeseal/main.go
anzboi: We did, I forgot to remove this function from kubeseal, thanks for pointing it out.

anzboi: We did, I forgot to remove this function from kubeseal, thanks for pointing it out.

cmd/controller/controller.go
```diff
@@ -303,3 +294,16 @@ func (c *Controller) Rotate(content []byte) ([]byte, error) {
...
 }
+
+func (c *Controller) attemptUnseal(ss *ssv1alpha1.SealedSecret) (*apiv1.Secret, string, error) {
+	return attemptUnseal(ss, c.keyRegistry)
+}
+
+func attemptUnseal(ss *ssv1alpha1.SealedSecret, keyRegistry *KeyRegistry) (*apiv1.Secret, string, error) {
+	for _, privKey := range keyRegistry.keys {
+		if secret, err := ss.Unseal(scheme.Codecs, privKey); err == nil {
+			return secret, "", nil
```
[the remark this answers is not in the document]
anzboi: Yes, that is correct.
The idea was to return the keyName that successfully decrypted a sealed secret. The check above was to avoid re-encryption with the same key a sealed secret was encrypted with. Fixed now.

Makefile
```diff
@@ -43,7 +43,7 @@ kubeseal: $(GO_FILES)
 	$(GO) build -o $@ $(GO_FLAGS) -ldflags "$(GO_LD_FLAGS)" ./cmd/kubeseal
 
 %-static: $(GO_FILES)
-	CGO_ENABLED=0 $(GO) build -o $@ -installsuffix cgo $(GO_FLAGS) -ldflags "$(GO_LD_FLAGS)" ./cmd/$*
+	CGO_ENABLED=0 GOOS=linux GOARCH=amd64 $(GO) build -o $@ -installsuffix cgo $(GO_FLAGS) -ldflags "$(GO_LD_FLAGS)" ./cmd/$*
```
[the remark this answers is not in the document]
anzboi: Not sure I understand the idea here. Make doesn't know whether a pre-requisite compiled file was built with the correct flags, unless I am missing something.

cmd/controller/keyregistry.go
```diff
@@ -0,0 +1,86 @@
...
+	}
+	// Only store key to local store if write to k8s worked
+	kr.registerNewKey(generatedName, key, cert)
+	return generatedName, nil
+}
+
+// blacklistKey deletes a key from the local store and marks the corresponding k8s secret
+// as compromised. This effectively deletes the key from the sealedsecrets controller
+// while the key is still available to admins if need be
+func (kr *KeyRegistry) blacklistKey(keyname string) error {
```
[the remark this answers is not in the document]
anzboi: Indeed

cmd/controller/keys.go
anzboi: No, it is now being used, thanks.

cmd/controller/main.go
anzboi: Thanks for the tip 👍

.gitignore
```diff
@@ -14,11 +14,13 @@
 .glide/
 
 /controller
+/ssadmin
```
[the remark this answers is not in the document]
anzboi: Done

cmd/controller/main.go
anzboi: Fair enough, I've simplified it to just validation and added the default value.

Makefile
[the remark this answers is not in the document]
anguslees: I explained it poorly (and incorrectly).  I meant this:

```patch
--- a/Makefile
+++ b/Makefile
@@ -45,6 +45,8 @@ kubeseal: $(GO_FILES)
 %-static: $(GO_FILES)
        CGO_ENABLED=0 $(GO) build -o $@ -installsuffix cgo $(GO_FLAGS) -ldflags "$(GO_LD_FLAGS)" ./cmd/$*
 
+docker/controller: export GOOS = linux
+docker/controller: export GOARCH = amd64
 docker/controller: controller-static
        cp $< $@
 
```

It's not bulletproof, but assuming `controller-static` is first built through the `docker/controller` dependency (rather than some other dependency path) then it should get the right GOOS/GOARCH.

cmd/controller/main.go
```diff
@@ -24,8 +25,7 @@ import (
 )
 
 var (
-	keyLabelName    = flag.String("key-label", "sealed-secrets-key", "Label used to identify public/private key pairs in k8s.")
-	keyPrefix       = flag.String("key-prefix", "", "Prefix used to name keys. Defaults to label name.")
+	keyPrefix       = flag.String("key-prefix", "sealed-secrets-key-", "Prefix used to name keys. Defaults to label name.")
```
anguslees: Remove "Defaults to label name."

cmd/controller/keys.go
```diff
@@ -0,0 +1,125 @@
...
+
+	data, err := x509.CreateCertificate(r, &cert, &cert, &key.PublicKey, key)
+	if err != nil {
+		return nil, err
+	}
+
+	return x509.ParseCertificate(data)
+}
+
+func blacklistKey(client kubernetes.Interface, namespace, keyname string) error {
```
anguslees: Remove?

cmd/controller/keys.go
```diff
@@ -0,0 +1,125 @@
...
+	"k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/client-go/kubernetes"
+	certUtil "k8s.io/client-go/util/cert"
+)
+
+const SealedSecretsKeyLabel = "sealedsecrets.bitnami.com/sealed-secrets-key"
+
+var (
+	ErrKeyBlacklisted   = errors.New("Key is blacklisted")
```
anguslees: Remove?

cmd/controller/keyregistry.go
anguslees: I think we still need to fix this.

Afaics, we can either:
- sort by creationTimestamp (newest last) when registering existing keys during controller startup, so the last registered is the latest key
- set currentKeyName directly based on tracking the latest, either by a bulk `registerNewKeys()` method, or via a setter called only during controller startup
- we replace this function with a loop that iterates over all the keys and does that comparison every time (it will be cheap, since it's O(10) keys, and the keys are already in memory).  This will also require tracking the creationTimestamp somewhere, since we no longer have the k8s Secret objects at this point, so that's not the least-typing choice.

cmd/controller/server.go
```diff
@@ -15,17 +15,18 @@ import (
 )
 
 var (
+	localAddr    = flag.String("local-addr", ":8081", "trigger rpc serving address.")
```
anguslees: Remove?

anzboi: Done

cmd/controller/keys.go
anzboi: Done

cmd/controller/keys.go
anzboi: Done

cmd/controller/main.go
anzboi: Done

README.md
```diff
@@ -118,29 +118,33 @@ This controller adds a new `SealedSecret` custom resource.  The
...
+#### Key rotation
+
+Keys are automatically rotated. This can be configured on controller startup with
+the `--rotate-period=<value>` flag. The `value` field can be given as golang duration flag (eg: `720h30m`).
+
+A key can be generated early in two ways
+1. Send `SIGUSR1` to the controller
+`kubectl exec -it <controller pod> -- kill -SIGUSR1 1`
+2. Label the current latest key as compromised (any value other than active)
+`kubectl label secrets <keyname> sealed-secrets-key=compromised`.
```
anguslees:
```suggestion
`kubectl label secrets <keyname> sealedsecrets.bitnami.com/sealed-secrets-key=compromised`.
```

anzboi:
Thanks for the lgtm.
As for release notes, that is the only minor breaking change. as long as current keys have that label added, upgraded controller will work with old infrastructure as is.

mkmik:
Sorry for the delay, bitnami went through an acquisition and we scrambled a bit.
The original maintainer left the company; I'm going to fill that role.

Let me see if this branch is still clean and fully understand out the backward compat implications.

README.md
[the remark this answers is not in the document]
mkmik: @anzboi I'd love to reopen the conversation about this point if needed; let's proceed piecewise, this contribution looks already useful in the current shape; merging

mkmik:
ok, sorry for the noise; turns out I can't tell ignore bors to Gus' review, so I'll perform what bors does manually:

1. push to the `staging` branch
2. manually check travis
3. push to master manually

PR #155, opened 2019-04-15
kubeseal: support loading cert from a well-known local file by default

ceralena:
This commit adds support to kubeseal to load the certificate / public
key from a well-known location by default:

	sealed-secrets.crt

We're proposing this change to open a conversation around supporting a
workflow that would be ideal for our developers. We'd like to document a
simple set of commands that each developer or team needs to run in their
local environment and/or CI pipeline to bootstrap the kubeseal command
for generating sealed secrets. In our case, we'd like to simply document
that developers need to run a command line this once in the workspace
where they will be running kubeseal:

```bash
kubeseal --controller-namespace=$FOO --fetch-cert > sealed-secrets.crt

kubeseal < secret.yaml > sealedsecret.yaml
```

The idea here is that if `--cert` is not specified when kubeseal runs,
and `sealed-secrets.crt` exists, it will use that file instead of
reaching out to the cluster - without developers having to specify
`--cert` every time. If we don't do this, then the error they see when
they forget `--cert`, and don't specify `--controller-namespace`, is a
little confusing to people who don't know how the sealed-secrets
controller works or where it runs:

	panic: Error fetching certificate: services "sealed-secrets-controller" not found

	goroutine 1 [running]:
	main.main()
		/home/travis/gopath/src/github.com/bitnami-labs/sealed-secrets/cmd/kubeseal/main.go:216 +0x3b9

Alternatively, for those of us who work on multiple clusters, we'd like
a slightly more advanced usage where an environment variable can be used to manage certificates for different clusters:

```bash
kubeseal --controller-namespace=$FOO --fetch-cert > ~/.sealed-secrets-prod.crt

SEALED_SECRETS_CERT=~/.sealed-secrets-prod.crt kubeseal < secret.yaml > sealedsecret.yaml
```

We're keen to discuss the workflow in more depth or to take another approach.
The idea is to leverage the existing `--fetch-cert` flag, not to break any
existing usage, and to slightly simplify the day-to-day workflow for developers
using kubeseal without knowing about the deployment or architecture of
`sealed-secrets`.

cc/ @yob

### DIFF
--- a/cmd/kubeseal/main.go
+++ b/cmd/kubeseal/main.go
@@ -46,6 +46,10 @@ var (
 	clientConfig clientcmd.ClientConfig
 )
 
+const (
+	defaultlocalCertPath = "sealed-secrets.crt"
+)
+
 func init() {
 	// The "usual" clientcmd/kubectl flags
 	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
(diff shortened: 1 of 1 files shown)


ceralena:
One other thing we wanted to note was that if this is an approach you're happy with, we could consider using `viper` so that all cpnfig supported environment variables as well as CLI flags out of the box, possibly with a prefix as used here (`SEALED_SECRETS_*`).

mkmik:
The problem of offline sealing resonates with me, but I'm not sure if it's a good idea to default to `sealed-secrets.crt` in the current directory though.

What do you think about searching for a `.sealedsecretsrc` up the dir tree (like git finds it's `.git` dir) which would then contain a path relative to that file?
This way you could issue the kubeseal command somewhere in your projects directory tree and have it use the right cert?

But with #178 you can achieve this with `direnv`; I don't know, it simplifies sealed-secrets but it assumes its users do more work.
Any thoughts?

ceralena:
Hey, thanks for the feedback. I think searching up the dir tree like git and vagrant do is a great idea. I'll close this PR for now to get it out of your hair.

PR #161, opened 2019-04-18
explicitly-pass-keyspaces-names

### LINKED ISSUE #157: run sealed-secrets controller for multiple namespaces
Hello everyone,

`Q1` Is there a way to run this controller for multiple namespaces with these conditions:

* I don't wanna run multiple instances of Sealed Secret controller for each namespace.
* I don't wanna run the controller with `ClusterRole` (because in this way it can access all the namespaces).

`Q2` How can I run this controller for a `custom namespace` (by dynamically specifying its name using `env var`) because after looking at the `controller.yaml` [file](https://github.com/bitnami-labs/sealed-secrets/releases/download/$release/controller.yaml), I found out that there is no `env` var is being passed to it except namespace specification in `metatdata` field?

Need help in this regards

Thank you.

aliartiza75:
### Feature
* Sealed secrets controller should accept namespace names as a list and watch only in those namespaces. 

Fixes #157


aliartiza75:
@anguslees kindly review PR description.

mkmik:
Sorry this got lost in time; We now have a section about the helm chart in: https://github.com/bitnami-labs/sealed-secrets#helm-chart

PR #163, opened 2019-04-18
Closes #162 - Add --raw flag for encrypting individual items ad-hoc

### LINKED ISSUE #162: Proposed --raw flag for encrypting individual data
We are looking to adopt sealed-secrets but hesitant because the tool requires an entire secret object as input. This means that developers must either:

- Write out the secret in YAML, unencrypted and in plain text, and then pipe it through the tool to create a sealed secret. If they want to update it later, they have to do this process again, so they will likely just keep the original secret file around on their hard drive
- Create a traditional secret in a Kubernetes cluster, then read it back out and create a sealed secret. They must keep the original secret around in case they want to update it and seal it again.

We prefer not to store secrets at all on developer machines or lower-level environments if possible, so I am working on extending `kubeseal` to support the following use case:

As a developer, I can write out a yaml definition of a `SealedSecret` in my source code:

```yaml
type: Opaque
kind: SealedSecret
metadata:
  name: my-secret
spec:
  encryptedData:
    databasePassword: ''
```

I can then encrypt individual items to put in here with only the public key and the secret namespace/name

```bash
echo "trustno1" | kubeseal --cert cert.pem --raw --namespace myns --name my-secret
```

This outputs the base64-encoded value to add to `encryptedData`. Thus I have managed to create a `SealedSecret` without ever saving the secret value to my hard drive or having sent it to a shared cluster.

tmlbl:
re: #162

### DIFF
--- a/cmd/kubeseal/main.go
+++ b/cmd/kubeseal/main.go
@@ -1,19 +1,23 @@
 package main
 
 import (
+	"crypto/rand"
 	"crypto/rsa"
+	"encoding/base64"
 	"errors"
 	goflag "flag"
 	"fmt"
 	"io"
 	"io/ioutil"
-	"k8s.io/apimachinery/pkg/util/net"
 	"net/http"
 	"os"
 	"strings"
 
+	"k8s.io/apimachinery/pkg/util/net"
+
+	"github.com/bitnami-labs/sealed-secrets/pkg/crypto"
 	flag "github.com/spf13/pflag"
-	"k8s.io/api/core/v1"
+	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"k8s.io/apimachinery/pkg/runtime"
 	runtimeserializer "k8s.io/apimachinery/pkg/runtime/serializer"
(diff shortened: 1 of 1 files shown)


cmd/kubeseal/main.go
```diff
@@ -285,6 +291,34 @@ func main() {
...
+		if err != nil {
+			panic(err.Error())
+		}
+		if ns == "" {
+			panic("Must provide the --namespace flag with --raw")
+		}
+		if *secretName == "" {
+			panic("Must provide the --name flag with --raw")
+		}
+		toEncrypt := make([]byte, base64.StdEncoding.DecodedLen(len(data)))
```
mkmik: actually I don't see why stdin should be base64 encoded

cmd/kubeseal/main.go
```diff
@@ -285,6 +291,34 @@ func main() {
...
+		_, err = base64.StdEncoding.Decode(toEncrypt, data)
+		if err != nil {
+			panic(err.Error())
+		}
+		out, err := crypto.HybridEncrypt(rand.Reader, pubKey, toEncrypt, []byte(fmt.Sprintf("%s/%s", ns, *secretName)))
```
mkmik: we need to make this work in cluster-wide and namespace-wide modes. I'm in the process of refactoring that part anyway.

mkmik:
Continuing this in #257

PR #174, opened 2019-07-01
Switch from alpine to distroless

mkmik:


docker/Dockerfile
```diff
@@ -1,5 +1,5 @@
-FROM alpine
-MAINTAINER Angus Lees <gus@inodes.org>
+FROM gcr.io/distroless/base@sha256:edc3643ddf96d75032a55e240900b68b335186f1e5fea0a95af3b4cc96020b77
```
jjo: Please document which distroless flavor we're using, guess gcr.io/distroless/static ?

mkmik: it's using the `base` flavour (not the `static` flavour):
```
gcr.io/distroless/base@sha256....
                  ^^^^
```

I could have used the static flavour but I have other low hanging fruits

mkmik:
> Changes look good to me, only suggestion is that we can run it as non-root user but as I can see in GCR , the non-root build is not tagged latest(prod ready) so good for now.

yeah, running as non-root is a good idea; will do in a separate PR

docker/Dockerfile
jjo: Ok, even more reasonable to add a comment re:that IMO.

mkmik: PTAL

PR #175, opened 2019-07-01
Test on go1.12

mkmik:
Test on go1.12 and go1.9.
We'll later aim at supporting the two latest versions of go, but for now we'll keep go1.9.

With go 1.9 `./...` now excludes `vendor`.
(We remove the vestigial ksonnet-lib checkout because it's unnecessary and because it breaks with ./...)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: go
 
 go:
-  - '1.10'
+  - '1.12'
   - '1.9'
 
 services:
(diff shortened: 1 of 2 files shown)


mkmik:
> As mentioned in the comment, we would be keeping GO 1.9 for now. Instead, it is changed to 1.12. Rest are changed looks well related to the removal of ksonnet and using GO_PACKAGES = ./... .

1.10 has been changed to 1.12 but 1.9 has been kept

PR #178, opened 2019-07-02
Use flagenv to expose all flags as env vars

mkmik:
Many flags depend on the environment of each user and the project they're working on (like context), and it might apply to several invocations. It makes sense for the user to apply those settings once and have kubeseal just use it (How the users sets the env vars is their business, but I expect some to use something like https://direnv.net/).

Also addresses part of the #155 proposal, CC: @ceralena


cmd/kubeseal/main.go
```diff
@@ -58,6 +66,8 @@ func init() {
 	clientcmd.BindOverrideFlags(&overrides, flag.CommandLine, kflags)
 	clientConfig = clientcmd.NewInteractiveDeferredLoadingClientConfig(loadingRules, &overrides, os.Stdin)
 
+	pflagenv.SetFlagsFromEnv(flagEnvPrefix, flag.CommandLine)
```
atomatt: Could you re-implement `github.com/bitnami-labs/flagenv` to operate on a `FlagSetVisitor` interface (just `Visit` and `VisitAll`, I think) to avoid the `pflagenv` variant fork?

mkmik: Pflag doesn't implement the same visit interface because it reimplements the Flag type.

Perhaps we can write an adapter.

PR #129, opened 2018-11-22
[WIP] Add SealedSecret 'template' to use when constructing Secret

anguslees:
This change adds a `spec.template` to the SealedSecrets schema, which
includes `metadata` and `type` used when constructing the new Secret.
(All the Secret fields _other_ than `data`/`encryptedData`)

Also this change adds:
- Events to the SealedSecret describing progress/errors while
  unsealing.
- A SealedSecret `status` structure that exposes overall
  success/failure.

Fixes #92
Fixes #72

PR #180, opened 2019-07-09
Add SealedSecret 'template' to use when constructing Secret

### LINKED ISSUE #72: Allow a user to know whether the controller actually unsealed a secret
Some users can create/read the sealedsecrets resource kind but they cannot list the secrets on the api server nor tail the logs of the sealed secrets controller (e.g. because of RBAC); it's currently impossible to know whether the secret has been successfully unsealed.

Possible ideas: events and/or a "status" section (like pods, deployments etc)

```
status:
  secret:
    created: True
```

(I was also wondering about the `condition` objects but it seems they might be going away https://github.com/kubernetes/kubernetes/issues/7856)

### LINKED ISSUE #92: Missing annotations & labels

mkmik:
Merge #129 into current codebase (which as diverged in the meantime)

```
This change adds a spec.template to the SealedSecrets schema, which
includes metadata and type used when constructing the new Secret.
(All the Secret fields other than data/encryptedData)

Also this change adds:

Events to the SealedSecret describing progress/errors while
unsealing.
A SealedSecret status structure that exposes overall
success/failure.
```

Fixes #92
Fixes #72


atomatt:
bors r+

based on a very superficial review :crossed_fingers:

mkmik:
bors delegate=atomatt

bric3:
Could be labelled as 0.8.0 as well :)

mkmik:
> Could be labelled as 0.8.0 as well :)

To avoid overcrowding the milestone "changelog" I tended to favour labelling tracking issues (such as #92 in this case) over PRs. My mind can easily be changed with arguments :-)

bric3:
@mkmik I had the opposite practice as it allows to track which work get into which milestone, after several years I found it useful. But your point is just as fine it just address different concerns ;)

PR #186, opened 2019-07-17
Use official docker image in main Makefile

mkmik:
Our CI system pushes the latest build to the `:latest` tag on quay.
It seems natural for users to desire that the controller.yaml file that gets built in master
points to such a build

(see https://github.com/bitnami-labs/sealed-secrets/issues/86#issuecomment-512087928)


mkmik:
CC: @demisx

PR #191, opened 2019-07-19
Generate new key only if needed

### LINKED ISSUE #190: Generate new key only if needed
The key rotation feature is still experimental and incomplete.
Key rotation has to happen only when users explicitly request that feature.

An example of a  blocking issues that prevents key rotation to be ready for the average user is #185.

Since #187, the timer that triggers periodic rotation is disabled by default, but we still trigger the key rotation at every controller boot.

mkmik:
Closes #190

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -98,10 +98,19 @@ func myNamespace() string {
 	return metav1.NamespaceDefault
 }
 
-// Initialises the first key and starts the rotation job. returns an early trigger function
+// Initialises the first key and starts the rotation job. returns an early trigger function.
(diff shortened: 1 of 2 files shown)


cmd/controller/main.go
```diff
@@ -98,10 +98,19 @@ func myNamespace() string {
 	return metav1.NamespaceDefault
 }
 
-// Initialises the first key and starts the rotation job. returns an early trigger function
+// Initialises the first key and starts the rotation job. returns an early trigger function.
+// A period of 0 disables automatic rotation, but manual rotation (e.g. triggered by signal) is still
```
jjo: Can you clarify what you mean by "signal" ?

mkmik: PTAL

PR #195, opened 2019-07-24
Document Go mod aware installation instructions

mkmik:


README.md
```diff
@@ -63,9 +63,15 @@ If you just want the latest client tool, it can be installed into
 `$GOPATH/bin` with:
 
 ```sh
-% go get github.com/bitnami-labs/sealed-secrets/cmd/kubeseal
+% (cd /tmp; GO111MODULE=on go get github.com/bitnami-labs/sealed-secrets/cmd/kubeseal@master)
```
atomatt: Is the `cd /tmp` relevant?

mkmik: It updates the `go.{mod,sum}` files if invoked inside an existing module. So I picked a dir that usually isn't a go module working dir; perhaps `/` is clearer?)

I tried `GO111MODULE=on go get -mod=readonly github.com/bitnami-labs/sealed-secrets/cmd/kubeseal@master` but it still updates `go.{mod,sum}`

atomatt: Would `GOFLAGS="-mod=readonly"` help?

mkmik: tried that too; it behaves as when passing the flag explicitly, i.e. it has no effect

mkmik: tools like [gobin](https://github.com/myitcv/gobin) exist to simplify this, but then you have to install that first....

atomatt: Well, that's all quite annoying, but I would not add `gobin` to the mix just yet. `cd /tmp` is fine.

PR #202, opened 2019-07-25
Writable /tmp

### LINKED ISSUE #200: readonly /tmp
`klog` (the glog fork used by kubernetes) commits suicide when trying to write an error

```
log: exiting because of error: log: cannot create log: open /tmp/controller.sealed-secrets-controller-665d687cc6-kjgl5.unknownuser.log.ERROR.20190725-153353.1: read-only file system
```

mkmik:
Our deployment template uses the `readOnlyRootFilesystem` but the k8s `klog` library
likes to write logs in `/tmp`.

Let's mount a writable `/tmp` dir.

Closes #200

### DIFF
--- a/controller-norbac.jsonnet
+++ b/controller-norbac.jsonnet
@@ -53,6 +53,16 @@ local v1beta1_Deployment(name) = kube.Deployment(name) {
                 runAsNonRoot: true,
                 runAsUser: 1001,
               },
+              volumeMounts_+: {
+                tmp: {
+                  mountPath: '/tmp',
+                },
+              },
+            },
+          },
+          volumes_+: {
+            tmp: {
+              emptyDir: {},
             },
           },
         },


controller-norbac.jsonnet
```diff
@@ -53,6 +53,19 @@ local v1beta1_Deployment(name) = kube.Deployment(name) {
                 runAsNonRoot: true,
                 runAsUser: 1001,
               },
+              volumeMounts_+: {
+                tmp: {
```
jjo: As we're using kube.libsonnet, I'd suggest not explicitly setting the `name` field but rather:
```
volumeMounts_+: {
  tmp_volume: {
    mountPath: '/tmp',
  },
},
```
then modify below to be just:
```
volumes_+: {
  tmp_volume: kube.EmptyDirVolume(),
},
```

mkmik: @jjo thanks for the name field! I'm not a fan of hiding `empty{}` behind `kube.EmptyDirVolume()` though

PR #203, opened 2019-07-25
use string for encryptedData

### LINKED ISSUE #201: Endless loop in controller on invalid base64 encrypted data
It seems that improperly encoded base64 encryptedData causes an endless loop in the controller. For example:
```
kind: SealedSecret
metadata:
  name: brokenss
spec:
  encryptedData:
    foo: foo1foo
    bar: bar1bar
```

As soon as this kind of broken sealed-secret is added, controller will log each second a new message like this:
```
E0725 15:10:11.704525       1 reflector.go:205] github.com/bitnami-labs/sealed-secrets/cmd/controller/controller.go:155: Failed to list *v1alpha1.SealedSecret: v1alpha1.SealedSecretList.Items: []v1alpha1.SealedSecret: v1alpha1.SealedSecret.Spec: v1alpha1.SealedSecretSpec.EncryptedData: decode base64: illegal base64 data at input byte 4, error found in #10 byte of ...|:"bar1bar","foo":"fo|..., bigger context ...|0000001"},"spec":{"encryptedData":{"bar":"bar1bar","foo":"foo1foo"}}},{"apiVersion":"bitnami.com/v1a|...
```

Unfortunately, this blocks the processing of any other sealed-secret added afterwards. The only way to break the loop is to delete the broken sealed-secret.

I tried looking in the code for a way to catch the base64-decoding error and consider the sealed-secret broken, as when the key doesn't match. But it seems this happens too deep in the call stack. The only way I found to "fix" the problem was to change the type of the `encryptedData` field in the `SealedSecretSpec struct` to `string` instead of `[]byte` so that we control the base64-decoding. You can find [here](https://github.com/bitnami-labs/sealed-secrets/compare/master...catac:fix-loop) the full set of changes for this.

Please let me know if this approach is ok for you and I'll be happy to adapt the code to your comments and come back with a PR. Or maybe you see a better approach?

Thanks!

mkmik:
Closes #201

### DIFF
--- a/pkg/apis/sealed-secrets/v1alpha1/sealedsecret_expansion.go
+++ b/pkg/apis/sealed-secrets/v1alpha1/sealedsecret_expansion.go
@@ -3,6 +3,7 @@ package v1alpha1
 import (
 	"crypto/rand"
 	"crypto/rsa"
+	b64 "encoding/base64"
 	"fmt"
 
 	v1 "k8s.io/api/core/v1"
@@ -97,7 +98,7 @@ func NewSealedSecret(codecs runtimeserializer.CodecFactory, pubKey *rsa.PublicKe
 				// ObjectMeta copied below
 				Type: secret.Type,
 			},
-			EncryptedData: map[string][]byte{},
+			EncryptedData: map[string]string{},
 		},
 	}
 	secret.ObjectMeta.DeepCopyInto(&s.Spec.Template.ObjectMeta)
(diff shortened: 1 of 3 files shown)


mkmik:
@catac  do you have the BW to ensure the tests pass?

pkg/apis/sealed-secrets/v1alpha1/sealedsecret_expansion.go
```diff
@@ -3,6 +3,7 @@ package v1alpha1
 import (
 	"crypto/rand"
 	"crypto/rsa"
+	b64 "encoding/base64"
```
atomatt: Is there any reason to rename the import here?

mkmik: addressed in #206

mkmik:
thanks @catac, I built on top of your fix in #206

PR #205, opened 2019-07-26
Add mod=vendor to build

mkmik:


mkmik:
> They need to make this the default already :)

indeed: https://github.com/golang/go/issues/27227

mkmik:
@atomatt PTAL

PR #206, opened 2019-07-26
Fix endless loop when a single sealedsecret has a base64 decode error

### LINKED ISSUE #201: Endless loop in controller on invalid base64 encrypted data

mkmik:
Closes #201 

Supersedes #203, adding fixes to tests and addressing review comments (created new PR because I cannot add commits in that fork)

### DIFF
--- a/integration/controller_test.go
+++ b/integration/controller_test.go
@@ -8,6 +8,7 @@ import (
 	"crypto/rand"
 	"crypto/rsa"
 	"crypto/x509"
+	"encoding/base64"
 	"fmt"
 	"io"
 	"sort"
@@ -191,7 +192,7 @@ var _ = Describe("create", func() {
 
 		Context("With renamed encrypted keys", func() {
 			BeforeEach(func() {
-				ss.Spec.EncryptedData = map[string][]byte{
+				ss.Spec.EncryptedData = map[string]string{
 					"xyzzy": ss.Spec.EncryptedData["foo"],
 				}
 			})
@@ -212,7 +213,7 @@ var _ = Describe("create", func() {
 				ciphertext, err := crypto.HybridEncrypt(rand.Reader, pubKey, []byte("new!"), []byte(label))
 				Expect(err).NotTo(HaveOccurred())
 
-				ss.Spec.EncryptedData["foo2"] = ciphertext
+				ss.Spec.EncryptedData["foo2"] = base64.StdEncoding.EncodeToString(ciphertext)
 			})
 			It("should produce expected Secret", func() {
 				expected := map[string][]byte{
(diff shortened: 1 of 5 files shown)


alexellis:
Hi @bakins, this appears to be a breaking change in the API, and yet the api version wasn't bumped?

We had a PR sent for OpenFaaS Cloud, but it wasn't tested properly by the author and has broken our pipeline now. We're trying to figure out what's wrong.

Would you mind taking a quick look and pointing us in the right direction? 

https://github.com/openfaas/openfaas-cloud/commit/0bd4c834fab25ec6a65a4f015a1271d00a81cbf1#diff-fa778be7508cf2401eb85fe1d5f4dece

We end up with a weird JSON value in the secret instead of a working pipeline.

If you had time, it would be most appreciated. Thank you.

mkmik:
were you using sealed-secrets as a library?

alexellis:
We are using your clientset in order to import SealedSecrets from a CI pipeline.

mkmik:
sorry about the breaking change then; I didn't know. We're focusing on not breaking wire format compatibility for end users interacting with k8s manifests in the normal way.

I'll keep it in mind that we have direct dependencies to the pkg

mkmik:
I think we could find an alternative way to deal with base64 encoding error handling. how bad is this breakage for you?

alexellis:
Very bad, but it looks like I've got it back to a working state now in the linked PR over on our repo.

I also saw a mention that "data" will be deprecated/go-away, is there any issue that we could track for that? Thank you for responding so quickly.

mkmik:
> I also saw a mention that "data" will be deprecated/go-away, is there any issue that we could track for that?

No; and I'm taking this very slowly; there are still tons of users on <=0.7, so this thing is not going away any time  soon (possibly ever)

PR #208, opened 2019-07-31
Allow access to sealed secret services/proxy to any authenticated user

### LINKED ISSUE #166: Install in an isolated namespace
I'm attempting to install sealed secrets helm chart into a namespace where my user is "admin" at the role level not the cluster level.  I was hoping to install sealed secrets in this namespace to isolate the users from needing access to other namespaces/secrets.

Am I being a numbskull here?

```---
kind: Role
apiVersion: rbac.authorization.k8s.io/v1beta1
metadata:
  name: {{ item.namespace }}-manager
  namespace: {{ item.namespace }}
rules:
- apiGroups:
  - "*"
  resources:
  - "*"
  verbs:
  - "*"
```

```
---
kind: RoleBinding
apiVersion: rbac.authorization.k8s.io/v1beta1
metadata:
  name: {{ item.namespace }}-binding
  namespace: {{ item.namespace }}
subjects:
{% for item in item.users | default([]) %}
- kind: User
  name: {{ item }}
  apiGroup: rbac.authorization.k8s.io
{% endfor -%}
roleRef:
  kind: Role
  name: {{ item.namespace }}-manager
  apiGroup: rbac.authorization.k8s.io
```

mkmik:
This allows kubeseal to fetch the certificate public key (and perform other actions such as /verify and /rotate endpoints) even if the caller doesn't have otherwise the rights to access the kube-system namespace (or any other namespace where the sealed-secrets controller might have been deployed), as it often happens that users are not granted such broad permissions on production clusters.

We historically suggested users to just distribute the certificate out of bound and use the `--cert` flag.
However, with the advent of master key rotation, this is becoming increasingly more cumbersome, especially since
it's critical that users end up using the right certificate (i.e. the certificate has to be authenticated).
Master key rotation also requires users to periodically rotate the secrets, which requires access to the /rotate endpoint.

This change includes a fine-grained RBAC rule that allows access to the sealed-secrets controller HTTP API to any authenticated user in the cluster.
Users are still free to disable this feature by applying an override during deployment, but our default RBAC config should include it.

The controller currently exposes the following endpoints:

- `/healthz'
- `/v1/verify`
- `/v1/rotate`
- `/v1/cert.pem`

The controller already must not expose any secrets via the HTTP endpoint, since while RBAC would prevent
end-users to access the service via the proxy, nothing prevents any unprivileged workload in the cluster unless
admins have explicitly configured a strict network policy rule set.

Closes #166
Rel #137

### DIFF
--- a/controller.jsonnet
+++ b/controller.jsonnet
@@ -38,6 +38,26 @@ controller {
     ],
   },
 
+  serviceProxierRole: kube.Role('sealed-secrets-service-proxier') + $.namespace {
+    rules: [
+      {
+        apiGroups: [
+          '',
+        ],
+        resources: [
+          'services/proxy',
+        ],
+        resourceNames: [
+          'http:sealed-secrets-controller:',  // kubeseal uses net.JoinSchemeNamePort when crafting proxy subresource URLs
+          'sealed-secrets-controller',  // but often services are referred by name only, let's not make it unnecessarily cryptic
+        ],
+        verbs: [
+          'get',
+        ],
+      },
+    ],
+  },
+
   unsealerBinding: kube.ClusterRoleBinding('sealed-secrets-controller') {
     roleRef_: $.unsealerRole,
     subjects_+: [$.account],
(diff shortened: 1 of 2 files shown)


controller.jsonnet
```diff
@@ -38,6 +38,26 @@ controller {
...
+      {
+        apiGroups: [
+          '',
+        ],
+        resources: [
+          'services/proxy',
+        ],
+        resourceNames: [
+          'http:sealed-secrets-controller:',  // kubeseal uses net.JoinSchemeNamePort when crafting proxy subresource URLs
+          'sealed-secrets-controller',  // but often services are referred by name only, let's not make it unnecessary cryptic
```
atomatt: s/unnecessary/unnecessarily

mkmik:
@atomatt as I mentioned in the PR description, even before this RBAC change leaking secrets was a terrible idea. That said, it doesn't hurt to add such a warning in the code

PR #215, opened 2019-08-06
Honour creation time during rotation

mkmik:
Instead of rotating every period seconds after the pod boots, take into consideration
the timestamp of the last secret and create the next one only when it gets stale.

Ref #190

Also tested by setting rotation period 5m and then periodically killing the pod
and checking that it indeed rotated the keys only after 5m.

```
while true; do ke -n kube-system get pod -lname=sealed-secrets-controller -oname |xargs ke -n kube-system delete; ke -n kube-system get secret -l sealedsecrets.bitnami.com/sealed-secrets-key=active; sleep 10s; done
```

### DIFF
--- a/cmd/controller/keys.go
+++ b/cmd/controller/keys.go
@@ -54,7 +54,19 @@ func readKey(secret v1.Secret) (*rsa.PrivateKey, []*x509.Certificate, error) {
 	}
 }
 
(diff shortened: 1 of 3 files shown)


mkmik:
please review but DO NOT MERGE, depends on #214

Issue #185, opened 2019-07-16
Key Rotation breaks appending new values to existing sealed secrets


chrisharm:
With new new key rotation feature introduced in #143, the ability to append new values to an existing sealed secret is broken.  Each time the controller restarts, a new key is generated and set as the active key.  It looks like the unseal process expects all of the keys in a SealedSecret object to be encrypted using the same key.  

Steps to Reproduce:

1. Encrypt foo=bar: 
`kubectl create secret generic mysecret --dry-run --from-literal=foo=bar -o yaml | kubeseal --format yaml` 

Output:
```
apiVersion: bitnami.com/v1alpha1
kind: SealedSecret
metadata:
  creationTimestamp: null
  name: mysecret
  namespace: default
spec:
  encryptedData:
[...]

PR #216, opened 2019-08-06
Handle items sealed with different keys

mkmik:
Ref #185

This is not the best implementation, but it technically unblocks #185 and lets us incrementally implement a better solution (the current plan is to make kubeseal encode the fingerprint of the public key used to seal the secret in each item).

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -1,6 +1,7 @@
 package main
 
 import (
+	"crypto/rsa"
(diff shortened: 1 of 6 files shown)


mkmik:
Please review but DO NOT MERGE; this is a layered commit on top of #215

mkmik:
@atomatt I was lazy. Will do in another PR. The existing tests at least ensure it doesn't regress

PR #218, opened 2019-08-07
Ensure /v1/cert.pem endpoint reads certificates after rotation

mkmik:
#143 introduced key rotation, but we didn't update the cert.pem serving path

### DIFF
--- a/cmd/controller/server_test.go
+++ b/cmd/controller/server_test.go
@@ -1,0 +1,87 @@
+package main
+
+import (
+	"context"
+	"crypto/x509"
+	"fmt"
+	"io/ioutil"
+	"net/http"
+	"strings"
+	"sync"
+	"testing"
+	"time"
+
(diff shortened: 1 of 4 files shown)


cmd/controller/server_test.go
```diff
@@ -0,0 +1,58 @@
...
+	"io/ioutil"
+	"net/http"
+	"strings"
+	"testing"
+	"time"
+
+	certUtil "k8s.io/client-go/util/cert"
+)
+
+func TestHttpCert(t *testing.T) {
```
atomatt: I see this is a new file so I assume there were no tests for the `/v1/cert.pem` endpoint until now? Thanks for adding one!

However, should there be a distinct test that replaces `cert` once the server is running (simulating a key rotation) and checks the latest cert is returned?

mkmik: thanks for encouraging me to write a proper test :-p

Issue #104, opened 2018-06-16
Add support for stringData


wmedlar:
The [Secret spec](https://kubernetes.io/docs/reference/generated/kubernetes-api/v1.10/#secret-v1-core) allows for non-binary data to be provided in the `stringData` key as a raw string, rather than base64-encoded, for convenience. Most of my secrets are just plain strings, so it'd be awesome to add support for this to kubeseal.

Currently:
* with no `data` key kubeseal panics:
```shell
$ cat secret-no-data.yaml
apiVersion: v1
kind: Secret
metadata:
  name: string-data-secret
  namespace: default
stringData:
  foo: baz

$ kubeseal < secret-no-data.yaml
panic: Secret.data is empty in input Secret, assuming this is an error and aborting

goroutine 1 [running]:
main.main()
	/Users/travis/gopath/src/github.com/bitnami-labs/sealed-secrets/cmd/kubeseal/main.go:233 +0x2a8
```

* with a `data` key kubeseal creates a SealedSecret that decrypts to the key-value pair `foo: bar`
```shell
$ cat secret.yaml
apiVersion: v1
kind: Secret
metadata:
  name: string-data-secret
  namespace: default
data:
  foo: YmFy # bar
stringData:
  foo: baz

$ kubeseal < secret.yaml | kubectl apply -f -
sealedsecret "string-data-secret" created

$ kubectl get secret string-data-secret -o jsonpath="{['data']['foo"]}" | base64 -D
bar
```

In both cases I would expect it to create a SealedSecret that decrypts to the key-value pair `foo: baz`.

Renz2018:
It's very useful!

hpurmann:
I have worked around that issue by [manually base64-encoding the yaml values below the `data` keys](https://github.com/actano/transform-yaml/blob/a6e8965220c86f5dbcf256615ed9391b34f6c252/example/run.sh)

At my company, we are using go template files to specify [from which Vault path to read secrets from](https://github.com/actano/vault-template). A script renders the templates into secrets and seals them. Because `sealed-secrets` don't yet understand `stringData`, we put this manual base64-encoding step in between. It's kind of ugly but works in the meantime.

If of interest, I could share this script.

Renz2018:
@hpurmann
Can you share your code with me?Thanks

hpurmann:
[...]

PR #221, opened 2019-08-10
Accept and seal stringData into secret

pinkavaj:
Ref #104

### DIFF
--- a/cmd/kubeseal/main.go
+++ b/cmd/kubeseal/main.go
@@ -167,7 +167,7 @@ func seal(in io.Reader, out io.Writer, codecs runtimeserializer.CodecFactory, pu
 		return err
 	}
 
-	if len(secret.Data) == 0 {
+	if len(secret.Data) == 0 && len(secret.StringData) == 0 {
 		// No data. This is _theoretically_ just fine, but
 		// almost certainly indicates a misuse of the tools.
 		// If you _really_ want to encrypt an empty secret,
--- a/cmd/kubeseal/main_test.go
+++ b/cmd/kubeseal/main_test.go
@@ -133,6 +133,9 @@ func TestSeal(t *testing.T) {
 		Data: map[string][]byte{
 			"foo": []byte("sekret"),
 		},
+		StringData: map[string]string{
+			"foos": string("stringsekret"),
+		},
 	}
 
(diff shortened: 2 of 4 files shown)


cmd/kubeseal/main.go
```diff
@@ -167,6 +167,16 @@ func seal(in io.Reader, out io.Writer, codecs runtimeserializer.CodecFactory, pu
 		return err
 	}
 
+	if len(secret.StringData) > 0 {
+		if secret.Data == nil {
+			secret.Data = map[string][]byte{};
+		}
+		for key, values := range secret.StringData {
+			secret.Data[key] = []byte(values)
+			delete(secret.StringData, key)
```
Crevil: What is the reasoning for deleting the values? Just curious. :)

pinkavaj: It is my obsession with doing thinks right. When for example someone latter would serialize this object it will get duplicate fileds in both `data` and `stringData` sections if delete is not done. I agree it is very improbable and actualy technicaly unnecessary.

mkmik: When I was first thinking of this issue I though we should make kubeseal take a secret that contains stringData as input, produce a sealed secret that once converted into a secret would be identical to the original, namely keeping the stringData intact. This means we would have to somehow encode that in the sealed secret CRD schema.

But this approach seems pragmatic enough. I wonder how would backwards compatibility look like if we later decide that reversibility is important.

Crevil: Makes good sense. Could a test for this maybe be usefull for future reference? So the behaviour becomes very explicit?

pinkavaj: I'm not sure I can write right now more complicated test suite in go (this is my second commit in go ... :). But if You give me a hint I will try.

pinkavaj: Actually kubectl does the same operation, data and stringData cannot be distinguished once pushed into Kubernetes api.

mkmik: Ah interesting, didn't know (fwiw I'm not fond of this kind of magic in k8s...).

Happy to merge when we have some tests

pinkavaj: I have managed to dig this think a bit deeper and add some test, please check if it is enought.

pinkavaj: The TestSealRoundTripStringDataConversion is more or less duplicate of TestSealRoundTrip, so mey be keep only one of those?

PR #225, opened 2019-08-28
add support for jenkins k8s credential provider plugin

peterbosalliandercom:
My 2 cents to support the annotations & labels of the credential provider plugin (https://github.com/jenkinsci/kubernetes-credentials-provider-plugin) . This is very useful in combination with argocd -> helm templating -> sealedsecrets used by a jenkins which uses the credential provider to automatically load credentials. Devops all the way.


Dockerfile
```diff
@@ -0,0 +1,15 @@
+FROM golang:1.11-alpine as builder
```
mkmik: we have a dockerfile in docker/Dockefile which we use in our official builds

peterbosalliandercom:
Thanks for creating support for annotations & labels, this PR is now redundant. We needed this on 0.7.0 and build it ourselves. Now it is generally supported, thanks for the good work!

PR #228, opened 2019-08-29
Strip strategic merge patch annotations

### LINKED ISSUE #227: stringData can leak in last-applied-configuration
We're using kubeseal v0.8.1 now, with support for metadata annotations, which is nice for secrets like [jenkins-credentials](https://jenkinsci.github.io/kubernetes-credentials-provider-plugin/examples/). But due to this change the `kubectl.kubernetes.io/last-applied-configuration` annotation also gets added. And in my way of using kubeseal the stringData will leak into the sealed secret this way.

**Reproduce issue (my workflow)**

I typically first create a secret with stringData which is convenient. I apply it, so I can test it. Then I finally read it out from the cluster, feed it through kubeseal and commit it to git. This ensures me that the `stringData` has been converted to `data` and the secret actually works. And this workflow produces this issue. Describe in more detail below:

1. Create a secret with `stringData` property rather than `data` property for convenience in a file.

    `mysecret.yaml`

    ```apiVersion: v1
    stringData:
      password: secretpassword
      username: username
    kind: Secret
    metadata:
      name: "mysecret"
      namespace: default
      labels:
        "jenkins.io/credentials-type": "usernamePassword"
      annotations:
        "jenkins.io/credentials-description" : "mysecret"
    type: Opaque
    ```

2. Apply the secret to the cluster.

    `kubectl apply -f mysecret.yaml`

3. The secret in the cluster now has the `stringData` property in the `spec:template:metadata:annotations:kubectl.kubernetes.io/last-applied-configuration`.
4. Checkout the secret and seal with kubeseal

    `kubectl get secret mysecret -o json | kubeseal --cert cert.pem > sealed-secret.json`

5. Now the sealed-secret.json contains the secret in plaintex in the metadata:

    ``` {
      "kind": "SealedSecret",
      "apiVersion": "bitnami.com/v1alpha1",
      "metadata": {
        "name": "mysecret",
        "namespace": "default",
        "creationTimestamp": null
      },
      "spec": {
        "template": {
          "metadata": {
            "name": "mysecret",
            "namespace": "default",
            "creationTimestamp": null,
            "labels": {
              "jenkins.io/credentials-type": "usernamePassword"
            },
            "annotations": {
              "jenkins.io/credentials-description": "mysecret",
              "kubectl.kubernetes.io/last-applied-configuration": "{\"apiVersion\":\"v1\",\"kind\":\"Secret\",\"metadata\":{\"annotations\":{\"jenkins.io/credentials-description\":\"mysecret\"},\"labels\":{\"jenkins.io/credentials-type\":\"usernamePassword\"},\"name\":\"mysecret\",\"namespace\":\"default\"},\"stringData\":{\"password\":\"secretpassword\",\"username\":\"username\"},\"type\":\"Opaque\"}\n"
            }
          },
          "type": "Opaque"
        },
        "encryptedData": {
          "password": "...",
          "username": "..."
        }
      },
      "status": {
    
      }
    }
    ```
6. The resulting secret is not safe to share or check into git.

*Probably the same will happen for the `data` property, which is just the base64 encoded version of the stringData*

**What should happen**

I think the best way of dealing with this issue is by **stripping the last-applied-configuration when ingesting a secret.**

mkmik:
Closes #227

### DIFF
--- a/pkg/apis/sealed-secrets/v1alpha1/sealedsecret_expansion.go
+++ b/pkg/apis/sealed-secrets/v1alpha1/sealedsecret_expansion.go
@@ -80,6 +80,19 @@ func NewSealedSecretV1(codecs runtimeserializer.CodecFactory, pubKey *rsa.Public
 	return s, nil
 }
 
+func StripLastAppliedAnnotations(annotations map[string]string) {
+	if annotations == nil {
+		return
+	}
+	keys := []string{
+		"kubectl.kubernetes.io/last-applied-configuration",
+		"kubecfg.ksonnet.io/last-applied-configuration",
+	}
+	for _, k := range keys {
+		delete(annotations, k)
+	}
+}
+
 // NewSealedSecret creates a new SealedSecret object wrapping the
 // provided secret. This encrypts only the values of each secrets
 // individually, so secrets can be updated one by one.
@@ -102,6 +115,13 @@ func NewSealedSecret(codecs runtimeserializer.CodecFactory, pubKey *rsa.PublicKe
 		},
 	}
 	secret.ObjectMeta.DeepCopyInto(&s.Spec.Template.ObjectMeta)
+
+	// the input secret could come from a real secret object applied with `kubectl apply` or similar tools
+	// which put a copy of the object version at application time in an annotation in order to support
+	// strategic merge patch in subsequent updates. We need to strip those annotations or else we would
+	// be leaking secrets in clear in a way that might be non obvious to users.
+	// See https://github.com/bitnami-labs/sealed-secrets/issues/227
+	StripLastAppliedAnnotations(s.Spec.Template.ObjectMeta.Annotations)
 
 	// RSA-OAEP will fail to decrypt unless the same label is used
 	// during decryption.
--- a/pkg/apis/sealed-secrets/v1alpha1/sealedsecret_test.go
+++ b/pkg/apis/sealed-secrets/v1alpha1/sealedsecret_test.go
@@ -411,6 +411,49 @@ func TestSealRoundTripWithMisMatchNamespaceWide(t *testing.T) {
 	}
 }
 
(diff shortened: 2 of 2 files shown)


pkg/apis/sealed-secrets/v1alpha1/sealedsecret_expansion.go
```diff
@@ -80,6 +80,19 @@ func NewSealedSecretV1(codecs runtimeserializer.CodecFactory, pubKey *rsa.Public
 	return s, nil
 }
 
+func StripLastAppliedAnnotations(annotations map[string]string) {
+	if annotations == nil {
+		return
+	}
+	keys := []string{
+		"kubectl.kubernetes.io/last-applied-configuration",
+		"kubecfg.ksonnet.io/last-applied-configuration",
```
nicorikken: Good catch!

mkmik: do you know any other tool that does something similar?

nicorikken: Not that I know of.
Perhaps make a regex on `*last-applied-configuration` ?

mkmik: I suspect I'd regret that choice when somebody would report bewilderment at a stripped annotation with a similar name but with a different semantics.

atomatt: Perhaps check all annotations and strip anything that contains original plain text?

mkmik: that wouldn't work with `"kubecfg.ksonnet.io/last-applied-configuration` since it's gzip compressed

PR #230, opened 2019-08-29
Update the install guide and add instructions for GKE private cluster

bzon:
I just tried installing kubeseal today and I encountered issues when running the cli against my private GKE cluster. As an addition, I also refined the install guide in the README.md. Let me know if that is needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -17,23 +17,91 @@ original Secret from the SealedSecret.
 See https://github.com/bitnami-labs/sealed-secrets/releases for the latest
 release.
 
-```sh
+
+### GKE Requirements
+
+If installing on a GKE cluster, a ClusterRoleBinding may be needed to successfully deploy the controller in the final command.  Replace <your-email> with a valid email, and then deploy the cluster role binding:
+ 
+```bash
+USER_EMAIL=<your-email>
+kubectl create clusterrolebinding $USER-cluster-admin-binding --clusterrole=cluster-admin --user=$USER_EMAIL
+```
+ 
+If you are using a **private GKE cluster**, you are required to create a Master-to-Node firewall rule to allow GKE to communicate to the kubeseal container endpoint port tcp/8080.
+ 
+```bash
+CLUSTER_NAME=foo-cluster
+gcloud config set compute/zone your-zone-or-region
+```
+
+Get the `MASTER_IPV4_CIDR`.
+
+```bash
+MASTER_IPV4_CIDR=$(gcloud container clusters describe $CLUSTER_NAME \
+  | grep "masterIpv4CidrBlock: " \
+  | awk '{print $2}')
+```
+
+Get the `NETWORK`.
+
+```bash
+NETWORK=$(gcloud container clusters describe $CLUSTER_NAME \
+  | grep "^network: " \
+  | awk '{print $2}')
+```
+
+Get the `NETWORK_TARGET_TAG`.
+
+```bash
+NETWORK_TARGET_TAG=$(gcloud compute firewall-rules list \
+  --filter network=$NETWORK --format json \
+  | jq ".[] | select(.name | contains(\"$CLUSTER_NAME\"))" \
+  | jq -r '.targetTags[0]' | head -1)
+```
+
+Check the values.
+
+```bash
+echo $MASTER_IPV4_CIDR $NETWORK $NETWORK_TARGET_TAG
+
+# example output
+10.0.0.0/28 foo-network gke-foo-cluster-c1ecba83-node
+```
+
+Create the firewall rule.
+
+```bash
+gcloud compute firewall-rules create gke-to-kubeseal-8080 \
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -17,23 +17,91 @@ original Secret from the SealedSecret.
...
+```bash
 $ release=$(curl --silent "https://api.github.com/repos/bitnami-labs/sealed-secrets/releases/latest" | sed -n 's/.*"tag_name": *"\([^"]*\)".*/\1/p')
 
-# Install client-side tool into /usr/local/bin/
-$ GOOS=$(go env GOOS)
-$ GOARCH=$(go env GOARCH)
-$ wget https://github.com/bitnami-labs/sealed-secrets/releases/download/$release/kubeseal-$GOOS-$GOARCH
-$ sudo install -m 755 kubeseal-$GOOS-$GOARCH /usr/local/bin/kubeseal
+# See the release binary versions at https://github.com/bitnami-labs/sealed-secrets/releases/
+RELEASE_BIN_URL=https://github.com/bitnami-labs/sealed-secrets/releases/download/v0.8.3/kubeseal-linux-amd64
```
mkmik: I see two problems with this change in the instructions:

1. This has to be updated for every release (and we risk forgetting it)
2. Not everybody runs linux on amd64

To deal with (1) so far I opted for just inlining installation instructions in each release notes (so that they include the correct version).

I assume that you prefer ditching `go env ...` because not everybody has the go toolchain installed.
What if we just point to the latest release page here?

README.md
```diff
@@ -17,23 +17,91 @@ original Secret from the SealedSecret.
...
+### GKE Requirements
+
+If installing on a GKE cluster, a ClusterRoleBinding may be needed to successfully deploy the controller in the final command.  Replace <your-email> with a valid email, and then deploy the cluster role binding:
+ 
+```bash
+USER_EMAIL=<your-email>
+kubectl create clusterrolebinding $USER-cluster-admin-binding --clusterrole=cluster-admin --user=$USER_EMAIL
+```
+ 
+If you are using a **private GKE cluster**, you are required to create a Master-to-Node firewall rule to allow GKE to communicate to the kubeseal container endpoint port tcp/8080.
```
mkmik: uh, why is that needed? to get the proxy to work (and thus fetch the certificate?)

README.md
bzon: > I assume that you prefer ditching go env ... because not everybody has the go toolchain installed.

Yes!

> What if we just point to the latest release page here?

I think that is a better idea to avoid having to maintain the main README. What do you think about the one-liner curl install though?

README.md
bzon: Yes, when running kubeseal to get the public certificate I had to add a firewall rule to make this work. This is actually related to [this](https://linkerd.io/2/reference/cluster-configuration/#private-clusters) which is a common issue in GKE private clusters.

mkmik: I'm considering changing the way we advertize the pubkey.
Currently the controller serves the certificate and the client uses the api server proxy feature to talk directly with the controller.

I wanted to change this so that the certificate gets written into a resource (e.g. a config-map) which would be made readable to all authenticated users. `kubeseal` would then have only to talk to the API server.

I'm not against documenting a useful tip&trick for GKE, but I'd rather solve this in a less invasive way.

That said, in the meantime would the offline signing with `--cert` work for you?

bzon: @mkmik that would be a good idea and it's totally fine for me. 👍 

I just wanted future GKE users of kubeseal to be aware of the issue. Because without the firewall settings, kubeseal output just displays a go runtime deadline exceeded error. What would be useful is to show a connection timeout error from `net.DialTimeout` or something.

README.md
mkmik: you mean 

```
RELEASE_BIN_URL=https://github.com/bitnami-labs/sealed-secrets/releases/download/v0.8.3/kubeseal-linux-amd64
curl -fsSL -o /usr/local/bin/kubeseal "$RELEASE_BIN_URL" && chmod +x /usr/local/bin/kubeseal
```

(probably should be `sudo curl` to work for all users)

as opposed to:

```
wget https://github.com/bitnami-labs/sealed-secrets/releases/download/v0.8.3/kubeseal-linux-amd64 -O kubeseal
sudo install -m 755 kubeseal /usr/local/bin/kubeseal
```

?

Yeah I have mixed feelings. wget vs curl? sudo curl vs sudo only the file copy? write directly vs tmp file in local dir? ...

I don't know, I'm tempted to just write a bash installer you download and run. That would also cover the arch/os detection and provide a simple

e.g.:

```
wget https://github.com/bitnami-labs/sealed-secrets/releases/download/v0.8.3/kubeseal-install.sh && sh kubeseal-install.sh
```

wdyt?

mkmik:
@bzon thanks for the commit; I included it in another PR while refactoring the docs to include cloud provider specific notes in separate documents.

PR #231, opened 2019-08-30
Readme overhaul

mkmik:


README.md
```diff
@@ -12,51 +12,135 @@ decrypted only by the controller running in the target cluster and
...
-# Note the second sealedsecret-crd.yaml file is not necessary for releases >= 0.8.0
-$ kubectl apply -f https://github.com/bitnami-labs/sealed-secrets/releases/download/$release/controller.yaml
-$ kubectl apply -f https://github.com/bitnami-labs/sealed-secrets/releases/download/$release/sealedsecret-crd.yaml
+* A cluster-side controller / operator
+* A client-side utility: `kubeseal`
+
+The `kubeseal` utility uses asymmetric crypto to encrypt secrets that only the controller can decrypt.
+
+These encrypted secrets are encoded in a `SealedSecret` resource, which you can see as a recipe for creating
+a secret. Here is how it looks like:
```
atomatt: probably s/ like//

README.md
```diff
@@ -12,51 +12,135 @@ decrypted only by the controller running in the target cluster and
...
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: mysecret
+data:
+  foo: bar  # <- base64 encoded "bar"
+```
+
+The [Usage](#usage) section explores in more detail how you craft SealedSecrets resources.
```
atomatt: s/SealedSecrets/`SealedSecret`/ I think?

README.md
```diff
@@ -12,51 +12,135 @@ decrypted only by the controller running in the target cluster and
...
-logs.
+```yaml
+apiVersion: v1
+kind: Secret
+metadata:
+  name: mysecret
+data:
+  foo: bar  # <- base64 encoded "bar"
+```
+
```
atomatt: It might be nice to mention here that, now unsealed, it can be used as usual, and maybe even link to `https://kubernetes.io/docs/concepts/configuration/secret/`. I feel that would help "round off" the overview story nicely.

README.md
```diff
@@ -12,51 +12,135 @@ decrypted only by the controller running in the target cluster and
...
+The [Usage](#usage) section explores in more detail how you craft SealedSecrets resources.
+
+### SealedSecrets as templates for secrets.
```
atomatt: s/.//, just for consistency.

README.md
```diff
@@ -12,51 +12,135 @@ decrypted only by the controller running in the target cluster and
...
+
+The previous example only focused on the encrypted secret items themselves, but the relationship between a `SealedSecret` custom resource and the `Secret` it unseals into is similar in many ways (but not in all of them) to the familiar `Deployment` vs `Pod`.
+
+In particular, the annotations and labels of a `SealedSecret` resource, are not the same as the annotations of the `Secret` that gets generated out of it.
```
atomatt: s/resource,/resource/

README.md
```diff
@@ -12,51 +12,135 @@ decrypted only by the controller running in the target cluster and
...
+
+To capture this distinction, the `SealedSecret` object has a `template` section which encodes all the fields want the controller uses when producing the unsealed `Secret`.
```
atomatt: "the fields want the controller uses" is weird, but I'm not quite sure what you meant to say.

README.md
```diff
@@ -12,51 +12,135 @@ decrypted only by the controller running in the target cluster and
...
 used. The certificate is not secret information, although you need to
-ensure you are using the correct file.
+ensure you are using the correct one.
 
 `kubeseal` will fetch the certificate from the controller at runtime
 (requires secure access to the Kubernetes API server), which is
-convenient for interactive use.  The recommended automation workflow
-is to store the certificate to local disk with
+convenient for interactive use, but it's known to be brittle when users
+have clusters with special configurations such *private GKE clusters* that have
```
atomatt: s/such/such as/

README.md
```diff
@@ -12,51 +12,135 @@ decrypted only by the controller running in the target cluster and
...
+An alternative workflow
+is to store the certificate somewhere (e.g. local disk) with
 `kubeseal --fetch-cert >mycert.pem`,
 and use it offline with `kubeseal --cert mycert.pem`.
 The certificate is also printed to the controller log on startup.
 
+
+> **NOTE**: we are working on providing key management mechanisms that offload the encryption to HSM based modules or managed cloud crypto solutions such as KMS.
+
+## Installation
```
atomatt: I can't decide if this should come just after the overview or not. The "SealedSecrets as templates for secrets" and "Public key / Certificate" sections are great info but more about how to use it, and perhaps not quite as important as how to install it in the first place.

What do you think?

README.md
```diff
@@ -12,51 +12,139 @@ decrypted only by the controller running in the target cluster and
 nobody else (not even the original author) is able to obtain the
 original Secret from the SealedSecret.
```
jjo: I'd add a wrap-up line on the lines of: 

End result is that you get a stock Secret (usable from a e.g. Pod manifest) after pushing the SealedSecret having the same namespace and name.

atomatt: I think there's some overlap between this and my comment here, https://github.com/bitnami-labs/sealed-secrets/pull/231/files#r319470607.

README.md
mkmik: I flipped those sections around a few times, unable to make up my mind about that.

I ended up with this order because of these reasons:

1. The installation from sources is long-ish and would unnecessarily deflect attention. On the other hand I didn't find a good place to bury it; it's kinda/sorta related to the Development section, but not necessarily. There are people who like to build stuff from sources even if they are not really doing that as contributors and thus it might not be fair to subject them to the full cost of a development environment (which admittedly in this case is just `make`, but it might change). Similar situation for homebrew, helm chart etc

2. It seems more logical to me to first let users know what this tool is and whether they actually want to use it, and once they decided they want to give it a try it should be easy for them to do it, but I don't feel comfortable imposing lengthly installation instructions users would need to skip over just to get a feeling of what this thing is.

3. The actual installation instructions are in the release page, so that they can be directly copy&paste a link with the full version number instead of having to do

But your feedback makes perfect sense too. What if we just link to the Install section as early as possible, and let the readers themselves choose when they want us to shut up and take the money?

Eventually I wanted to split the verbose parts in separate docs so perhaps we'd have to have a link to it anyway.

README.md
mkmik: PTAL

README.md
atomatt: An alternative view ...

Someone tells me I should use SealedSecrets because it does exactly what I want - makes my k8s secret actually secret :). I go to the github page, read the overview and am satisfied I reached the right place (the overview). All I want to do next is install it and start playing, but there's a load of other text to skip over first.

But I'm not sure either. Feel free to land as-is, it's just a suggestion.

README.md
mkmik: @jjo thanks; makes sense. Is this similar to what matt said in https://github.com/bitnami-labs/sealed-secrets/pull/231#discussion_r319470607 ?

is it ok how addresses it there?

jjo: Sure, thanks much for this update.

PR #232, opened 2019-08-30
Address linter comments

mkmik:
Some of the comments added are a bit trivial, but In order for the linter to be useful its output shouldn't be noisy.


cmd/controller/keys.go
```diff
@@ -18,10 +18,12 @@ import (
 	"k8s.io/client-go/util/keyutil"
 )
 
+// SealedSecretsKeyLabel is that label used to locate active key pairs used to decrypt sealed secrets.
 const SealedSecretsKeyLabel = "sealedsecrets.bitnami.com/sealed-secrets-key"
 
 var (
-	ErrPrivateKeyNotRSA = errors.New("Private key is not an rsa key")
+    // ErrPrivateKeyNotRSA is returned when the private key is not a valid RSA key.
```
atomatt: Bad indentation?

mkmik: good catch

PR #235, opened 2019-09-03
Put support for v1 secret format behind feature flag

mkmik:
In v0.7.0 (Mar 2018) we introduced per-item encrypted data and we deprecated the old "data" field that encrypted all the items in one blob.

Since then every use of `kubeseal` would have produced sealed secret resources in such new format.

Although it's unlikely that users have many such secrets around it would be unfriendly to just ditch this old format without giving a chance to users to notice whether they are having such secrets or not.

Thus, the deprecation of this old format is gated via a feature flag.
Since it's very unlikely people are going to be affected by this, the default value of the flag is set to break backward compatibility which means that only people who

a) have sealed secrets resources that use the old flag, and
b) cannot just reseal the secrets (with with `kubeseal --rotate`)

would have flip the flag (and thus fiddle with the sealed-secrets controller env vars etc).

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -41,6 +41,7 @@ var (
 	myCN            = flag.String("my-cn", "", "CN to use in generated certificate.")
 	printVersion    = flag.Bool("version", false, "Print version information and exit")
 	keyRotatePeriod = flag.Duration("rotate-period", 0, "New key generation period (automatic rotation disabled if 0)")
+	acceptV1Data    = flag.Bool("accept-deprecated-v1-data", false, "Accept deprecated V1 data field")
 
 	// VERSION set from Makefile
 	VERSION = "UNKNOWN"
@@ -227,6 +228,8 @@ func main2() error {
 func main() {
 	flag.Parse()
 	goflag.CommandLine.Parse([]string{})
+
+	ssv1alpha1.AcceptDeprecatedV1Data = *acceptV1Data
 
 	if *printVersion {
 		fmt.Printf("controller version: %s\n", VERSION)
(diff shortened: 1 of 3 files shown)


jjo:
Please add a note to README.md, with recommended handling for those who possible still carry v1 sealedsecrets on how to move them to v2.

mkmik:
> Please add a note to README.md, with recommended handling for those who possible still carry v1 sealedsecrets on how to move them to v2.

My intention was to add this to the release announcement.

Do you think the readme would be better? My main concern with GitHub is the way most projects are setup, people get to see the README of the unreleased next version, which is often confusing.

jjo:
> > Please add a note to README.md, with recommended handling for those who possible still carry v1 sealedsecrets on how to move them to v2.
> 
> My intention was to add this to the release announcement.
> 
> Do you think the readme would be better? My main concern with GitHub is the way most projects are setup, people get to see the README of the unreleased next version, which is often confusing.

IMO still worth, then if put in a _Backward compatibility notes for pre-v0.x.x_ it would be easily skippable if it doesn't affect the reader.

mkmik:
I don't understand. Could you make an example of how the mention would look in the README?

mkmik:
(I'd love to have a checked-in release notes document, perhaps that would be so much cleaner)

mkmik:
Added instructions for pre-v0.7.0 in the release notes document

mkmik:
since #237 the release notes are searchable with github search (which doesn't search in the github releases text body)

mkmik:
uh I _thought_ I had pushed the release notes to this PR but I forgot; I updated them independently

Issue #226, opened 2019-08-28
Per item key fingerprints for efficient key retrieval


mkmik:
Currently the controller tries to unseal each item of a secret with all the active private keys.
This becomes inefficient if the key rotation is aggressive enough and/or when enough time passes to accumulate enough active keys without obsoleting old keys.

We need a better way. I propose to prepend a key identifier of the key used to encrypt an item to the encrypted data item itself. The controller would then locate the right private key.

Implementation notes:

I suggest using the hash of the public key rather than a stateful ID such as the key pair secret name.
We already internally keep an in-memory fingerprint->key map; we could also actively fill this cache by including the hash in a label of the secret objects that store the controller key pairs.

PR #236, opened 2019-09-03
Turn on key rotation every 30 days

### LINKED ISSUE #137: Rotating master key

mkmik:
All major blocker issues have been already closed (#185, #190) and stakeholders
have been given some time to play with the feature.

Users can easily turn this off by setting an env var, see #234.

I think we're ready to turn this on for everybody.
We have time to improve the efficiency, see #226.

Closes #137

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -40,7 +40,7 @@ var (
 	validFor        = flag.Duration("key-ttl", 10*365*24*time.Hour, "Duration that certificate is valid for.")
 	myCN            = flag.String("my-cn", "", "CN to use in generated certificate.")
 	printVersion    = flag.Bool("version", false, "Print version information and exit")
-	keyRotatePeriod = flag.Duration("rotate-period", 0, "New key generation period (automatic rotation disabled if 0)")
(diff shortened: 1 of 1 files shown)


jjo:
LGTM, of course needs to be clearly stated in documentation, specially since the "pre-saved" cert workflow will be deprecated by default.

mkmik:
Good point. I already removed the "recommended" qualifier to the "offline" workflow docs.

That said, the offline workflow is not per se obsoleted by rotation. You just need to update the cert more often, which usually means you need a more distribution mechanism possibly more scalable than mail or chat. For example I know a team that uses a keybase team folder to hold the (authenticated) certificate for the production cluster they cannot otherwise access the certificate of.

jjo:
> Good point. I already removed the "recommended" qualifier to the "offline" workflow docs.
> 
> That said, the offline workflow is not per se obsoleted by rotation. You just need to update the cert more often, which usually means you need a more distribution mechanism possibly more scalable than mail or chat. For example I know a team that uses a keybase team folder to hold the (authenticated) certificate for the production cluster they cannot otherwise access the certificate of.

Sure, but people *already* using it will get it deprecated, unless of course they fix it as per your comment.

PR #238, opened 2019-09-04
Update key renewal doc and ann

mkmik:


README.md
```diff
@@ -245,12 +245,21 @@ labelled as active.
 2. Create a new key
 3. Start the key rotation cycle
 
-#### Key rotation
+#### Key renewal
 
-Keys are automatically rotated. This can be configured on controller startup with
+Keys are automatically renewed. This can be configured on controller startup with
 the `--rotate-period=<value>` flag. The `value` field can be given as golang
```
jjo: Please mention default value, in sensible units (ie don't want to run `bc -l` to convert to days/months).

README.md
```diff
@@ -245,12 +245,21 @@ labelled as active.
...
 
+The feature has been historically called "key rotation" but this term can be confusing.
+Sealed secrets are not automatically rotated and old keys are not deleted
+when new keys are generated. Old sealed secrets resources can be still decrypted.
+
+If you want to rotate the sealed secrets themselves (i.e. re-encrypt them with the latest private key),
+you need to invoke `kubeseal --rotate` which will produce a new sealed secret file freshly encrypted with
+the latest key, without making the secrets leave the cluster to the client. You can then save that file
+in your version control system (`kubeseal --rotate` doesn't update the in-cluster object).
+
```
jjo: please provide a cli cmds example, as it's a bit confusing to read `kubeseal --rotate` verbatim (without `-n NAMESPACE NAME` or alike) that IMO reads as general/cluster-wide operation to be performed, rather than an outputted sealedsecret content (as it becomes clearer near the end of the paragraph).

RELEASE-NOTES.md
```diff
@@ -8,6 +8,16 @@ Latest release:
...
 
+### Private key renewal
+
+This realease turns on an important security feature: a new private key will be now created every 30 days by default.
+Existing sealed-secrets resources will still be decrypted until the keys are manually phased out.
+
+You can read more about this feature in this [README section](https://github.com/bitnami-labs/sealed-secrets#key-renewal) or in the original GH issue #137.
+
+This feature impacts people who use the offline workflow with `kubeseal --cert`, who will have to update
+the certificates more often.
```
jjo: I think this deserves more clarification than «more often», as technically the validity of the saved cert would depend on the previous/old privkeys garbage collection that people would implement by themselves (something that would also need to be documented/hinted about IMO).

mkmik: actually I had a brain fart, good catch

README.md
mkmik: good catch; done

README.md
mkmik: PTAL

mkmik:
@jjo, @atomatt: I just realized the explaination about what key rotation actually tries to accomplish was not clear enough and I gave that section of the README an overhaul. PTAL

RELEASE-NOTES.md
```diff
@@ -8,6 +8,17 @@ Latest release:
 
 ## Announcement
 
+### Private key renewal
+
+This realease turns on an important security feature: a new private key will be now created every 30 days by default.
```
dbyron0:
```suggestion
This release turns on an important security feature: a new private key will be now created every 30 days by default.
```

README.md
```diff
@@ -230,32 +230,60 @@ that only the intended `SealedSecret` is uploaded to the cluster.  The
...
 `Secret` are now hidden while outside the cluster.
 
-## Details
+## Secret Rotation
 
-This controller adds a new `SealedSecret` custom resource. The
-interesting part of a `SealedSecret` is a base64-encoded
-asymmetrically encrypted `Secret`.
+You should always rotate your secrets. But since your secrets are encrypted with another secret,
+you need to understand how these two layers relate in order to take the right.
```
atomatt: take the right ???

README.md
```diff
@@ -264,6 +292,38 @@ Labelling a secret with anything other than `active` effectively deletes
 the key from the sealed secrets controller, but it is still available in k8s for
 manual encryption/decryption if need be.
 
+### Re-encryption
+
+Before you can get rid of some old sealing keys you need to re-encrypt your SealedSecrets with with the latest private key).
```
atomatt: s/with with/with/

PR #245, opened 2019-09-10
Don't panic on error

mkmik:
Currently when `kubeseal` wants to report an error, we just `panic` from main, which is ugly.
This change also weaves CLI flags into a `run` function so that we can test that in isolation.


mkmik:
Done. Added actual integration test and also a some unit tests that directly `run`.

PR #252, opened 2019-09-11
Rename --rotate to --re-encrypt

### LINKED ISSUE #250: Rename --rotate to --re-encrypt
`kubeseal --rotate` has a confusing name and several users have expressed the concern that invoking it will have some side effects on the cluster state (which is not what that command does).

(Keeping backward compat for a bit is not a bad idea)

mkmik:
Closes #250

I tried also the name "--refresh" which appears to be less ugly but ultimately I think `--re-encrypt` really spells it out and thus offers less changes for confusion.

The README is intentionally still referring to `--rotate` until we cut a new release)


Crevil:
I like that name better as well. Thanks for keeping developers in mind on this.

jjo:
> The README is intentionally still referring to --rotate until we cut a new release)
and RELEASE-NOTES.md (which I guess you'd fix alongside README).

PR #259, opened 2019-09-13
Make unsealer ClusterRole naming consistent with other resources

ajcann:
All sealed-secret kubernetes resources follow the naming scheme sealed-secret-<purpose> except for the secrets-unsealer ClusterRole. Changing to sealed-secrets-unsealer will make all resource naming consistent.

### DIFF
--- a/controller.jsonnet
+++ b/controller.jsonnet
@@ -8,7 +8,7 @@ controller {
 
   account: kube.ServiceAccount('sealed-secrets-controller') + $.namespace,
 
-  unsealerRole: kube.ClusterRole('secrets-unsealer') {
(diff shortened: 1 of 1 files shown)


mkmik:
I have mixed feelings about this. On one hand inconsistency sucks, but changing this now could be a pain for upgrades. Is it worth it?

ajcann:
Fair point. Helm releases should take care of this scenario without too much trouble, but for those installing via kustomize/kubectl it would leave a stale ClusterRole resource behind. Not a lot of gain for what would necessitate an upgrade path deviation. I'll go ahead and close.

PR #260, opened 2019-09-13
Add note about disabling key renewal/rotation

ajcann:
I couldn't find any documentation about disabling automatic key renewal other than in the help line of the controller source ( https://github.com/bitnami-labs/sealed-secrets/blob/e559e1006a6de8f1b93ee57aeb47e750520c9e80/cmd/controller/main.go#L43 )


mkmik:
oops; thanks!

PR #274, opened 2019-09-20
List items that failed to decrypt

mkmik:
Sometimes people update existing secrets by merging the output of another kubeseal execution
(e.g. plain kubeseal or kubeseal --raw) bit by mistake they don't use the same scope, or the same name
or namespace.

Currently it's hard to figure out what's wrong, partially because all the controller says is that
a given secret failed to unseal.

This change allows the user to know which item is the culprit:

```
0s          Warning   ErrUnsealFailed   sealedsecret/mysecret   Failed to unseal: no key could decrypt secret (["bar"])
```


pkg/apis/sealed-secrets/v1alpha1/sealedsecret_expansion.go
```diff
@@ -254,18 +254,27 @@ func (s *SealedSecret) Unseal(codecs runtimeserializer.CodecFactory, privKeys ma
...
 			if err != nil {
-				return nil, err
+				badKeys = append(badKeys, key)
+				unsealingError = err
 			}
 			secret.Data[key] = plaintext
 		}
 
+		if unsealingError != nil {
+			return nil, fmt.Errorf("%v (affected encrypted data items: %q)", unsealingError, badKeys)
```
jjo: We should clarify that `%v` was the last observed unsealing error (rather than the "common/same" error on all `badKeys`)

mkmik:
PTAL; uses multierror package to correclty deal with the corner-case where some of the keys raise different errors but at the same time deduplicates the common case all the errors are the same (and yet we still want to see which keys are affected). Link to docs, to save you a search: https://godoc.org/github.com/mkmik/multierror

PR #275, opened 2019-09-25
--raw requires --name in all cases

ajcann:
Though I'm wondering if this is actually a bug given --scope [namespace-wide, cluster-wide] should not require a secret name as an encryption label.


mkmik:
Yeah. A better fix would be to not require --name if scope is greater than strict. Would you like to give a stab at it?

PR #276, opened 2019-09-26
Fix support for non-strict scopes when using --raw

ajcann:
Allows the use of --raw without specifying secret name and namespace for cluster-wide scope and without specifying secret name for namespace-wide scope. Warns if secret name is provided when using non-strict scope.

(Note: I've never written a lick of Golang before so please feel free to discard with impunity :) )

### DIFF
--- a/cmd/kubeseal/main.go
+++ b/cmd/kubeseal/main.go
@@ -442,26 +442,13 @@ func run(w io.Writer, secretName, controllerNs, controllerName, certFile string,
 		if err != nil {
 			return err
 		}
-		
-		switch sealingScope {
-		case ssv1alpha1.ClusterWideScope:
-			if len(secretName) != 0 {
-				fmt.Fprintf(w, "Warning: cluster-wide scope specified, ignoring provided secret name: %s\n", secretName)
-			}
-		case ssv1alpha1.NamespaceWideScope:
-			if ns == "" {
-				return fmt.Errorf("must provide the --namespace flag with --raw and namespace-wide scope")
-			}
-			if len(secretName) != 0 {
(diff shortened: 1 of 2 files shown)


cmd/kubeseal/main.go
```diff
@@ -442,11 +442,26 @@ func run(w io.Writer, secretName, controllerNs, controllerName, certFile string,
 		if err != nil {
 			return err
 		}
-		if ns == "" {
-			return fmt.Errorf("must provide the --namespace flag with --raw")
-		}
-		if secretName == "" {
-			return fmt.Errorf("must provide the --name flag with --raw")
+		
+		switch sealingScope {
```
mkmik: I have mixed feelings about the warning.

The --raw command is also intended to be a hook for automation (custom shell scripts or IDE integration) so it's possible that a --name flag will be passed regardless of the scope.

I don't think it's worth complicating such scripts so they "know" that they have to remove the --name flag depending on the scope

----

The scopes strictly ordered so this could be written also as:

```
if ns == "" && sealingScope < ssv1alpha1.ClusterWideScope {
  return fmt.Errorf("must provide the --namespace flag with --raw and --scope %s", sealingScope)
}

if name == "" && sealingScope < ssv1alpha1.NamespaceWideScope {
  return fmt.Errorf("must provide the --name flag with --raw and --scope %s", sealingScope)
}
```

The sealingScope variable already defaults to strict. The invariant is implemented in the ssv1alpha1 package.

(Please double-check; I might have fat-fingered something, typing this on my phone from a hospital bed)

ajcann: Thanks - much cleaner and agreed re: your thinking on the warnings.

mkmik:
Thanks!

PR #287, opened 2019-10-02
Teach --cert to fetch from URLs

### LINKED ISSUE #281: Accept URLs in --cert
Certificates can be renewed periodically, and it's very important from a security POV that new sealed secrets are sealed with the latest certificate.

Forcing users to manually copy certificates locally is error-prone. Ideally the certificates are served in some trusted location, accessed via https (possibly live-served via an ingress from the controller itself).

Unfortunately `--cert` only accepts local files.

CC: @jjo

mkmik:
Closes #281

### DIFF
--- a/cmd/kubeseal/main.go
+++ b/cmd/kubeseal/main.go
@@ -12,6 +12,7 @@ import (
 	"io"
 	"io/ioutil"
 	"net/http"
+	"net/url"
 	"os"
 	"strings"
 
@@ -44,7 +45,7 @@ const (
 
 var (
 	// TODO: Verify k8s server signature against cert in kube client config.
-	certFile       = flag.String("cert", "", "Certificate / public key to use for encryption. Overrides --controller-*")
+	certURL        = flag.String("cert", "", "Certificate / public key file/URL to use for encryption. Overrides --controller-*")
 	controllerNs   = flag.String("controller-namespace", metav1.NamespaceSystem, "Namespace of sealed-secrets controller.")
 	controllerName = flag.String("controller-name", "sealed-secrets-controller", "Name of sealed-secrets controller.")
 	outputFormat   = flag.StringP("format", "o", "json", "Output format for sealed secret. Either json or yaml")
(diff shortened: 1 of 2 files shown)


cmd/kubeseal/main.go
```diff
@@ -142,15 +143,41 @@ func prettyEncoder(codecs runtimeserializer.CodecFactory, mediaType string, gv r
...
-	f, err := os.Open(certFile)
+func isFilename(name string) (bool, error) {
+	u, err := url.Parse(name)
 	if err != nil {
+		return false, err
+	}
+	return u.Scheme == "", nil
+}
+
+func openCertLocal(certURL string) (io.ReadCloser, error) {
```
jjo: IMO we should rename it to `openCertURI()`

cmd/kubeseal/main_test.go
```diff
@@ -109,18 +109,23 @@ func TestParseKey(t *testing.T) {
...
 
-	f, err := openCert(certFile)
-	if err != nil {
-		t.Fatalf("Error reading test cert file: %v", err)
-	}
+	testCases := []string{"", "file://"}
+	for _, prefix := range testCases {
+		certURL := fmt.Sprintf("%s%s", prefix, certFile)
 
-	data, err := ioutil.ReadAll(f)
```
jjo: Would it make sense to also test `http` URI? (dunno how much much code-vs-lib test coverage we'd be increasing)

cmd/kubeseal/main.go
mkmik: except a local filename is not an URI.

We even have a test that highlights that: we check for a file called `"/?this/file/cannot/possibly/exist/right?"`; if we treat arguments to `--cert` as URIs, possibly implicitly in the `file://` scheme, we'd require users to URI-encode the paths.

I thus opted for "Local" meaning "fetch the file from the POV of the client", as opposed from "use the k8s api server proxy to fetch things from within the cluster"

cmd/kubeseal/main_test.go
mkmik: not sure; the only benefit I see would be to guard against us replacing the straight-forward `client.Get` call with something that treats `file://` specially and skips the `net/http` package. I don't find that possibility particularly likely, but you never know ... :-)

mkmik: anyway it turned out to be straightforward to test, so done

cmd/kubeseal/main.go
mkmik: naming is hard; added comments

mkmik:
PTAL

cmd/kubeseal/main.go
jjo: > except a local filename is not an URI.

Looks like it is: https://tools.ietf.org/html/rfc8089

jjo: > naming is hard; added comments

Fair, comments "fulfill" it, but I'd still raise rfc8089 to claim the openCertURI() rename :)

mkmik: sure, rfc8089 is invoked when the user literally provides a parameter that starts with `file:...`

It's incorrect to assume that any host pathname that is valid to open with `os.Open` can be prefixed with `file:` and transformed to a rf8090 uri though.

I'll just make a couple of examples, using `curl` but the Go's FileTransport behaves in the same way.

1. 

```bash
$ echo foo >'/tmp/cert%E3%84%93.pem'
$ echo bar >'/tmp/certㄓ.pem'
$ cat '/tmp/cert%E3%84%93.pem'
foo
$ curl 'file:///tmp/cert%E3%84%93.pem'
**bar**
```

2.

```bash
$ echo foo >'/tmp/cert.pem'
$ echo bar >'/tmp/cert.pem?'
$ cat  '/tmp/cert.pem'
foo
$ curl 'file:///tmp/cert.pem'
foo
$ cat '/tmp/cert.pem?'
bar
$ curl 'file:///tmp/cert.pem?'
**foo**
```

I'll spare you the examples with windows backslashes.

mkmik: that said, I hear you; the naming is confusing. So I split it further and made it explicit what is a filename what is an URI

Crevil:
This is super useful! Thanks for doing it. 🎉

PR #293, opened 2019-10-14
New Feature: Add ability to encrypt/decrypt with vault transit engine

arturo-c:
Problem: Want a way to use vault which is already in use, to encrypt/decrypt secrets, such that no additional pub key/cert is needed.  

Solution: Make use of Vault Kubernetes authentication to map service accounts with the appropriate role to be able to use encryption within a k8s cluster

Considerations: A vault cluster must be addressable with corresponding vault token or kubernetes authentication enabled to be able to use vault for secret encryption

This PR is working progress and at this point would like some feedback from maintainers as to the appropriateness and alignment with the sealed-secrets project.

We are actively using this forked branch in our dev clusters and is working as intended, if this is desired functionality I can provide additional details and README.md additions on how to use as well as cleaning up code and writing some more tests.


mkmik:
thanks. I was thinking that given the sheer amount of possible encryption backends, we should have some sort of plugin architecture to decouple the encryption part from the sealed secrets controller. I was thinking about a binary receiving RPC requests from the sealed-secrets controller, (possibly sitting in a sidecar and talking via localhost or unix domain socket). Would you be interested in porting your vault encryption logic into that form once we sketch together the API?

arturo-c:
@mkmik yup, definitely would be interested, we are actively using this branch but would like to contribute the feature to sealed-secrets.  Let me know how/when I can help

PR #294, opened 2019-10-16
Document restrictions of `strict` scope

AndiDog:
A quite typical workflow is that _developers_ engineer a Kubernetes-based application which uses certain secrets, then ask _devops_ to generate such secrets. In case the secret name mismatches – by mistake, or because the names weren't fully decided at the time – the controller would fail with "decryption error" since changing namespace/name isn't possible. Document that for other users who ran into this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -174,11 +174,13 @@ Furthermore, namespaces are not the only level at which RBAC configurations can 
 
 That said, there are many scenarios where you might not care about this level of protection. For example, the only people who have access to your clusters are either admins or they cannot read any secret resource at all. You might have a use case for moving a sealed secret to other namespaces (e.g. you might not know the namespace name upfront), or you might not know the name of the secret (e.g. it could contain a unique suffix based on the hash of the contents etc).
 
-You can select the "scope":
+These are the possible scopes:
 
-* strict (default)
-* namespace-wide: you can freely rename the sealed secret within a given namespace
-* cluster-wide: you
+* `strict` (default): the secret must be sealed with exactly the same *name* and *namespace*. These attributes become *part of the encrypted data* and thus changing name and/or namespace would lead to "decryption error".
+* `namespace-wide`: you can freely *rename* the sealed secret within a given namespace
+* `cluster-wide`: the secret can be unsealed in *any* namespace and can be given *any* name.
+
+In contrast to the restrictions of *name* and *namespace*, secret *data keys* (e.g. `spec.encryptedData.my-key`) can be renamed at will without losing the ability to decrypt the sealed secret.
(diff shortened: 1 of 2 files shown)


README.md
```diff
@@ -190,6 +190,9 @@ The lack of any of such annotations means `strict` mode. If both are set, `clust
 > NOTE: next release will consolidate this into a single `sealedsecrets.bitnami.com/scope` annotation.
 > NOTE: next release will enhance the kubeseal tooling so you don't have to fiddle with these annotations.
 
+By default, the `strict` scope is used, which means *both the namespace and name are part of the
```
mkmik: This is described in the paragraph at line 173.
But I know that people faced with a wall of text tend to not read it, so it's ok if we repeat, summarize and highlight common pitfalls.

That said, shouldn't we add this text above at line 179? I just noticed that I butchered that list (see line 181 that only says "you"):

What about:

```diff
- You can select the "scope":
+ Depending on the scope, the name and/or namespace become *part of the encrypted data* and thus changing name and/or namespace would lead to "decryption error". These are the possible scopes:

- * strict (default)
+ * strict (default): the secret must be sealed with exactly the same *name* and *namespace*
  * namespace-wide: you can freely *rename* the sealed secret within a given namespace
- * cluster-wide: you
+ * cluster-wide: the secret can be unsealed in *any* namespace and can be given *any* name.
```

AndiDog: I've put the scope explanation into (only) this bullet point list, and added a distinction to secret _data keys_ which can be freely renamed.

README.md
```diff
@@ -174,11 +174,13 @@ Furthermore, namespaces are not the only level at which RBAC configurations can
...
+These are the possible scopes:
 
-* strict (default)
-* namespace-wide: you can freely rename the sealed secret within a given namespace
-* cluster-wide: you
+* `strict` (default): the secret must be sealed with exactly the same *name* and *namespace*. These attributes become *part of the encrypted data* and thus changing name and/or namespace would lead to "decryption error".
+* `namespace-wide`: you can freely *rename* the sealed secret within a given namespace
+* `cluster-wide`: the secret can be unsealed in *any* namespace and can be given *any* name.
+
+In contrast to the restrictions of *name* and *namespace*, secret *data keys* (e.g. `spec.encryptedData.my-key`) can be renamed at will without losing the ability to decrypt the sealed secret.
```
mkmik: Minor (will fix in a separate PR): to avoid confusion with encryption keys, we use the word _items_ throughout the documentation to refer to entries in the encryptedData map

mkmik:
thanks!

PR #301, opened 2019-10-29
Fix GenerateName in test

mkmik:
Currently the controller tests are plagued by:
```
Failed to generate new key : secrets "" already exists
```

which is caused by the tests creating secrets with empty Name assuming that GenerateName will issue a new name. Unfortunately the fake client doesn't honour GenerateName. According to https://github.com/kubernetes/client-go/issues/439 users are supposed to implement this in their own testing code, so here we are.

I need this in order to implement the test for #299

### DIFF
--- a/cmd/controller/main_test.go
+++ b/cmd/controller/main_test.go
@@ -4,11 +4,14 @@ import (
 	"crypto/rsa"
 	"crypto/x509"
 	"encoding/pem"
+	"fmt"
 	"testing"
 	"time"
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/runtime"
+	krand "k8s.io/apimachinery/pkg/util/rand"
 	"k8s.io/client-go/kubernetes"
 	"k8s.io/client-go/kubernetes/fake"
 	ktesting "k8s.io/client-go/testing"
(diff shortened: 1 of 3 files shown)


cmd/controller/main_test.go
```diff
@@ -29,9 +32,20 @@ func hasAction(fake *fake.Clientset, verb, resource string) bool {
 	return findAction(fake, verb, resource) != nil
 }
 
+// generateNameReactor implements the logic required for the GenerateName field to work when using
+// the fake client. Add it with client.PrependReactor to your fake client.
+func generateNameReactor(action ktesting.Action) (handled bool, ret runtime.Object, err error) {
+	s := action.(ktesting.CreateAction).GetObject().(*v1.Secret)
+	if s.Name == "" && s.GenerateName != "" {
+		s.Name = fmt.Sprintf("%s-%d", s.GenerateName, mathrand.Intn(10000))
```
atomatt: This is probably good enough but may cause super weird failures if the same int is generated 2+ times.

Maybe it would be better to use a per-test generator? Would a simple counter then be sufficient to provide uniqueness?

But if you stick with unique, does it need to be `Seed`ed?

mkmik: what if we just use the same random string function that k8s uses to generate ranm pod names? (yes there is still the theoretical chance of conflict, but if that was the only source of flakiness I think I can live with that)

atomatt: Works for me. I really just wanted to point out the potential issue in case it was likely to cause issues.

mkmik: yeah indeed 1000 was too small a number

PR #307, opened 2019-11-04
Implement decrypt functionality for disaster recovery

mkmik:


cmd/kubeseal/main_test.go
```diff
@@ -377,7 +417,7 @@ func TestMergeInto(t *testing.T) {
 
 func TestVersion(t *testing.T) {
 	var buf strings.Builder
-	err := run(&buf, "", "", "", "", true, false, false, false, false, nil, "")
+	err := run(&buf, "", "", "", "", true, false, false, false, false, nil, "", false, nil)
```
anguslees: ouch :/

cmd/kubeseal/main.go
```diff
@@ -59,6 +61,8 @@ var (
 	fromFile       = flag.StringSlice("from-file", nil, "(only with --raw) Secret items can be sourced from files. Pro-tip: you can use /dev/stdin to read pipe input. This flag tries to follow the same syntax as in kubectl")
 	sealingScope   ssv1alpha1.SealingScope
 	reEncrypt      bool // re-encrypt command
+	unseal         = flag.Bool("recovery-unseal", false, "Decrypt a sealed secrets file obtained from stdin, using the private key passed with --recovery-private-key. Intended to be used in disaster recovery mode.")
+	privKeys       = stringlist.Flag("recovery-private-key", "Private key filename used by the --recovery-unseal command. Multiple files accepted either via comma separated list of by repetition of the flag.")
```
anguslees: Is this the same as `flag.StringSlice`?  https://godoc.org/github.com/spf13/pflag#StringSlice

cmd/kubeseal/main_test.go
mkmik: Yeah, I'd really like to move away from this pattern...

cmd/kubeseal/main.go
mkmik: Uh, I forgot we were using pflag here...

PR #316, opened 2019-11-14
Give an option to search only the current namespace

alexciouca:
Giving an option to search only the current namespace is very helpful if somebody wants to deploy the controller in multiple namespaces in the same cluster and use the controller only per its own namespace.

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -8,6 +8,7 @@ import (
 	"fmt"
 	"io"
 	"io/ioutil"
+	v1 "k8s.io/api/core/v1"
 	"log"
 	"os"
 	"os/signal"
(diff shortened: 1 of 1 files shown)


cmd/controller/main.go
```diff
@@ -43,8 +44,9 @@ var (
 	myCN           = flag.String("my-cn", "", "CN to use in generated certificate.")
 	printVersion   = flag.Bool("version", false, "Print version information and exit")
 	keyRenewPeriod = flag.Duration("key-renew-period", defaultKeyRenewPeriod, "New key generation period (automatic rotation disabled if 0)")
-	acceptV1Data   = flag.Bool("accept-deprecated-v1-data", false, "Accept deprecated V1 data field")
+	acceptV1Data   = flag.Bool("accept-deprecated-v1-data", false, "Accept deprecated V1 data field.")
 	keyCutoffTime  = flag.String("key-cutoff-time", "", "Create a new key if latest one is older than this cutoff time. RFC1123 format with numeric timezone expected.")
+	namespaceAll   = flag.Bool("namespaceAll", true, "Scan all namespaces or only the current namespace (default=true).")
```
mkmik: The other flags use snake-case.

Perhaps reordering the words to be `--all-namespaces` (the same name used by kubectl)?
```suggestion
	namespaceAll   = flag.Bool("all-namespaces", true, "Scan all namespaces or only the current namespace (default=true).")
```

alexciouca: Agree, makes more sense like this. Thanks!

PR #334, opened 2019-12-12
Add flag to allow sealing of empty secret objects

alexciouca:


cmd/kubeseal/main.go
```diff
@@ -53,6 +53,7 @@ var (
 	controllerName = flag.String("controller-name", "sealed-secrets-controller", "Name of sealed-secrets controller.")
 	outputFormat   = flag.StringP("format", "o", "json", "Output format for sealed secret. Either json or yaml")
 	dumpCert       = flag.Bool("fetch-cert", false, "Write certificate to stdout. Useful for later use with --cert")
+	forceEmpty     = flag.Bool("force-empty-data", false, "Allow empty data in the secret object")
```
mkmik: I see that "force" here is used in the common unix pattern of "-f" == "do what I told you", but in this case wouldn't it be clearer to just say `"allow-empty-data"`?

cmd/kubeseal/main.go
```diff
@@ -225,18 +226,13 @@ func openCert(certURL string) (io.ReadCloser, error) {
...
 	}
 
-	if len(secret.Data) == 0 && len(secret.StringData) == 0 {
-		// No data. This is _theoretically_ just fine, but
-		// almost certainly indicates a misuse of the tools.
-		// If you _really_ want to encrypt an empty secret,
-		// then a PR to skip this check with some sort of
-		// --force flag would be welcomed.
+	if len(secret.Data) == 0 && len(secret.StringData) == 0 && !forceEmpty {
 		return fmt.Errorf("Secret.data is empty in input Secret, assuming this is an error and aborting")
```
mkmik: perhaps we could mention the `--allow-empty-data` flag in the error message?

mkmik:
you need to fix tests. (note to self: I really have to change this long function parameter lists into objects..., in another PR)

cmd/kubeseal/main.go
alexciouca: Agree. Thank you for the review.

PR #339, opened 2019-12-31
Set fsGroup to nobody

### LINKED ISSUE #338: Crash when running on cluster with BoundServiceAccountTokenVolume feature gate enabled
The BoundServiceAccountTokenVolume feature gate modifies service account tokens to be provided via a projected volume.  Permissions are slightly different for the token file itself:

```
-rw-------    1 root     root          1058 Dec 31 14:31 /var/run/secrets/kubernetes.io/serviceaccount/..data/token
```

Normally, the token file is mode 644.  The difference results in a crash:

```
2019/12/31 14:37:26 Starting sealed-secrets controller version: v0.9.6+dirty
panic: open /var/run/secrets/kubernetes.io/serviceaccount/token: permission denied

goroutine 1 [running]:
main.main()
	/home/travis/gopath/src/github.com/bitnami-labs/sealed-secrets/cmd/controller/main.go:271 +0x215
```

The fix is to set `securityContext.fsGroup` for the podspec to something like the following:

```yaml
spec:
  securityContext:
    fsGroup: 65534
```

This results in the token file having mode 640 with a group owner of 65534.  Any `fsGroup` value is acceptable, but this one is typically the `nobody` group.

mkmik:
Closes #338

### DIFF
--- a/controller-norbac.jsonnet
+++ b/controller-norbac.jsonnet
@@ -21,6 +21,9 @@ local namespace = 'kube-system';
     spec+: {
       template+: {
         spec+: {
+          securityContext+: {
+            fsGroup: 65534,
+          },
           containers_+: {
             controller: kube.Container('sealed-secrets-controller') {
               image: $.controllerImage,


controller-norbac.jsonnet
```diff
@@ -37,6 +37,7 @@ local namespace = 'kube-system';
                 readOnlyRootFilesystem: true,
                 runAsNonRoot: true,
                 runAsUser: 1001,
+                fsGroup: 65534,
```
javanthropus: From my testing, this field is not allowed in the securityContext of a containerSpec.  It needs to be in a new securityContext section of the podSpec template.

mkmik: indeed

```
To further debug and diagnose cluster problems, use 'kubectl cluster-info dump'.
error: error validating "controller.yaml": error validating data: ValidationError(Deployment.spec.template.spec.containers[0].securityContext): unknown field "fsGroup" in io.k8s.api.core.v1.SecurityContext; if you choose to ignore these errors, turn validation off with --validate=false
```

sorry I was lazy and wanted the testing system to guide me.
Fixed and run through `kubecfg validate` to pre-check.

PTAL

controller-norbac.jsonnet
```diff
@@ -21,6 +21,9 @@ local namespace = 'kube-system';
     spec+: {
       template+: {
         spec+: {
+          securityContext+: {
+            fsGroup: 65534,
+          },
```
javanthropus: I'm not familiar with jsonnet, but this seems to be right.

PR #352, opened 2020-02-10
Warn if using --recovery-private-key without --recovery-unseal

### LINKED ISSUE #351: Offline Decrypting a Sealed Secret with kubeseal throws an error
Trying to decrypt a sealed secret with the privatekey file using `kubeseal <mysecret.yaml --recovery-private-key master.key`outputs an error
```
error: converting (v1alpha1.SealedSecret) to (v1.Secret): Data not present in src
```
Steps to reproduce:

1 - Fetch your sealed secret in yaml format: `kubectl get sealedsecrets mysecret -o yaml > mysecret.yaml`
2 -  Extract your privatekey list into a single file: ` kubectl get secret -n kube-system -l sealedsecrets.bitnami.com/sealed-secrets-key -o yaml >master.key`
3 - Try to unseal the sealedsecret with kubeseal: `kubeseal <mysecret.yaml --recovery-private-key master.key`
4 - Get error: `error: converting (v1alpha1.SealedSecret) to (v1.Secret): Data not present in src`

mkmik:
Why are we not just raising an error if --recovery-private-key is set and --recovery-unseal is not,
or just implying the "unseal" mode if "--recovery-private-key" is set?

1. because doing so would break backwards compat and backwards compat is an important rule to not break lightly
2. because we use `flagenv` to default some flag values, so it's entirely possible that some users pass the `--recovery-private-key` implicitly while using kubeseal's other commands.

The current "flag based" UX sucks a bit, and I do want to address that, but not know as there are bigger fishes to fry and changing the UX would create a lot of noise.

Closes #351

### DIFF
--- a/cmd/kubeseal/main.go
+++ b/cmd/kubeseal/main.go
@@ -618,6 +618,9 @@ func run(w io.Writer, secretName, controllerNs, controllerName, certURL string, 
 	if unseal {
 		return unsealSealedSecret(w, os.Stdin, scheme.Codecs, privKeys)
 	}
+	if len(privKeys) != 0 && isatty.IsTerminal(os.Stderr.Fd()) {
+		fmt.Fprintf(os.Stderr, "warning: ignoring --recovery-private-key because unseal command not chosen with --recovery-unseal\n")
+	}
 
 	if raw {
 		ns, _, err := namespaceFromClientConfig()


cmd/kubeseal/main.go
```diff
@@ -618,6 +618,9 @@ func run(w io.Writer, secretName, controllerNs, controllerName, certURL string,
 	if unseal {
 		return unsealSealedSecret(w, os.Stdin, scheme.Codecs, privKeys)
 	}
+	if len(privKeys) != 0 && isatty.IsTerminal(os.Stderr.Fd()) {
+		fmt.Fprintf(os.Stderr, "warning: you passed --recovery-private-key without also choosing the unseal command with --recovery-unseal, are you sure?\n")
```
atomatt: the "are you sure?" bit is a bit weird.

mkmik: indeed; I hoped you could help me come up with a better phrasing :-)

mkmik:
PTAL

mkmik:
that's even better, thanks!

PR #356, opened 2020-02-18
More clarifications around sealing key renewal

mkmik:


README.md
```diff
@@ -345,17 +381,25 @@ TL;DR:
...
+### Sealing key renewal
 
-Keys are automatically renewed every 30 days. This can be configured on controller startup with
-the `--key-renew-period=<value>` flag. The `value` field can be given as golang
-duration flag (eg: `720h30m`). A value of `0` will disable automatic key renewal.
+Sealing keys are automatically renewed every 30 days. Which means a new sealing key is created and appended to the set of active sealing keys the controller can use to unseal Sealed Secret resources.
+
+The most recently created sealing key is the one used to seal new secrets when you use `kubeseal` and it's the one whose certificate is downloaded when you use `kubeseal --fetch-cert`.
+
+The renewal time of 30d is reasonable default, but it can be tweaked as needed
```
mkmik:
```suggestion
The renewal time of 30d is just a reasonable default, but it can be tweaked as needed
```

README.md
```diff
@@ -345,17 +381,25 @@ TL;DR:
...
+with the `--key-renew-period=<value>` flag. The `value` field can be given as golang
+duration flag (eg: `720h30m`).
+
+A value of `0` will disable automatic key renewal. Of course, it's possible you have a valid use for disabling automatic sealing key renewal; but experience has shown that new users often tend to jump to conclusions that they want control over key renewal before fully understanding how sealed secrets work. Read more about this in the [common misconceptions](#common-misconceptions-about-key-renewal) section below.
```
mkmik:
```suggestion
A value of `0` will disable automatic key renewal. Of course, it's possible you have a valid use case for disabling automatic sealing key renewal; but experience has shown that new users often tend to jump to conclusions that they want control over key renewal, before fully understanding how sealed secrets work. Read more about this in the [common misconceptions](#common-misconceptions-about-key-renewal) section below.
```

README.md
```diff
@@ -384,6 +428,22 @@ start sealing your new rotated secrets, otherwise you'll be giving attackers acc
 
 A key can be generated early by passing the current timestamp to the controller into a flag called `--key-cutoff-time` or an env var called `SEALED_SECRETS_KEY_CUTOFF_TIME`. Expected format is RFC1123, you can generate it with the `date -R` unix command.
 
+### Common misconceptions about key renewal
+
+Sealed secrets sealing keys are not access control keys (e.g. like a password); they are more like the GPG key you might use to read encrypted mail sent to you. Let's conitinue with the email analogy for a bit:
```
mkmik:
```suggestion
Sealed secrets sealing keys are not access control keys (e.g. like a password); they are more like the GPG key you might use to read encrypted mail sent to you. Let's continue with the email analogy for a bit:
```

README.md
```diff
@@ -384,6 +428,22 @@ start sealing your new rotated secrets, otherwise you'll be giving attackers acc
...
+
+Imagine you have reasons to believe your private GPG key might have been compromised, you have more to lose than to gain if the first thing you do is to just delete your private key. All the previous emails sent with that key are no longer accessible to you (unless you have a decrypted copy of those emails), nor are new emails sent by your friends whom you have not yet managed to tell to use the new key.
```
mkmik:
```suggestion
Imagine you have reasons to believe your private GPG key might have been compromised. You'd have more to lose than to gain if the first thing you do is to just delete your private key. All the previous emails sent with that key are no longer accessible to you (unless you have a decrypted copy of those emails), nor are new emails sent by your friends whom you have not yet managed to tell to use the new key.
```

weslleycamilo:
Hi guys, 

I've been reading the docs I found and got to this PR which is great to improve the documentation.

What about **to include more details about a case of disaster of a K8s** cluster :thinking: 

I've been testing the sealed secrets and I've considering use its key  renewal but faced to some steps which I've doubts, some of them:
 

-  need to re-encrypt everything and I don't know a good way to do it to many secrets once it looks like I'll have to automate it to find all the secrets on kubernetes and re-encrypt it right ? or there is a flag ?

- If I would like to create my own key and not renew it can I do that ? Would I have this flexibility ? I considered it after not know what to do in case I lose my whole cluster and don't have all the keys which was lost in the broken cluster. So if I have my own key it won't be a problem to encrypt, decrypt and launch a new cluster which can decrypt my sealedsecrets to the cluster from github whith Fluxcd.

mkmik:
@weslleycamilo sure, disaster recovery is important and I'd surely like to improve the docs and/or the feature set to make disaster recovery more approachable.

Would you mind creating an issue for that; it's hard to have a good discussion in a non directly related PR that is about to be merged and thus closed.

weslleycamilo:
sure  @mkmik. Thank you for your feedback.

Issue #357, opened 2020-02-18
Improve documentation to provide more details to work in a case of disaster recovery


weslleycamilo:
Hi guys,

I've been reading the docs and I found some good details which helped me to test it.

What about to **include more details about a case of disaster of a K8s** cluster :thinking: 

I've been testing the sealed secrets and I've considering use its key renewal but faced some doubts, some of them are:

- need to re-encrypt everything and I don't know a good way to do it to many secrets once it looks like I'll have to automate it to find all the secrets on kubernetes and re-encrypt it right ? or there is a flag ?

- If I would like to create my own key and not renew it can I do that ? Would I have this flexibility ? I considered it after not know what to do in case I lose my whole cluster and don't have all the keys which was lost in the broken cluster. So if I have my own key it won't be a problem to encrypt, decrypt and launch a new cluster which can decrypt my sealedsecrets to the cluster from github whith Fluxcd.

So my building/testing scenario is I've all the sealed secrets in the Github and in case of launch/migration to new cluster I would like to just deploy FluxCD and Sealed secrets and get everything on my k8s manifests deployed ASAP and I'm not sure how to do it which sealed renew once I might lose it in the broken cluster, or have to manually re-encrypt everything :(  It looks easier to work on if I can create and use my own keys and use in all my cluster and when I would like to re-encrypt it I can do but in a planned way and not get surprised by  a disaster recovery.

mkmik:
> need to re-encrypt everything and I don't know a good way to do it to many secrets once it looks like I'll have to automate it to find all the secrets on kubernetes and re-encrypt it right ? or there is a flag ?

The key renewal feature doesn't require re-encrypting everything.

Could you please help me figure out what part of the documentation makes you think that the automatic renewal feature requires you to re-encrypt everything?

The backup process described in https://github.com/bitnami-labs/sealed-secrets#how-can-i-do-a-backup-of-my-sealedsecrets is fully compatible with automatic key renewal; obviously you have to backup periodically.

weslleycamilo:
One of the places I got its doubt was 
- https://github.com/bitnami-labs/sealed-secrets#re-encryption-advanced which say "Before you can get rid of some old sealing keys you need to re-encrypt your SealedSecrets with the latest private key)."
[...]

Issue #298, opened 2019-10-28
Manual key renewal


guylabs:
Hi

What we try to achieve is to be able to manually update the key on the controller and not rely on the automatic key renewal. We have our own credentials cycling process and we want to know what key pair is used to encrypt the sealed secrets.

We setup the controller with the `--key-renew-period=0` flag to disable the key renewal. We then retrieve the current key pair via [this](https://github.com/bitnami-labs/sealed-secrets#how-can-i-do-a-backup-of-my-sealedsecrets) command and store them in our infrastructure.

Now we see the following workaround to make the manual key renewal working:

1. Backup the key as described [here](https://github.com/bitnami-labs/sealed-secrets#how-can-i-do-a-backup-of-my-sealedsecrets)
2. Create a new key pair and add the Base64 encoded private and public key pair to the `tls.key` and `tls.crt` entries of the backuped `master.json`
3. Reencrypt all sealed secrets via `kubeseal --cert new-key-pair.pub --re-encrypt < old-sealed-secret.json > new-sealed-secret.json`
4. Restore the `master.json` with the new key pair, restart the controller and then push all new encrypted sealed secrets.

Would that process work or do you see a better solution? Does step 3. work with reencrypting a sealed secret with a specified certificate via `--cert`?

Thanks in advance!

mkmik:
re-encrypt is a controller-side operation. The controller will re-encrypt the sealed secret using the latest encryption key available in the cluster.

Are you aware that the sealed-secret controller can operate with multiple keys?

Currently forcing generating a new key-pair is a bit clumsy (it requires sending a HUP signal and the docker image currently lacks the `kill` command).
In https://github.com/bitnami-labs/sealed-secrets/issues/199#issuecomment-515497780 I suggested a possible solution

mkmik:
Would that work for you?

1. add a `renew-if-older-than` flag/env var to your controller deployment specifying the current timestamp

2. wait until the controller has restarted with this new flag/env vare

3. optionally re-encrypt all sealed-secrets with `kubeseal --re-encrypt ...` if you plan to GC your old keys.

---

That said, I'd like to better understand why do you want to re-encrypt all the existing secrets.

You mention you have a "credentials cycling process". Are you also cycling the credentials that you have sealed with sealed secrets or do you only plan to cycle the sealed-secrets encryption keys?

guylabs:
[...]

PR #358, opened 2020-02-19
feat: update the documentation to include details of how to use its o…

weslleycamilo:
Related to its Issue/request: https://github.com/bitnami-labs/sealed-secrets/issues/357

Common needs in the community and where I got to test my own sealed key: https://github.com/bitnami-labs/sealed-secrets/issues/298

### DIFF
--- a/README.md
+++ b/README.md
@@ -401,6 +401,21 @@ It doesn't help that this feature has been historically called "key rotation", w
 Sealed secrets are not automatically rotated and old keys are not deleted
 when new keys are generated. Old sealed secrets resources can be still decrypted (that's because old sealing keys are not deleted).
 
+#### Using your own keys
+
+> It's not recommended to use it since the best security would be provived by renewing its sealed keys. But it would be helpful in a disaster recovery so you don't have to re-encrypt everything or lose any secrets.
+
+- You need to disable the sealed key renew to use it. once disabled use the following commands.
+
+```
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -401,6 +401,21 @@ It doesn't help that this feature has been historically called "key rotation", w
 Sealed secrets are not automatically rotated and old keys are not deleted
 when new keys are generated. Old sealed secrets resources can be still decrypted (that's because old sealing keys are not deleted).
 
+#### Using your own keys
+
+> It's not recommended to use it since the best security would be provived by renewing its sealed keys. But it would be helpful in a disaster recovery so you don't have to re-encrypt everything or lose any secrets.
```
mkmik: I don't understand.

Using your own key doesn't change the "problems" some people have with automatic renewal.

I guess what really matters for people who are afraid of their backups to become stale, is not which private keys are on the cluster, but which certificates people use to seal the secrets!

`kubeseal --fetch-cert` will always return the latest one, but you don't have to use that if you want.
You can use "offline signing" with `--cert` and use any certificate (i.e. public key) you want; it can be a certificate for a key you created or the certificate for the first key that was created when the cluster was created. 

Thus, I'm really confused about this PR.

----

I have the feeling that one feature that would really help is a way for admins to control what is the "current" key (confusingly, current != active).
Currently, the "current" key is the "latest" key.

I guess it would help if there was a way to manually set the current key after you successfully ran a backup operation.

mkmik: do you have a periodic backup automation? if yes, how often does it run?

PR #379, opened 2020-03-22
Add recommended labels

emmanuelCarre:
Hi,

First, thank you for this great tool :) 

I am interested about out of the box labels to be able to manage my kubernetes resources. I added only ones define here: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

I am not sure to use the best way to add them; so let me know if I should change. 

Some additional information about my change: 
* I didn't add `app.kubernetes.io/instance` label because I didn't know witch value use.
* I didn't apply labels to CRD. I didn't enough worked with this kind of object to to know if I should or not.
*  I add dedicated field for `app.kubernetes.io/managed-by` label to be able to override it easily. There many tools over jsonnet so I am not able to define a value. In my use case, it is [tanka](https://tanka.dev/).

### DIFF
--- a/controller-norbac.jsonnet
+++ b/controller-norbac.jsonnet
@@ -12,12 +12,23 @@ local namespace = 'kube-system';
   crd: kube.CustomResourceDefinition('bitnami.com', 'v1alpha1', 'SealedSecret'),
 
   namespace:: { metadata+: { namespace: namespace } },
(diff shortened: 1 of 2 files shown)


mkmik:
Thanks. The tests fail due to the way we create multi-arch containers during CI tests (we require a push to the docker registry). We'll eventually fix that.

Issue #346, opened 2020-01-23
Update status field from sealedsecret


benjamin-bergia:
I would like the controller to populate and update the status field of the sealedSecrets that have been created.
Currently when the controller fails to decrypt a sealedSecret or push a secret, the only way to troubleshoot or even realize that something is wrong is to go through the controller logs.

Exposing the status/health of the sealedSecret would simplify troubleshooting but also allow tools that rely on this field for monitoring resources (as ArgoCD for example) to support sealedSecrets.

mkmik:
Good idea!

 Just for the record though: not only the logs, but also the events (see `kubectl describe`)

benjamin-bergia:
@mkmik yeah that's true, my bad.

torstenwalter:
This would help a lot. Sealed secrets are often used in conjunction with tools like ArgoCD. There a status would help as it would allow to add a health check which provides visual feedback if a SealedSecret could not be decrypted.

dewe:
When using Flux CD this could potentially introduce a loop: 

1. Flux applies manifests and create a sealed secret. 
2. The controller unseal the secret and update the sealed secret. 
3. Flux applies the manifests (at sync interval, typically every 5 minutes) which will restore the sealed secret
4. The controller unseal the secret and update the sealed secret again. 
5. Go to 3.

So, please make this optional and turned off by default.

torstenwalter:
@dewe It won't introduce a loop except if you would store the status field in your got repository which you should not do.

Status fields are common in Kubernetes. They are not managed by the user, but updated by the controller which manages an object. Just look at https://kubernetes.io/docs/concepts/workloads/controllers/deployment/#deployment-status
That does not create a loop for flux either. So from my point of view there is no need to make it optional.

Crevil:
Yeah, to my understanding that’s what we get by `kubectl apply`. Flux obly controls the fields you specify in your source repo.

I’d love a status field on the objects. It makes it even simpler for users to understand what is happening to their resources. 🙏

dewe:
Well, then yes, that makes perfect sense. 👍

benjamin-bergia:
[...]

PR #380, opened 2020-03-23
Upgrade k8s client to v0.16.8

mkmik:
And regenerate client.

The new client has the UpdateStatus method which will be used to address #346


pkg/apis/sealed-secrets/v1alpha1/register.go
```diff
@@ -16,6 +16,8 @@ var (
 
 	// SchemeBuilder adds this group to scheme
 	SchemeBuilder = runtime.NewSchemeBuilder(addKnownTypes)
+	// AddToScheme is a global function that registers this API group & version to a scheme
+	AddToScheme = SchemeBuilder.AddToScheme
```
atomatt: Is this used? Would it make sense to split it out to a different PR?

pkg/apis/sealed-secrets/v1alpha1/doc.go
```diff
@@ -1,5 +1,5 @@
 // go mod vendor doesn't preserve executable perm bits
-//go:generate bash ../../../../vendor/k8s.io/code-generator/generate-groups.sh all github.com/bitnami-labs/sealed-secrets/pkg/client github.com/bitnami-labs/sealed-secrets/pkg/apis sealed-secrets:v1alpha1 --go-header-file boilerplate.go.txt
+//go:generate bash ../../../../vendor/k8s.io/code-generator/generate-groups.sh deepcopy,defaulter,client,lister github.com/bitnami-labs/sealed-secrets/pkg/client github.com/bitnami-labs/sealed-secrets/pkg/apis sealed-secrets:v1alpha1 --go-header-file boilerplate.go.txt
```
atomatt: I assume this was necessary due to an updated `generate-groups.sh` script? It looks like not using `all` means no `informer` generator, is that correct?

mkmik: yeah, the generated informer is buggy: it generates the file as `package sealed-secret` which then fails to gofmt and aborts.

I created this PR mostly to make use of the CI integration test to check I'm on the right track.

I'll find a way to correctly generate the informer with the new version of the generator before merging this

pkg/apis/sealed-secrets/v1alpha1/register.go
mkmik: it seems that github swallowed my answers.

k8s code-generator requires some boilerplate; this is now part of that boilerplate

atomatt: +1 thanks for checking

PR #381, opened 2020-03-25
Update status subresource

mkmik:
Part of #346

TL;DR we now update the CRD status using `UpdateStatus`, updating `ObservedGeneration` correctly, but we don't yet write out any status "conditions".

---

We already had some code to update the `status` field of the CRD (`func updateSealedSecretStatus`).
It was buggy but the bug wasn't triggered because the write to the status field was always a no operation.

The `Status.ObservedGeneration` field is designed to reflect the generation number of the customer resource that the controller has last processed. We used to write the generation number of the new unsealed secret resource we're creating in memory, which is always 0.

Once we fixed bug, though, each `.Update` of the SealedSecret CRD would create a feedback loop because the update to the ObservedGeneration field will cause an increment of the `Generation` field and so on.

This PR, switches to `.UpdateStatus`, since now the `CustomResourceSubresources` feature gate is enabled
in all supported k8s versions (see table https://kubernetes.io/docs/reference/command-line-tools-reference/feature-gates/).

This requires a change in our RBAC configs since the controller needs to be able to update the status subresource.

In the next PR we'll write out actual "status conditions" (success, error, message, etc)

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -197,7 +197,7 @@ func (c *Controller) processNextItem() bool {
 	return true
 }
 
-func (c *Controller) unseal(key string) error {
+func (c *Controller) unseal(key string) (unsealErr error) {
 	unsealRequestsTotal.Inc()
 	obj, exists, err := c.informer.GetIndexer().GetByKey(key)
 	if err != nil {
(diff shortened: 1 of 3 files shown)


cmd/controller/controller.go
```diff
@@ -228,6 +228,16 @@ func (c *Controller) unseal(key string) error {
 	ssecret := obj.(*ssv1alpha1.SealedSecret)
 	log.Printf("Updating %s", key)
 
+	// any exit of this function at this point will cause an update to the status subresource
+	// of the SealedSecret custome resource. If the error is non-nil it will
```
atomatt: "it will ..."? :)

cmd/controller/controller.go
```diff
@@ -197,7 +197,7 @@ func (c *Controller) processNextItem() bool {
 	return true
 }
 
-func (c *Controller) unseal(key string) error {
+func (c *Controller) unseal(key string) (err error) {
```
atomatt: Is the named return used? I can't see a naked `return` which makes this change look a bit accidental. Or ... perhaps the new deferred bit is supposed to leave this set on the way out?

mkmik: the deferred bit reads the value of the return value and passes it down to `c.updateSealedSecretStatus(ssecret, err)`

I changed the code so that it's obvious what is happening without having to deeply understand how Go variable scopes work.

atomatt: Thanks! It seems the subtlety got lost in the a sea of `err` before, renaming it helps it stand out.

mkmik:
the https://travis-ci.org/github/bitnami-labs/sealed-secrets/builds/666826347 build suceeded but apparently travis is not able to update GH status and thus bors can't proceed; merging manually

PR #383, opened 2020-03-26
Tag Status field as omitempty

mkmik:
Checked that nobody depends on the Status field since we're technically breaking source backward compat
https://pkg.go.dev/github.com/bitnami-labs/sealed-secrets@v0.11.0/pkg/apis/sealed-secrets/v1alpha1?tab=doc#SealedSecretStatus

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -288,10 +288,12 @@ func (c *Controller) unseal(key string) (unsealErr error) {
 
 func (c *Controller) updateSealedSecretStatus(ssecret *ssv1alpha1.SealedSecret, unsealError error) error {
 	ssecret = ssecret.DeepCopy()
-	st := &ssecret.Status
+	if ssecret.Status == nil {
+		ssecret.Status = &ssv1alpha1.SealedSecretStatus{}
+	}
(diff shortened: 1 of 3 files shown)


cmd/controller/controller.go
```diff
@@ -288,7 +288,11 @@ func (c *Controller) unseal(key string) (unsealErr error) {
 
 func (c *Controller) updateSealedSecretStatus(ssecret *ssv1alpha1.SealedSecret, unsealError error) error {
 	ssecret = ssecret.DeepCopy()
-	st := &ssecret.Status
+	st := ssecret.Status
+	if st == nil {
+		st = &ssv1alpha1.SealedSecretStatus{}
+	}
+	ssecret.Status = st
```
atomatt: Maybe ...

```
if ssecret.Status == nil {
    ssecret.Status = &ssv1alpha1.SealedSecretStatus{}
}
st := ssecret.Status
```

Could even drop `st` now the secret's `Status` is a pointer value, although it does reduce a bit of line noise.

mkmik: yeah it was ugly

PR #388, opened 2020-03-31
Put status update feature behind a disabled feature flag

mkmik:
A few people have reported that upgrading to v0.12.0 results in a feedback loop where the controller
keeps updating the sealed secrets resources without any respite.

The v0.12.0 release also ships the prometheus feature thus some users will love to update but currently can't.

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -21,6 +21,7 @@ import (
 	"k8s.io/client-go/tools/cache"
 	"k8s.io/client-go/tools/record"
 	"k8s.io/client-go/util/workqueue"
+	"k8s.io/klog"
 
 	ssv1alpha1 "github.com/bitnami-labs/sealed-secrets/pkg/apis/sealed-secrets/v1alpha1"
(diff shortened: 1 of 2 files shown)


jkleinlercher:
do you know the reason for this? Is this happening in every installation or just in rare circumstances?

mkmik:
I didn't manage to reproduce this yet

mkmik:
Users reported this on slack

snuggie12:
We tried turning this on in dev and saw the CPU jump from like `0.01` to `0.4`. The logs show that the only thing changing is `status.conditions[0].lastUpdateTime` (and `metadata.resourceVersion`).

I suspect that when reconciling you need to check that the submitted resource's `metadata.generation` !=  the live state's `metadata.generation` as this is not increasing due to the fact that `metadata` and `status` are ignored.

This is just a shot in the dark, but I think that check would go [here](https://github.com/bitnami-labs/sealed-secrets/blob/07df9fa6f723b8c482c1ea703e7e9eb4d7f74922/cmd/controller/controller.go#L235)?

PR #394, opened 2020-04-06
fix(kubeseal): remove kubeconfig deps from recovery-unseal

### LINKED ISSUE #341: kubeseal should not require a `~/.kube/config` when unsealing secrets when given a key
`kubeseal --recovery-unseal --recovery-private-key master.key -o yaml < some_secret.yaml` fails with error message `error: invalid configuration: no configuration has been provided` if you don't have a `~/.kube/config` file.

I'd assume the fix would be similar to that of https://github.com/bitnami-labs/sealed-secrets/issues/101

I actually managed to circumvent the problem by just adding the `--cert` option to the command, like this:
`kubeseal --recovery-unseal --recovery-private-key master.key -o yaml --cert secrets.pem < some_secret.yaml`, which apparently triggers the fix from the issue linked above so that the command runs successfully also without the `~/.kube/config` file.

If you're wondering why I need/want to run it like this, without the `~/.kube/config`, the answer is that I've created a CLI tool that wraps this command (and the command for sealing) for ease of use and I want to have tests that verify it's able to seal and unseal the secrets correctly. The tests are of course using a dummy keypair for this purpose and the tests run in an evironment that doesn't have access to any kubernetes cluster.

Experienced with `kubeseal version: v0.9.6`

saviogl:
`openCert()` in `kubeseal` triggers a kubeconfig check as it needs to fetch certificate information
from the running kubernetes cluster. Some people have used the `--cert` option as a way of
circumvent that check as it allows the kubeconfig check on `openCert()` to be bypassed by using a
previously backed up certificate. This commit applies a very basic "fix" by moving the `unseal`
command before the `openCert()` execution as it is a fairly self-sufficient command. 

fix #341

### DIFF
--- a/cmd/kubeseal/main.go
+++ b/cmd/kubeseal/main.go
@@ -587,6 +587,13 @@ func run(w io.Writer, secretName, controllerNs, controllerName, certURL string, 
 		}
 	}
 
+	if unseal {
+		return unsealSealedSecret(w, os.Stdin, scheme.Codecs, privKeys)
+	}
+	if len(privKeys) != 0 && isatty.IsTerminal(os.Stderr.Fd()) {
+		fmt.Fprintf(os.Stderr, "warning: ignoring --recovery-private-key because unseal command not chosen with --recovery-unseal\n")
+	}
+
 	if validateSecret {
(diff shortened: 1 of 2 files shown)


mkmik:
thanks!

PR #401, opened 2020-04-28
Correct ineffassign

gliptak:


gliptak:
Build fail seems to be unrelated
https://travis-ci.org/github/bitnami-labs/sealed-secrets/jobs/680534978
```
kubectl: Correctly Configured: pointing to minikube-vm at 10.20.1.57

WARNING! Using --password via the CLI is insecure. Use --password-stdin.

Error response from daemon: Get https://quay.io/v2/: unauthorized: Could not find robot with username: bitnami+sealedsecrets_travis and supplied password.
```

mkmik:
yeah don't worry about the test failure; it's a regression introduced when we moved to multi-arch images (to build multi-arch manifests you need to push the per-arch images somewhere first and that requires creds which are not available in forks); will fix later

PR #421, opened 2020-06-22
Add `condition_info` metric to expose SealedSecrets status

sa-spag:
Hello there 👋 

In order to notify when a specific SealedSecret cannot be synced, I added new Gauges that expose SealedSecrets status. One may create an alert rule such as:
```yaml
alert: SealedSecretSyncFailed
expr: sealed_secrets_controller_condition_info{condition="Synced"} == -1
annotations:
  summary: The Sealed Secrets controller failed to sync {{ $labels.namespace }}:{{ $labels.name }}
```

I have not put any effort in extending the Prometheus mixin with this new metric yet. Happy to get into it if that's a must have.

### DIFF
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -239,6 +239,8 @@ func (c *Controller) unseal(key string) (unsealErr error) {
 			// Non-fatal.  Log and continue.
 			log.Printf("Error updating SealedSecret %s status: %v", key, err)
 			unsealErrorsTotal.WithLabelValues("status").Inc()
+		} else {
+			ObserveCondition(ssecret)
 		}
 	}()
 
--- a/cmd/controller/metrics.go
+++ b/cmd/controller/metrics.go
@@ -1,11 +1,25 @@
 package main
 
 import (
+	"github.com/bitnami-labs/sealed-secrets/pkg/apis/sealed-secrets/v1alpha1"
 	"github.com/prometheus/client_golang/prometheus"
+	v1 "k8s.io/api/core/v1"
 )
 
 // Define Prometheus Exporter namespace (prefix) for all metric names
 const metricNamespace string = "sealed_secrets_controller"
+
+const (
+	labelNamespace = "namespace"
+	labelName      = "name"
+	labelCondition = "condition"
+)
+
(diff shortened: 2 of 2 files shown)


mkmik:
I'm always a bit wary of putting resource names as labels. I know that in most cases it doesn't generate an excessive load to the TSDB, but technically each secret becomes a new timeseries in prometheus and depending on the size of the cluster the prometheus admin could be unhappy.

I'm not sure prometheus is the right place to get this level of debug info. That said, if this was a private application I develop and run in my cluster, I'd just do it, since it's a pragmatic thing to do; I'm just not sure if this is the correct behavior of an infrastructure component that needs to work
in many different environments.

Perhaps a gauge metric (perhaps per-namespace?) that counts the total amount of secrets in each state.

sa-spag:
I understand your concerns. Note also that as new conditions are exposed (I can't think of any but who knows), the number of Gauges also increases.

It's heavily inspired on [what's being done with the Helm Operator](https://docs.fluxcd.io/projects/helm-operator/en/stable/references/monitoring/#release-conditions), which, in my case at least, turns out to be pretty handy. I'm not aware of a specific pattern that has been broadly adopted by Kubernetes operators.

My personal use case applies on a multi tenant cluster, where teams own their namespaces and SealedSecrets. Ultimately the goal is to be able to fire alerts for specific teams. The alternative you suggest would work too, but I'm not sure it makes that much of a difference (e.g. we have an average of 6 SealedSecrets per namespace). Also, I like the straightforwardness of being able to pinpoint the faulty SealedSecrets right from the alert.

mkmik:
the difference is that namespace names are generally constant, while secret names can contain a random suffix (e.g. a hash, thus over time each individual secret resource allocates a timeseries, even if at a given time you only have 6 per namespace)

mkmik:
just to be clear, I'm not against it; I'm just cautious; I would like to hear more opinions

n0rad:
If you look at prometheus metrics exposed by kube or istio for example, they shamelessly add pod names to metrics. With nodes reboot, rolling restart, etc..., timeseries number is pretty high.

Secrets (and even more sealedSecrets) evolves way way less that pods do, so to me it will be a mere drop in the ocean.

mkmik:
ok, I'm sold

PR #438, opened 2020-07-28
Make the sealedsecrets.bitnami.com/managed annotation value more correct

coderanger:
Annotation values must be strings, and `true` in YAML is a boolean.


mkmik:
Thanks

PR #439, opened 2020-07-29
Add -f and -w flags as an alternative to stdin/out

### LINKED ISSUE #405: re-encrypt in place
add a `-f` flag to pass to `re-encrypt` so that people don't have to learn:

```
kubeseal --re-encrypt <$file | sponge $file
```

and instead just do:

```
kubeseal --re-encrypt -f $file
```

Other useful possibilities:

1. detect format (filename extension? content-base heuristic?) so that people that have yaml files don't re-encrypt them to json and the other way around
3. accept a non-sealed-secret file as input and do nothing (only log), this way people can `find -name "*.yaml" -exec kubeseal --re-encrypt -f {} \;` or equivalent
2. support `-f dir` and ignore all non sealed-secret files

mkmik:
Also closes #405

This PR intentionally doesn't document the new flags in the README since
those changes would be visible before this PR gets released.


cmd/kubeseal/main.go
```diff
@@ -586,25 +590,59 @@ func run(w io.Writer, secretName, controllerNs, controllerName, certURL string,
...
+		}
+
+		var f *renameio.PendingFile
+		f, err = renameio.TempFile("", outputFileName)
+		if err != nil {
+			return err
+		}
+		defer func() {
+			if err == nil {
+				f.CloseAtomicallyReplace()
```
aruiz14, line 626: I believe it's worth adding a comment mentioning that this is preventing the output file from being written if the overall function result is an error.

mkmik: done

PR #446, opened 2020-08-17
add support for templated data

### LINKED ISSUE #445: add ability to template arbitrary data keys within resulting secrets
sometimes when deploying services on k8s you end up needing to [create a secret which contains a entire config file.](https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/user-guides/alerting.md) in such cases you often only have a small amount of data that is actually secret, the rest of the config file could be safely kept in plaintext. in such cases `sealed-secrets` could offer a large usability win by allowing arbitrary keys within the resulting secrets to be created from a plaintext template which uses the decrypted data keys as inputs. 

`sealed-secrets` already has the concept of secret "templating", perhaps a `TemplatedData` field could be added to the `SecretTemplateSpec` to enable this.

---
**Example** (extended from the README):
```
--- 
kind: SealedSecret
metadata: 
  annotation: 
    kubectl.kubernetes.io/last-applied-configuration: "...."
  name: mysecret
  namespace: mynamespace
spec: 
  encryptedData: 
    dockerAuth: AgBy3i4OJSWK+PiTySYZZA9rO43cGDEq.....
  template: 
    metadata: 
      annotations: 
        jenkins.io/credentials-description: "credentials from Kubernetes"
      labels: 
        jenkins.io/credentials-type: usernamePassword
    templatedData: 
      .dockercfg: |-
          {
            "auths": {
              "https://index.docker.io/v1/": {
                "auth": "{{index .Data "dockerAuth"}}"
              }
            },
            "HttpHeaders": {
              "User-Agent": "Docker-Client/19.03.9-ce (linux)"
            }
          }
```
would result in a secret like:
```
apiVersion: v1
kind: Secret
metadata:
  name: mysecret
  namespace: mynamespace
  labels:
    "jenkins.io/credentials-type": usernamePassword
  annotations:
    "jenkins.io/credentials-description": credentials from Kubernetes
  ownerReferences:
    controller: true
    kind: SealedSecret
    name: mysecret
    uid: 5caff6a0-c9ac-11e9-881e-42010aac003e
type: kubernetes.io/dockercfg
data:
  dockerAuth: ewogICJjcmVk...
  .dockercfg: ewoJImF1dGh...
```

where the decoded `.dockercfg` looks like:
```
{
	"auths": {
		"https://index.docker.io/v1/": {
			"auth": "hello world..."
		}
	},
	"HttpHeaders": {
		"User-Agent": "Docker-Client/19.03.9-ce (linux)"
	}
}
```

evq:
fixes #445. the new templating support has been tested against a sandbox cluster.

I also updated the config templating docs example to reflect built-in support.


pkg/apis/sealed-secrets/v1alpha1/sealedsecret_expansion.go
```diff
@@ -264,6 +266,7 @@ func (s *SealedSecret) Unseal(codecs runtimeserializer.CodecFactory, privKeys ma
 		secret.Type = s.Spec.Template.Type
 
 		secret.Data = map[string][]byte{}
+		data := map[string]string{}
```
mkmik, line 269: what if the secret data is not a valid utf-8 string?

mkmik (commented):
I like the idea, thanks! I have an inline comment about binary secrets

Also, we need to add some tests before merging this.

pkg/apis/sealed-secrets/v1alpha1/sealedsecret_expansion.go
mkmik: friendly ping

mkmik:
merged in #580

PR #451, opened 2020-09-15
Don't require a namespace when using --raw and cluster-wide scope

### LINKED ISSUE #393: "kubeseal --raw --scope cluster-wide" requires namespace
I would expect the following to work

```bash
echo -n secretText| kubeseal --raw --from-file=/dev/stdin --scope cluster-wide --cert mycert.pem
```
but it responds with 

```bash
error: invalid configuration: no configuration has been provided
```

Adding a namespace fixes the problem.

```bash
echo -n secretText | kubeseal --raw --from-file=/dev/stdin --scope cluster-wide --cert mycert.pem --namespace something
```
I do not see why a namespace is required when scope is cluster-wide.

blacksails:
This will skip trying to determine the namespace when using `--raw` and
the `cluster-wide` scope.

Fixes #393

### DIFF
--- a/cmd/kubeseal/main.go
+++ b/cmd/kubeseal/main.go
@@ -667,17 +667,23 @@ func run(w io.Writer, inputFileName, outputFileName, secretName, controllerNs, c
 	}
 
 	if raw {
-		ns, _, err := namespaceFromClientConfig()
-		if err != nil {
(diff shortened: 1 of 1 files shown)


mkmik (approved):
Thanks

PR #453, opened 2020-09-18
Improve error message when encrypting with wrong scope

mkmik:
CC: @swade1987

### DIFF
--- a/pkg/apis/sealed-secrets/v1alpha1/sealedsecret_expansion.go
+++ b/pkg/apis/sealed-secrets/v1alpha1/sealedsecret_expansion.go
@@ -41,8 +41,8 @@ type SealedSecretExpansion interface {
 // in which scopes
 type SealingScope int
 
-func (s *SealingScope) String() string {
-	switch *s {
+func (s SealingScope) String() string {
+	switch s {
 	case StrictScope:
 		return "strict"
 	case NamespaceWideScope:
@@ -50,7 +50,7 @@ func (s *SealingScope) String() string {
 	case ClusterWideScope:
 		return "cluster-wide"
 	default:
-		return fmt.Sprintf("undefined-%d", *s)
+		return fmt.Sprintf("undefined-%d", s)
 	}
 }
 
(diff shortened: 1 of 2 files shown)


jjo (approved):
LGTM, thanks for this.

pkg/apis/sealed-secrets/v1alpha1/sealedsecret_expansion.go
```diff
@@ -247,17 +247,37 @@ func NewSealedSecret(codecs runtimeserializer.CodecFactory, pubKey *rsa.PublicKe
...
+		return plaintext, nil
+	}
+
+	// Sometimes it happens that users encrypt the secret with a scope and then later change the scope.
+	// Let's not give them a helpless error message "cannot decrypt the key" and instead do some work
+	// on their behalf: try to see if there is any other scope that would have worked.
+	// Performing this work is safe since the user could (and would) do it anyway to troubleshoot why it doesn't decrypt.
+	scopes := []SealingScope{StrictScope, NamespaceWideScope, ClusterWideScope}
+	for _, s := range scopes {
+		label := EncryptionLabel(smeta.GetNamespace(), smeta.GetName(), s)
```
absoludity, line 268: Not sure it's worth it, but rather than shadowing `label` here, you could use `labelOtherScope` and then use it to avoid retesting the original error (and, depending on possible errors, avoid a "secret item encrypted with scope 'Strict', being decrypted with scope 'Strict`")

pkg/apis/sealed-secrets/v1alpha1/sealedsecret_expansion.go
```diff
@@ -247,17 +247,37 @@ func NewSealedSecret(codecs runtimeserializer.CodecFactory, pubKey *rsa.PublicKe
...
+		if _, err := crypto.HybridDecrypt(rand.Reader, privKeys, valueBytes, label); err == nil {
+			return nil, fmt.Errorf("secret item encrypted with scope %q, being decrypted in scope %q", s.String(), SecretScope(smeta).String())
```
absoludity, line 270: Wouldn't it be worth wrapping the original error here? (again, it depends on possible errors, in particular, whether reaching this line means it's definitely because of the scope, or whether there can be other intermittent errors or similar).

absoludity: But more importantly, I don't think you want to shadow `err` here, as if the error is unrelated to scope (and so all the checks for decrypting with other scopes here fail), you'll have overwritten the original error which will be returned below, I think?

absoludity (approved):
Nice mkm. One small thought about being able to return the original error when it's not related to the scope.

PR #375, opened 2020-03-12
Prometheus metrics & mixin

kskewes:
Rebased Prometheus PR on master.

Closes: #177 

- add http handler on `/metrics`. 
  There aren't any sensitive endpoints on this port, but if metrics are considered sensitive for some then I can move to a custom port.
- add standard `build_info` and `go_build_info` metrics
- add custom unseal error & request metrics - RE(D)
- add mixin with:
  - Grafana dashboard
  - Alerts
- add `PodMonitor` jsonnet for those using Prometheus Operator (us)

Additional thoughts:
- `make test` is currently failing for me on master as well as the `prometheus` branch: 
  `--- FAIL: TestHttpCert (1.37s)    `
- I did the bulk of this before key rotation was done and didn't experience any errors due to prometheus integration. 
- Key rotation metrics/dashboard/alert could be easily added in a subsequent PR.

Testing done:
- with `kind`
- create sealed secret - observe request counter increment
- submit SS with new secret - observe request counter increment
- submit SS with modified namespace (incorrect) - observe error increment
- submit SS with modified encrypted value (incorrect) - observe error increment
- delete secret - observe request counter increment
- delete SS - observe request counter increment

PR #463, opened 2020-10-21
add namespace to unseal error metric

povilasv:
I've been running sealed secrets and I love it thanks!

One caveat is that it's impossible to direct `SealedSecretsUnsealErrorRateHigh` alert to specific teams. So it ends up that one team, gets all of the failures.

I think to solve it we can add the namespace to the metric so that alerts can be directed to teams that own approperiate namespaces.

One typical example is when a secret already exists and then you add a new sealed secret and it doesn't override the existing secret. In this case I would like to alert the appropriate namespace owner ;)

### DIFF
--- a/cmd/controller/metrics.go
+++ b/cmd/controller/metrics.go
@@ -47,7 +47,7 @@ var (
 			Name:      "unseal_errors_total",
 			Help:      "Total number of sealed secret unseal errors by reason",
 		},
-		[]string{"reason"},
+		[]string{"reason", "namespace"},
 	)
 
 	conditionInfo = prometheus.NewGaugeVec(prometheus.GaugeOpts{
@@ -64,11 +64,6 @@ func init() {
 	prometheus.MustRegister(unsealRequestsTotal)
 	prometheus.MustRegister(unsealErrorsTotal)
 	prometheus.MustRegister(conditionInfo)
-
-	// Initialise known label values
-	for _, val := range []string{"fetch", "status", "unmanaged", "unseal", "update"} {
-		unsealErrorsTotal.WithLabelValues(val)
-	}
 }
 
 // ObserveCondition sets a `condition_info` Gauge according to a SealedSecret status.
--- a/cmd/controller/controller.go
+++ b/cmd/controller/controller.go
@@ -208,7 +208,7 @@ func (c *Controller) unseal(key string) (unsealErr error) {
 	obj, exists, err := c.informer.GetIndexer().GetByKey(key)
 	if err != nil {
 		log.Printf("Error fetching object with key %s from store: %v", key, err)
-		unsealErrorsTotal.WithLabelValues("fetch").Inc()
+		unsealErrorsTotal.WithLabelValues("fetch", "").Inc()
 		return err
 	}
 
@@ -242,7 +242,7 @@ func (c *Controller) unseal(key string) (unsealErr error) {
 		if err := c.updateSealedSecretStatus(ssecret, unsealErr); err != nil {
 			// Non-fatal.  Log and continue.
 			log.Printf("Error updating SealedSecret %s status: %v", key, err)
-			unsealErrorsTotal.WithLabelValues("status").Inc()
+			unsealErrorsTotal.WithLabelValues("status", ssecret.GetNamespace()).Inc()
 		} else {
 			ObserveCondition(ssecret)
 		}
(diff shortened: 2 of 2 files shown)


cmd/controller/metrics.go
```diff
@@ -64,11 +64,6 @@ func init() {
 	prometheus.MustRegister(unsealRequestsTotal)
 	prometheus.MustRegister(unsealErrorsTotal)
 	prometheus.MustRegister(conditionInfo)
-
```
mkmik, line 67: Why did you have to remove this?

povilasv: becaue we can't populate it any more, as the label combination is status + namespace

mkmik: I don't know why it had to be initialized in the first place.

@kskewes I see you added that in #375; is it really necessary these counters are set to 0?

kskewes: The backstory is here: https://www.robustperception.io/existential-issues-with-metrics

How it relates:

Sealed Secrets Controller behaviour in an error case as I understand it.

1. Startup
2. Attempt unseal but there's error. Increase the associated counter metric (with label).
3. Retry unseal attempt 5 times and then give up.

Prometheus behaviour:

1. Startup
2. Detect Sealed Secrets Controller pod (service discovery)
3. Scrape `/metrics` endpoint on retry interval (10s or 60s or as configured)
    These scrapes can happen at any point after the controller has started up.

Situation:

1. Prometheus client libraries initialise counters with no labels to 0 by default. 
2. Prometheus client libraries don't initialise counters with labels by default as it doesn't know what labels will be set.
    So we do this manually. 
3. Above enables easy alerting rules as you can check for an increase in the counter. An increase from null/non-existing metric to X value is not easily alert-able on. There's an `absent()` function in Prometheus but there are a lot of edge cases we have tried to handle in our other applications with varying degrees of success.

Options (in no specific order):

1. Add the namespace label per this PR as is - this is likely to result in missed alerts as a error metric for a namespace is likely to go from `NaN` to `5` within the scrape period interval due to fast retry loop and Prometheus won't catch it. The second increase (or unseal error) will alert via the existing alert.
2. Add the namespace label per this PR and have a go at rewriting the alert with namespace label and trying to catch `absent()` metrics with `offset` etc per the above link.
3. Add the namespace label and have the controller register the metrics per error reason per namespace. Controller would need to watch Namespaces via Kubernetes API and initialise metric for new namespaces as added. I think cardinality wise this if fine.
4. Do nothing (We'd like the label, but little bit to do per above).

mkmik: Other options:

5. List+watch for namespaces via k8s api and initialize metrics based in existing namespaces.

6. Advise users on how to craft alerts correctly in case of missing values (using `or` and other trickeries as Brian hinted at in the linked doc)

povilasv: :+1: Which one should we implement?

We can do a namespace list and then init this metric, but this would require more privileges?

I feel like the simplistic approach should sort of work, cause we queue in unseals with this config:
```
		NewItemExponentialFailureRateLimiter(5*time.Millisecond, 1000*time.Second),
		// 10 qps, 100 bucket size.  This is only for retry speed and its only the overall factor (not per item)
		&BucketRateLimiter{Limiter: rate.NewLimiter(rate.Limit(10), 100)},
```

So in theory errors metric would increase in time and so Prometheus might see it as:
scrape 0 unseal_errors_total 2
scrape 1 unseal_errors_total 3
scrape 4 unseal_errors_total 4

so simple rate() > 0 would work, no?

mkmik: sounds reasonable; thanks

mkmik (approved):

cmd/controller/metrics.go
kskewes: Cool. The retry interval must have changed since I last tested. :)
With 10s scrape interval provided we have retries over at least 2 it might be ok.

It would be good to update the alert unit test in tests.yaml and the alert itself in jsonnet.
Would you like to do in separate PR? Or I can once we pull next release and can validate.

mkmik: Thanks Karl, I don't have bandwidth to take care of this, it would be really great if I could delegate that to you

kskewes: Sure. Happy to pickup after current tasks. :)
Can validate retry versus scrape intervals in the unit test.

povilasv:
thank you for taking a look and let me know if you have any problems with it, I might be able to help ;)

cmd/controller/metrics.go
simon-anz: Can we initialize at least "fetch" (which doesn't have a namespace) so the counter exists. See https://github.com/bitnami-labs/sealed-secrets/issues/543

Issue #114, opened 2018-09-11
When sealing multiple secrets in singe yaml file only first is sealed


elmariofredo:
Steps to reproduce

```
10:27:56 in ~/Downloads/kubeseal-test
➜ kubectl create secret generic mysecret --dry-run --from-literal=foo=bar -o yaml >mysecret.yaml

10:28:13 in ~/Downloads/kubeseal-test
➜ cat mysecret.yaml                                                                             
apiVersion: v1
data:
  foo: YmFy
kind: Secret
metadata:
  creationTimestamp: null
  name: mysecret

10:28:14 in ~/Downloads/kubeseal-test
➜ echo "---" >> mysecret.yaml

10:28:34 in ~/Downloads/kubeseal-test
➜ kubectl create secret generic mysecret2 --dry-run --from-literal=foo2=bar2 -o yaml >>mysecret.yaml

10:28:56 in ~/Downloads/kubeseal-test
➜ cat mysecret.yaml                                                                       
apiVersion: v1
data:
  foo: YmFy
kind: Secret
metadata:
  creationTimestamp: null
  name: mysecret
---
apiVersion: v1
data:
  foo2: YmFyMg==
kind: Secret
metadata:
  creationTimestamp: null
  name: mysecret2

10:28:59 in ~/Downloads/kubeseal-test
➜ kubeseal --format yaml <mysecret.yaml >mysealedsecret.yaml

10:29:36 in ~/Downloads/kubeseal-test
➜ cat mysealedsecret.yaml 
apiVersion: bitnami.com/v1alpha1
kind: SealedSecret
metadata:
  creationTimestamp: null
  name: mysecret
  namespace: default
spec:
  encryptedData:
[...]

PR #475, opened 2020-11-11
Raise error when encountering multi-document YAML

mkmik:
Users reported in #114 that they expected multi-document YAML files to be sealed and validated.
The current implementation only parses the first document and silently ignores the rest.

Ideally we should handle multiple documents but until then at least we should report an error

### DIFF
--- a/cmd/kubeseal/main.go
+++ b/cmd/kubeseal/main.go
@@ -19,6 +19,7 @@ import (
 
 	"github.com/bitnami-labs/sealed-secrets/pkg/buildinfo"
 	"github.com/bitnami-labs/sealed-secrets/pkg/crypto"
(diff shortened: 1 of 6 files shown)


cmd/kubeseal/main.go
```diff
@@ -143,6 +149,13 @@ func readSecret(codec runtime.Decoder, r io.Reader) (*v1.Secret, error) {
 	return &ret, nil
 }
 
+func isMultiDocumentYAML(src []byte) bool {
```
aruiz14, line 152: We could use this from `pkg/multidocyaml` if it wasn't private.

aruiz14 (approved):

cmd/kubeseal/main.go
mkmik: I moved this function to that package so that it could be used by kubeseal and the contoller

PR #487, opened 2020-12-10
bump crd version

### LINKED ISSUE #486: Update chart to work on newer clusters
Related to #389.
Do a new version which avoids outdated api versions (rbac, deployment etc).
Right now sealed-secrets is not deployable onto a modern (1.19) cluster.
Will probably require an updated CRD as well (not just the api-version).

You can assign me it.

davidkarlsen:
Signed-off-by: David J. M. Karlsen <david@davidkarlsen.com>

---
This change is [Reviewable](https://reviewable.io/reviews/bitnami-labs/sealed-secrets/487)

Closes #486

### DIFF
--- a/helm/sealed-secrets/Chart.yaml
+++ b/helm/sealed-secrets/Chart.yaml
@@ -3,7 +3,7 @@ description: A Helm chart for Sealed Secrets
 version: 1.14.0
 icon: https://avatars0.githubusercontent.com/u/34656521?s=200&v=4
 appVersion: 0.13.1
-kubeVersion: ">=1.9.0-0"
+kubeVersion: ">=1.16.0-0"
 home: https://github.com/bitnami-labs/sealed-secrets
 apiVersion: v1
 maintainers:
--- a/helm/sealed-secrets/Chart.yaml
+++ b/helm/sealed-secrets/Chart.yaml
@@ -1,6 +1,6 @@
 name: sealed-secrets
 description: A Helm chart for Sealed Secrets
-version: 1.13.2
+version: 1.14.0
 icon: https://avatars0.githubusercontent.com/u/34656521?s=200&v=4
 appVersion: 0.13.1
 kubeVersion: ">=1.9.0-0"
--- a/helm/sealed-secrets/Chart.yaml
+++ b/helm/sealed-secrets/Chart.yaml
@@ -5,7 +5,8 @@ icon: https://avatars0.githubusercontent.com/u/34656521?s=200&v=4
 appVersion: 0.13.1
 kubeVersion: ">=1.16.0-0"
 home: https://github.com/bitnami-labs/sealed-secrets
-apiVersion: v1
+apiVersion: v2
+type: application
 maintainers:
 - name: mkmik
   email: mkm@bitnami.com
--- a/helm/sealed-secrets/crds/sealedsecret-crd.yaml
+++ b/helm/sealed-secrets/crds/sealedsecret-crd.yaml
@@ -16,6 +16,12 @@ spec:
     plural: sealedsecrets
     singular: sealedsecret
   scope: Namespaced
-  subresources:
-    status: {}
-  version: v1alpha1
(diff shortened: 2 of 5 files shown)


davidkarlsen:
@mkmik can you add an updated CRD to this PR so we can see it pass?
Maybe we should up the chart-api version to v2 as well? are you ok with that?

helm/sealed-secrets/Chart.yaml
```diff
@@ -1,11 +1,12 @@
 name: sealed-secrets
 description: A Helm chart for Sealed Secrets
-version: 1.13.2
+version: 1.14.0
```
mkmik, line 3: let's not publish this chart yet

helm/sealed-secrets/Chart.yaml
```diff
@@ -1,11 +1,12 @@
...
 icon: https://avatars0.githubusercontent.com/u/34656521?s=200&v=4
 appVersion: 0.13.1
-kubeVersion: ">=1.9.0-0"
+kubeVersion: ">=1.16.0-0"
```
mkmik, line 6: currently we officially support 1.15, but I think we can drop support now; I have to update the CI environment and documentation too though first

helm/sealed-secrets/Chart.yaml
```diff
@@ -1,11 +1,12 @@
...
 home: https://github.com/bitnami-labs/sealed-secrets
-apiVersion: v1
+apiVersion: v2
```
mkmik, line 8: what does this do?

helm/sealed-secrets/Chart.yaml
mkmik: ... yaks have to be shaven ... #492

helm/sealed-secrets/Chart.yaml
mkmik: ok I found: https://helm.sh/docs/topics/charts/#the-apiversion-field

helm/sealed-secrets/crds/sealedsecret-crd.yaml
```diff
@@ -0,0 +1,21 @@
...
+  scope: Namespaced
+  versions:
+  - name: v1alpha1
+    served: true
+    storage: true
+    subresources:
+      status: {}
+    schema:
+      openAPIV3Schema:
+        type: object
```
mkmik, line 21:
```suggestion
      openAPIV3Schema:
        type: object
        properties:
          spec:
            type: object
            x-kubernetes-preserve-unknown-fields: true
            properties: {}
```

helm/sealed-secrets/crds/sealedsecret-crd.yaml
```diff
@@ -0,0 +1,26 @@
...
+        properties:
+          spec:
+            type: object
+            x-kubernetes-preserve-unknown-fields: true
+            properties: {}
```
jsoref, line 26: This line is causing pain for argocd because empty things are odd...

image

mkmik: What do you suggest?

jsoref: First, please note that I'm by no means an expert (I'm unfortunately bumping into all of these things more or less at once).

Can you explain to me what you're trying to say by having this field?

Are you saying that properties is mandatory but can have no items?

For your field, would this be valid:
```yaml
properties: null
```

I'm assuming you're saying that's invalid.

And that you only want:

```yaml
properties:
  ...
```

---
It's possible that something like:

```yaml
properties:
  spec:
    properties:
      type: object
```
might do what you want. It's also possible I'm totally wrong :-(
I'm picking that syntax from https://github.com/kubernetes/apiextensions-apiserver/issues/25#issuecomment-755099112

It's possible that you want:
```yaml
properties:
  spec:
    properties:
      type: object
      nullable: true
```

Issue #504, opened 2021-01-21
`condition_info` no more exposed on a running container


ViBiOh:
# Actual behavior

We have a sealed-secret container that is running since 9 days. It doesn't expose the `condition_info` metrics since 3 days, all others metrics are OK.

Prometheus indicates that the last metrics of this kind received was juste before the rollout of a new deployment where we changed secrets.

# Expected behavior

The `condition_info` are kept when doing a rollout of a new deployment that destroy or creates secrets.

# My own investigation

It seems that [SealedSecret listen on delete event](https://github.com/bitnami-labs/sealed-secrets/blob/master/cmd/controller/controller.go#L128), and [`UnregisterCondition` in that case](https://github.com/bitnami-labs/sealed-secrets/blob/master/cmd/controller/metrics.go#L108).

The method for unregister looks like the observe one. It seems to unregister from all secrets. I guess the expected behavior was to unregister only from one. I don't have deep knowledge in implementation, maybe I'm totally wrong 🤭

PR #508, opened 2021-01-25
Fix `condition_info` metrics that disappear completely when removing one sealed secret

ViBiOh:
This pull-request fixes the https://github.com/bitnami-labs/sealed-secrets/issues/504.

I'm not familiar with vendoring, I've isolated the `vendor/` folder changes on a dedicated commit in case I've made mistake.

The label deletion wanted when deleting a secret is only available since `1.8.0` of `prometheus_client`, that expose the underlying `MetricVec` object. `1.9.0` is the latest so I've opted for being up-to-date.

I've tested it on a local cluster.


mkmik:
thanks!!

PR #521, opened 2021-02-12
Add label-selector to filter sealed secrets

pfrybar:
Allow filtering which sealed secrets this controller acts on by passing in a label selector.


mkmik:
thank you!

integration tests are broken at the moment.
can you run them locally?

mkmik (approved):

mkmik:
thanks! could you also please add a test case for this?

pfrybar:
Hey, yeah I commented about this on slack, but I'll mention here too: do you have a suggestion for how/where to add the test? Ideally the controller would be tested with and without the filter option, but this would require deploying two different versions of the controller, and potentially two separate integration tests.

mkmik:
yeah; let's see how we currently test the (`--all-namespaces`) flag

mkmik:
> yeah; let's see how we currently test the (--all-namespaces) flag

:facepalm: we don't test it; I guess it's a similar problem

mkmik:
ok; let's land this and open a separate issue about testing different configuration of the controller

PR #532, opened 2021-02-18
up minor version of helm chart to ensure #529 change gets released correctly

KlavsKlavsen:
I hope this is correct :)


mkmik:
no, the chart version (minor or patchlevel) changes only when we do a release of the controller image; changes to the helm chart alone are reflected in the  chart revision (`-rN`) suffix only.

Anyway, I'm going to cut a v1.15 release today

mkmik:
(FYI @dbarranco, @vikrantest as you can see this subtle rule is captured in a CI step called "Release automation / Lint")

PR #540, opened 2021-02-25
Allow viewers to read SealedSecrets

mkilchhofer:
Since SealedSecrets only contains encrypted data, we allow reading this kind of resource for
users with view permissions. There's an existing "view" role (besides "edit" and "admin") in every cluster
which gets updated with ClusterRoles containing special aggregate instructions via annotations:
~~~
rbac.authorization.k8s.io/aggregate-to-view: "true"
rbac.authorization.k8s.io/aggregate-to-edit: "true"
rbac.authorization.k8s.io/aggregate-to-admin: "true"
~~~
The view role in my use case is also used by event exporters (ie. the one provided by Opsgenie). All occurring
events are received, enriched with the metadata of the relevant object and forwarded to an arbitrary
monitoring/log system of choice. Without this role the event exporter cannot read the SealedSecrets object.

### DIFF
--- a/helm/sealed-secrets/templates/cluster-role.yaml
+++ b/helm/sealed-secrets/templates/cluster-role.yaml
@@ -40,4 +40,28 @@ rules:
     verbs:
       - create
       - patch
+---
+apiVersion: rbac.authorization.k8s.io/v1
+kind: ClusterRole
+metadata:
+  name: {{ template "sealed-secrets.fullname" . }}-view
+  namespace: {{ template "sealed-secrets.namespace" . }}
+  labels:
+    app.kubernetes.io/name: {{ template "sealed-secrets.name" . }}
+    helm.sh/chart: {{ template "sealed-secrets.chart" . }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/version: {{ .Chart.AppVersion }}
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
+rules:
+  - apiGroups:
+      - bitnami.com
+    resources:
+      - sealedsecrets
+    verbs:
+      - get
+      - list
+      - watch
 {{ end }}


helm/sealed-secrets/templates/cluster-role.yaml
```diff
@@ -40,4 +40,28 @@ rules:
...
+  namespace: {{ template "sealed-secrets.namespace" . }}
+  labels:
+    app.kubernetes.io/name: {{ template "sealed-secrets.name" . }}
+    helm.sh/chart: {{ template "sealed-secrets.chart" . }}
+    app.kubernetes.io/managed-by: {{ .Release.Service }}
+    app.kubernetes.io/instance: {{ .Release.Name }}
+    app.kubernetes.io/version: {{ .Chart.AppVersion }}
+    rbac.authorization.k8s.io/aggregate-to-view: "true"
+    rbac.authorization.k8s.io/aggregate-to-edit: "true"
+    rbac.authorization.k8s.io/aggregate-to-admin: "true"
```
mkmik, line 57: are these really necessary?

mkmik:
I'm a bit confused; I've worked in clusters where I had only read-only access and I remember I was able to see all resources in all the namespaces I had access.

Does this PR adds the ability for people to see SealedResources of namespaces I wouldn't otherwise have access?

mkilchhofer:
This PR adds the ability to gain cluster-wide view access through the existing ClusterRole "view". There are several ways to provide view access.

In a up-to-date cluster there are 3 special ClusterRoles which can be bound with a ClusterRolebinding:
- view
- edit
- admin

Several projects use this pattern:
- [cert-manager](https://github.com/jetstack/cert-manager/blob/master/deploy/charts/cert-manager/templates/rbac.yaml#L422)
- [OLM (operator lifecycle manager)](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/deploy/upstream/manifests/0.17.0/0000_50_olm_09-aggregated.clusterrole.yaml#L28)
- [metrics-server](https://github.com/kubernetes-sigs/metrics-server/blob/master/manifests/base/rbac.yaml#L7)
- (..)

mkmik:
Ok.thanks. can you please undo the version bump to the chart version? I'd like to add this also to the standalone yaml manifest (generated from the controller.jsonnet) and make a new release of both sealed-secrets and the helm chart combined

mkilchhofer:
> can you please undo the version bump to the chart version? 

Done :)
Regarding the change in general, I have no problem when you challenge me, maybe there are other solutions for this. But as I saw how the mentioned three projects solves this, I thought this would be a good idea. And it works for me: https://github.com/mkilchhofer/k8s-gitops/blob/main/clusters/k3s/kube-system/sealed-secrets.yaml#L35

mkilchhofer:
Hi @mkmik ,
Is there anything open for discussing, or is it ready to merge?

mkmik:
@juan131 can you please take a look if this is still relevant?

juan131:
@mkilchhofer sorry but I just saw this.. Would you be willing to reopen this PR so we can reconsider it?

PR #546, opened 2021-03-04
update Helm installation Guide

HamzaZo:
+ Add support for Helm 3 installation


mkmik:
Moved commits to #554 so I can land it

PR #557, opened 2021-03-23
Replace deprecated docker maintainer with label

jsoref:
https://docs.docker.com/engine/reference/builder/#maintainer-deprecated


juan131:
This was done at #668

PR #558, opened 2021-03-23
Add status to crd so health checks work on modern k8s

### LINKED ISSUE #555: Status not updating despite --update-status
First of all, I know very little about how CRDs etc work, so take the following with a salt of grain.
Running k3os with `GitVersion:"v1.19.5+k3s2"`.

When I enable `--update-status`, my SealedSecret objects still weren't getting the `Status` field. I checked everything as in:
- Checked the Go-code was really updating the status: it was
- Checked the argument was really set: it was
- Checked that the status update was send: it was (I revoked the permission, and it was giving an error)
- Checked k3s (which I am using) doesn't magically not support subresources or something: that was not the issue.

In the end, by looking how others did it (longhorn in my case), I found that the `openAPIV3Schema` doesn't set the magic `x-kubernetes-preserve-unknown-fields` to true for `status`. After doing this, it magically works!

```
      openAPIV3Schema:
        properties:
          spec:
            type: object
            x-kubernetes-preserve-unknown-fields: true
          status:
            x-kubernetes-preserve-unknown-fields: true
        type: object
```

I have no clue why this would be the case etc etc, and I am sure you guys will know a lot better what is going on here, but without that addition, I cannot get this to work at all.

Am I right that the schema definition is wrong, or am I just doing something else wrong?

Tnx!

jsoref:
closes #555

### DIFF
--- a/helm/sealed-secrets/crds/sealedsecret-crd.yaml
+++ b/helm/sealed-secrets/crds/sealedsecret-crd.yaml
@@ -23,3 +23,5 @@ spec:
           spec:
             type: object
             x-kubernetes-preserve-unknown-fields: true
+          status:
+            x-kubernetes-preserve-unknown-fields: true


mkmik (approved):

jannfis:
@mkmik I'm not sure how far 0.16.0 is away, but would the CRD fixes justify a 0.15.1 release maybe? Many people that are expecting `--update-status` to work are stumbling over this issue.

PR #587, opened 2021-05-11
make helm chart reference image with tag v0.16.0

patrickhuy:
In https://github.com/bitnami-labs/sealed-secrets/blob/helm-v1.16.0/helm/sealed-secrets/values.yaml#L3 the tag is still set to v0.15.0 I think it should be v0.16.0 instead. Could you check @mkmik

It might be appropriate to add something for this to `linter.go` but I currently don't have the tooling or the time available to do it.


mkmik:
Superseded by #588


Thanks for letting me know about the problem!

PR #601, opened 2021-06-17
Reenable managing crd by helm template

### LINKED ISSUE #600: Updating controller using helm chart removes CRD and SealedSecrets
On my k8s clusters I still have deployed sealed secrets v1.13.2. The installation was done using the helm charts.

I now have tried to upgrade my deployment to the current version 0.16.1 using the corresponding helm chart. I was surprised that helm removed the sealedsecrets.bitnami.com crd and therefore also all sealed secrets.

When checking the helm chart I noticed that the crd [has been removed](https://github.com/bitnami-labs/sealed-secrets/commit/4cfc45b2c806a700e7faa02349317e5f711cd1de) from the helm templates. It seems that this has broken the update.

Is there any reason why the crd has been removed? What is the upgrade path from 0.13.1 to 0.14 or even the current version?

timbrd:
Fixes #600

### DIFF
--- a/helm/sealed-secrets/Chart.yaml
+++ b/helm/sealed-secrets/Chart.yaml
@@ -1,7 +1,7 @@
 name: sealed-secrets
 description: Helm chart for the sealed-secrets controller.
 
-version: 1.16.1
+version: 1.17.0
 appVersion: v0.16.0
 
 kubeVersion: ">=1.16.0-0"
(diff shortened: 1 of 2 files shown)


juan131:
At #687 we moved the CRDs to the `crd/` folder as it's recommended by Helm, see:

- https://helm.sh/docs/chart_best_practices/custom_resource_definitions/

PR #605, opened 2021-06-23
Remove new line from the end of the unencoded secret

fardin01:
Kubeseal adds a new line to the end of (only) some secrets which Kubernetes
does not like. So far I've been able to verify this using an AWS Secret
Access Key.

Using version `0.16.0`, to reproduce the issue:

```
kubeseal --raw --cert <public key> --from-file=aws_secret_access_key.txt --scope cluster-wide
```
or
```
kubeseal --raw --cert <public key path> --from-file=/dev/stdin --scope cluster-wide  <<<  $(echo -n <an AWS Secret Access Key>)
```
then decrypt it using:
```
kubeseal --recovery-unseal --recovery-private-key <private key path> -o yaml < sealed_secret_yaml_manifest_containing_the_secret.yaml
```
and compare the `base64` encoded that the above command returns with the output of `echo -n <an AWS Secret Access Key>  | base64` (if doing this on Ubuntu, you need to run `echo -n <an AWS Secret Access Key>  | base64 -w 0` ).

`sealed_secret_yaml_manifest_containing_the_secret.yaml`:
```
kind: SealedSecret
metadata:
  name: test
  annotations:
    sealedsecrets.bitnami.com/cluster-wide: "true"
spec:
  encryptedData:
    test: <encrypted secret>
```

### DIFF
--- a/cmd/kubeseal/main.go
+++ b/cmd/kubeseal/main.go
@@ -715,7 +715,7 @@ func run(w io.Writer, inputFileName, outputFileName, secretName, controllerNs, c
 		if err != nil {
 			return err
 		}
-
+		data = bytes.TrimSuffix(data, []byte{10})
 		return encryptSecretItem(w, secretName, ns, data, sealingScope, pubKey)
 	}
 


fardin01:
This will probably cause issues if someone is trying to encrypt a certificate like:
```
-----BEGIN CERTIFICATE-----\nMIIIPDCCBySgAwIBAgIQB2XGTnTlkdaAOcoqhHVj8DANBgkqhkiG9w0BAQsFADBw\nMuasSRnjFqefjJadT ...
```

fardin01:
`var data []byte` is populated using either of the following code:
```
data, err = ioutil.ReadFile(filename)
```
or 
```
data, err = ioutil.ReadAll(os.Stdin)
```
and in either case, there is a newline. I'm confused about how this happens 😕

mkmik:
yeah sorry I meant to actually try this out; I was a bit surprised about that

fardin01:
> yeah sorry I meant to actually try this out; I was a bit surprised about that

@mkmik It turned out that my input actually did have a training new line. Apparently redirect to stdin (`<<<`) adds a new line. However, I think it would still be beneficial to double-check for a training new line and remove it.

PR #610, opened 2021-07-17
helm: add ability to add custom environment variable

zufardhiyaulhaq:
Add ability to add custom environment variables to the container deployment.

### DIFF
--- a/helm/sealed-secrets/templates/deployment.yaml
+++ b/helm/sealed-secrets/templates/deployment.yaml
@@ -57,6 +57,12 @@ spec:
           volumeMounts:
           - mountPath: /tmp
             name: tmp
+          {{- if .Values.env }}
+          env:
+            {{- with .Values.env }}
+              {{- toYaml . | nindent 12 }}
+            {{- end }}
+          {{- end }}
           livenessProbe:
             httpGet:
               path: /healthz
--- a/helm/sealed-secrets/values.yaml
+++ b/helm/sealed-secrets/values.yaml
@@ -22,6 +22,11 @@ controller:
 # namespace: Namespace to deploy the controller.
 namespace: ""
 
+# Extra env variables to to be added to controller deployment
+env: []
+# - name: SEALED_SECRETS_UPDATE_STATUS
+#   value: 1
+
 serviceAccount:
   # serviceAccount.create: Whether to create a service account or not
   create: true


zufardhiyaulhaq:
cc @anguslees

helm/sealed-secrets/templates/deployment.yaml
```diff
@@ -57,6 +57,12 @@ spec:
           volumeMounts:
           - mountPath: /tmp
             name: tmp
+          {{- if .Values.env }}
+          env:
+            {{- with .Values.env }}
+              {{- toYaml . | nindent 12 }}
+            {{- end }}
+          {{- end }}
```
juan131, line 65: This is my suggestion for supporting custom env vars from different sources:
```suggestion
         {{- if .Values.extraEnvVars }}
         env: {{- include "sealed-secrets.render" (dict "value" .Values.extraEnvVars "context" $) | nindent 12 }}
          {{- end }}
          {{- if or .Values.extraEnvVarsCM .Values.extraEnvVarsSecret }}
          envFrom:
            {{- if .Values.extraEnvVarsCM }}
            - configMapRef:
                name: {{ .Values.extraEnvVarsCM }}
            {{- end }}
            {{- if .Values.extraEnvVarsSecret }}
            - secretRef:
                name: {{ .Values.extraEnvVarsSecret }}
            {{- end }}
          {{- end }}
```

helm/sealed-secrets/values.yaml
```diff
@@ -22,6 +22,11 @@ controller:
 # namespace: Namespace to deploy the controller.
 namespace: ""
 
+# Extra env variables to to be added to controller deployment
+env: []
+# - name: SEALED_SECRETS_UPDATE_STATUS
+#   value: 1
```
juan131, line 28: Following the previous recommendation:
```suggestion
## @param extraEnvVars Array with extra environment variables to add to controller container
## e.g:
## extraEnvVars:
##   - name: FOO
##     value: "bar"
##
extraEnvVars: []
## @param extraEnvVarsCM Name of existing ConfigMap containing extra env vars for controller container
##
extraEnvVarsCM: ""
## @param extraEnvVarsSecret Name of existing Secret containing extra env vars for controller container
##
extraEnvVarsSecret: ""
```

juan131 (changes requested):
Thanks so much for this PR @zufardhiyaulhaq !! Please check my recommendations when you have a chance.

zufardhiyaulhaq:
@juan131 I am not sure why we only allow a single configmap and secret to be able to mount to the deployment. I think it's not a generic approach

juan131:
What do you mean @zufardhiyaulhaq ? Are you suggesting that `extraEnvVarsCM` and `extraEnvVarsSecret` properties should be an array?

zufardhiyaulhaq:
yes @juan131 wdyt? it's better for array right?

Issue #543, opened 2021-03-01
Prometheus metrics missing


alexanderdalloz:
The sealed secrets controller does not expose all the metrics in the metrics namespace "sealed_secrets_controller_" it should according to the ode in https://github.com/bitnami-labs/sealed-secrets/blob/v0.15.0/cmd/controller/metrics.go.
```
$ curl -s 127.0.0.1:8080/metrics | egrep 'sealed_secrets_controller_'
# HELP sealed_secrets_controller_build_info Build information.
# TYPE sealed_secrets_controller_build_info gauge
sealed_secrets_controller_build_info{revision="v0.15.0"} 0
# HELP sealed_secrets_controller_unseal_requests_total Total number of sealed secret unseal requests
# TYPE sealed_secrets_controller_unseal_requests_total counter
sealed_secrets_controller_unseal_requests_total 2
```
Especiall missing the "unseal_errors_total" metric.

rescDev:
Hi @alexanderdalloz , i had this issue myself (also with v0.15.0). Seems like the `unseal_errors_total` metric does not get exposed when there are no unseal errors (so `0` value does not get exposed). I tested this by purposely adding a secret with wrong values in a playground environment. The unsealing process failed and the metric has been exposed :)

I don't know if you're still missing other metrics though, this was just my experience with the `unseal_errors_total` metric :)

simon-anz:
I have same issue

simon-anz:
The bug came with https://github.com/bitnami-labs/sealed-secrets/pull/463/files
The code that initialised the vector was removed when "namespace" was added to the labels.

simon-anz:
https://www.robustperception.io/existential-issues-with-metrics

simon-anz:
Created PR https://github.com/bitnami-labs/sealed-secrets/pull/612

PR #612, opened 2021-07-20
Initialize unsealErrorsTotal with "fetch" = 0

simon-anz:
Add part of the initialization from https://github.com/bitnami-labs/sealed-secrets/pull/463
See https://github.com/bitnami-labs/sealed-secrets/issues/543

### DIFF
--- a/cmd/controller/metrics.go
+++ b/cmd/controller/metrics.go
@@ -89,7 +89,7 @@ func init() {
 	prometheus.MustRegister(httpRequestsTotal)
 	prometheus.MustRegister(httpRequestDurationSeconds)
 	// Initialise known label values
-	unsealErrorsTotal.WithLabelValues("fetch")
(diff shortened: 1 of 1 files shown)


simon-anz:
@kskewes @mkmik

kskewes:
LGTM per code incrementing error here: https://github.com/bitnami-labs/sealed-secrets/blob/62cfb264f53e9a378ef1b8df051ac4d8a2af5da0/cmd/controller/controller.go#L212

simon-anz:
What is needed to progress this PR?

alvneiayu:
hi @simon-anz 

Could you do a rebase to check that the test are passing, please?

Thanks a lot

Álvaro

alvneiayu:
Thanks a lot for the contribution but this solution, I think that it is incomplete. We are solving just this specific case but the problem is still present for the other cases.

Thanks again

PR #614, opened 2021-07-20
contrib/prometheus: Honor Labels added by controller

### LINKED ISSUE #613: Prometheus metrics `namespace` label collision
Since #463 we have the `namespace` label on metrics. :tada: 

This is great but it clashes with the default namespace label , eg: `namespace="sealed-secrets"` that Prometheus adds as part of service discovery.
This means that any dashboards and alerts will still have `sealed-secrets` as the namespace instead the namespace the secret was unable to be unsealed in, eg: `test`.

We can prefer the controller's metric namespace, eg: `test` by toggling the `honor_labels` flag to true in the `PodMonitor` per docs snippet: 
- https://prometheus.io/docs/prometheus/latest/configuration/configuration/#scrape_config
- https://github.com/prometheus-operator/prometheus-operator/blob/master/Documentation/api.md#podmetricsendpoint

```
# honor_labels controls how Prometheus handles conflicts between labels that are
# already present in scraped data and labels that Prometheus would attach
# server-side ("job" and "instance" labels, manually configured target
# labels, and labels generated by service discovery implementations).
#
# If honor_labels is set to "true", label conflicts are resolved by keeping label
# values from the scraped data and ignoring the conflicting server-side labels.
#
# If honor_labels is set to "false", label conflicts are resolved by renaming
# conflicting labels in the scraped data to "exported_<original-label>" (for
# example "exported_instance", "exported_job") and then attaching server-side
# labels.
<snip>
[ honor_labels: <boolean> | default = false ]
```

kskewes:
Closes #613 

Preserve `namespace="test"` labels added by the controller so that
dashboards and alerts will correctly show corresponding SealedSecrets
namespace and not the default, eg: `namespace="sealed-secrets"`

The helm chart already has this set.

### DIFF
--- a/controller-podmonitor.jsonnet
+++ b/controller-podmonitor.jsonnet
@@ -1,9 +1,9 @@
-// Prometheus Service Monitor manifest
+// Prometheus Pod Monitor manifest
 
 local controller = import 'controller.jsonnet';
 
 controller {
-  serviceMonitor: {
+  podMonitor: {
     apiVersion: 'monitoring.coreos.com/v1',
     kind: 'PodMonitor',
     metadata: {
@@ -27,6 +27,7 @@ controller {
       },
       podMetricsEndpoints: [
         {
+          honorLabels: true,  // prefer controller metric namespace
           port: 'http',
           interval: '30s',
         },
--- a/controller-podmonitor.jsonnet
+++ b/controller-podmonitor.jsonnet
@@ -1,4 +1,5 @@
 // Prometheus Pod Monitor manifest
+// ref: https://github.com/prometheus-operator/prometheus-operator#customresourcedefinitions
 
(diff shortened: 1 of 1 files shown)


juan131 (approved):
Thanks so much! LGTM

controller-podmonitor.jsonnet
```diff
@@ -1,9 +1,9 @@
-// Prometheus Service Monitor manifest
+// Prometheus Pod Monitor manifest
```
juan131, line 1: Add reference to PodMonitor def
```suggestion
// Prometheus Pod Monitor manifest
// ref: https://github.com/prometheus-operator/prometheus-operator#customresourcedefinitions
```

alvneiayu (approved):
LGTM

PR #617, opened 2021-08-02
Make both pod and container security context configurable

chgl:
This allows complete control over both the pod and container security context. The default container security context is now a bit more secure by explicitely setting `allowPrivilegeEscalation: false` and dropping all capabilites. The namings and usage of the two security contexts should now also be more aligned with other Bitnami charts.

The changes may not be backwards compatible, especially when using OpenShift, since setting `securityContext.runAsUser: ""` will no longer work, instead the entire pod security context will need to be disabled, I've updated the README accordingly.

### DIFF
--- a/helm/sealed-secrets/Chart.yaml
+++ b/helm/sealed-secrets/Chart.yaml
@@ -1,7 +1,7 @@
 name: sealed-secrets
 description: Helm chart for the sealed-secrets controller.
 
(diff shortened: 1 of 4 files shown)


juan131:
At #687 we included two parameters `podSecurityContext` and `containerSecurityContext` that allow customizing these settings.

Issue #602, opened 2021-06-18
Upgrade Role's namespace


haf:
```
customresourcedefinition.apiextensions.k8s.io/sealedsecrets.bitnami.com configured
serviceaccount/sealed-secrets-controller unchanged
Warning: rbac.authorization.k8s.io/v1beta1 Role is deprecated in v1.17+, unavailable in v1.22+; use rbac.authorization.k8s.io/v1 Role
role.rbac.authorization.k8s.io/sealed-secrets-key-admin unchanged
role.rbac.authorization.k8s.io/sealed-secrets-service-proxier unchanged
Warning: rbac.authorization.k8s.io/v1beta1 ClusterRole is deprecated in v1.17+, unavailable in v1.22+; use rbac.authorization.k8s.io/v1 ClusterRole
clusterrole.rbac.authorization.k8s.io/secrets-unsealer unchanged
Warning: rbac.authorization.k8s.io/v1beta1 RoleBinding is deprecated in v1.17+, unavailable in v1.22+; use rbac.authorization.k8s.io/v1 RoleBinding
rolebinding.rbac.authorization.k8s.io/sealed-secrets-controller unchanged
rolebinding.rbac.authorization.k8s.io/sealed-secrets-service-proxier unchanged
Warning: rbac.authorization.k8s.io/v1beta1 ClusterRoleBinding is deprecated in v1.17+, unavailable in v1.22+; use rbac.authorization.k8s.io/v1 ClusterRoleBinding
clusterrolebinding.rbac.authorization.k8s.io/sealed-secrets-controller unchanged
```

Like the warnings say

dewe:
- Migrate manifests and API clients to use the rbac.authorization.k8s.io/v1 API version, available since v1.8.
- All existing persisted objects are accessible via the new APIs
- **No notable changes**

https://kubernetes.io/docs/reference/using-api/deprecation-guide/#rbac-resources-v122

chriscardillo:
Thanks for raising this issue.

I am still a little confused. Does that mean sealed-secrets 0.16.0 will **not** work with k8s 1.22+ currently until the manifests are updated?

haf:
The manifests with the old versions will stop working yes. If they have already been applied and GKE performs the upgrade on its own it means the existing code will also stop working immediately.

xak2000:
AFAIK, GKE clusters will soon be updated to Kubernetes v1.22 in stable branch, so this issue is pretty important.

[...]

PR #627, opened 2021-08-26
feat: #602 Bump rbac api version

dosmanak:
rbac.authorization.k8s.io/v1beta1 Role is deprecated in v1.17+,
unavailable in v1.22+; use rbac.authorization.k8s.io/v1

done using command:
```
find -type f -exec sed -i -e 's%rbac.authorization.k8s.io/v1beta1%rbac.authorization.k8s.io/v1%g' {} \;
```


snuggie12:
Was this the correct way to fix this? I would have thought the jsonnet lock file should be updated since it's been 2 years since sourcing the library.

PR #630, opened 2021-08-28
Set default service name properly

### LINKED ISSUE #571: Chart misleading setting
Using kubeseal version 0.15.0 and chaer 0.15.0-r3

after installation following the docs to clean cluster get back the message:

```bash
$ kubeseal --fetch-cert
error: cannot fetch certificate: services "sealed-secrets" not found
```

The problem is that kubeseal looks by default for controller `sealed-secrets-controller`, while the chart generate the deployment with name `sealed-secrets` (and it was possible to access the cert with`-controller-name sealed-secrets` arg for kubeseal)

Adding the pluck to `values.yaml` 

```yaml 
sealed-secrets:
  fullnameOverride: "sealed-secrets-controller"
```

solved the problem

But for me it feels like it should work with defaults straight out of the box.
If you like this improvement will be happy to make a pr to fix it, otherwise just hope that this information will be useful for somebody

chevdor:
When installing the helm chart, the chances are high (it depends on a few variables so you *may* also end up lucky...) that the service will be named `sealed-secrets` whereas `kubeseal`expects `sealed-secrets-controller` by default. This PR fixes ther default to ensure it works directly. 

fix #571

### DIFF
--- a/helm/sealed-secrets/values.yaml
+++ b/helm/sealed-secrets/values.yaml
@@ -4,6 +4,8 @@ image:
   pullPolicy: IfNotPresent
   pullSecret: ""
 
+fullnameOverride: "sealed-secrets-controller"
+
 resources: {}
 nodeSelector: {}
 tolerations: []


chevdor:
The repo seems to be dead...

PR #631, opened 2021-08-30
Fix Chart controller name

### LINKED ISSUE #571: Chart misleading setting

kvokka:
Close #571

### DIFF
--- a/helm/sealed-secrets/Chart.yaml
+++ b/helm/sealed-secrets/Chart.yaml
@@ -1,7 +1,7 @@
 name: sealed-secrets
 description: Helm chart for the sealed-secrets controller.
 
-version: 1.16.1
+version: 1.16.2
 appVersion: v0.16.0
 
 kubeVersion: ">=1.16.0-0"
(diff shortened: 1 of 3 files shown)


juan131:
As I mentioned at https://github.com/bitnami-labs/sealed-secrets/issues/571, you can use the parameter `fullnameOverride` that you can use to overwrite the service name.

Issue #502, opened 2021-01-20
Bug in regards to "cannot fetch certificate: no endpoints available for service "http:sealed-secrets-controller:"


ghostsquad:
This is after reading #317 #397 and #368 all slightly related.

## Description

Error message:

```
cannot fetch certificate: no endpoints available for service "http:sealed-secrets-controller:
```

## Repro Steps

if my `sealed-secrets-controller` `service` is configured with:

```
spec:
  clusterIP: 172.20.54.13
  ports:
  - name: http
    port: 8080
    protocol: TCP
    targetPort: 8080
```

it does NOT work

but if you remove the port name `http` it works fine, i.e.

```
spec:
  clusterIP: 172.20.54.13
  ports:
  - port: 8080
    protocol: TCP
    targetPort: 8080
```

### Testing

shell 1
```
kubectl proxy
```

shell2
```
curl http://127.0.0.1:8001/api/v1/namespaces/kube-system/services/http:sealed-secrets-controller:/proxy/v1/cert.pem
```

## Other Related Information

I use the bitnami kube.libsonnet libraries to manage my deployments & services, only using upstream yamls as reference (not truth), and `Service()` function does this automatically!

```
Service(name): $._Object("v1", "Service", name) {
    local service = self,

    target_pod:: error "service target_pod required",
    port:: self.target_pod.spec.containers[0].ports[0].containerPort,

    // Helpers that format host:port in various ways
    host:: "%s.%s.svc" % [self.metadata.name, self.metadata.namespace],
    host_colon_port:: "%s:%s" % [self.host, self.spec.ports[0].port],
    http_url:: "http://%s/" % self.host_colon_port,
    proxy_urlpath:: "/api/v1/proxy/namespaces/%s/services/%s/" % [
      self.metadata.namespace,
      self.metadata.name,
    ],
    // Useful in Ingress rules
    name_port:: {
      serviceName: service.metadata.name,
      servicePort: service.spec.ports[0].port,
    },

    spec: {
      selector: service.target_pod.metadata.labels,
      ports: [
        {
          port: service.port,
          name: service.target_pod.spec.containers[0].ports[0].name,
          targetPort: service.target_pod.spec.containers[0].ports[0].containerPort,
        },
      ],
      type: "ClusterIP",
    },
  },
```
  
https://github.com/bitnami-labs/kube-libsonnet/blob/master/kube.libsonnet#L181
  
---
  
The curl call can be resolved by appending the port name:
```
[...]

PR #648, opened 2021-10-06
Fix Cannot fetch certificate: no endpoints available for service "http:sealed-secrets-controller:"

sathieu:
See https://istio.io/latest/docs/ops/configuration/traffic-management/protocol-selection/#explicit-protocol-selection

- fixes https://github.com/bitnami-labs/sealed-secrets/issues/502
- fixes https://github.com/bitnami-labs/sealed-secrets/issues/694


sathieu:
@juan131 Please review :pray:.

juan131:
hanks so much for your contribution @sathieu!!

We're performing a major refactorization of the chart at https://github.com/bitnami-labs/sealed-secrets/pull/687
Let's wait until this is done before continue with your changes

juan131:
@sathieu the new major version was just released! 
```
$ helm repo update sealed-secrets
Hang tight while we grab the latest from your chart repositories...
...Successfully got an update from the "sealed-secrets" chart repository
Update Complete. ⎈Happy Helming!⎈
$  helm search repo sealed-secrets/sealed-secrets
NAME                         	CHART VERSION	APP VERSION	DESCRIPTION
sealed-secrets/sealed-secrets	2.0.0        	v0.17.1    	Helm chart for the sealed-secrets controller.
```
Please feel to rebase this branch in your fork with the latest changes in the main branch, and update this PR.

cmd/kubeseal/main.go
```diff
@@ -51,24 +51,25 @@ const (
...
-	sealingScope   ssv1alpha1.SealingScope
-	reEncrypt      bool // re-encrypt command
-	unseal         = flag.Bool("recovery-unseal", false, "Decrypt a sealed secrets file obtained from stdin, using the private key passed with --recovery-private-key. Intended to be used in disaster recovery mode.")
-	privKeys       = flag.StringSlice("recovery-private-key", nil, "Private key filename used by the --recovery-unseal command. Multiple files accepted either via comma separated list or by repetition of the flag. Either PEM encoded private keys or a backup of a json/yaml encoded k8s sealed-secret controller secret (and v1.List) are accepted. ")
+	certURL            = flag.String("cert", "", "Certificate / public key file/URL to use for encryption. Overrides --controller-*")
+	controllerNs       = flag.String("controller-namespace", metav1.NamespaceSystem, "Namespace of sealed-secrets controller.")
+	controllerName     = flag.String("controller-name", "sealed-secrets-controller", "Name of sealed-secrets controller.")
+	controllerPortName = flag.String("controller-port-name", "", "Port name of sealed-secrets controller.")
```
juan131, line 57: It seems this parameter isn't used in the `main()` function when invoking the `run` function... Also.. It seems we're already assuming that the SealedSecret service has a port with name "http" in functions such as `validateSealedSecret` or `reEncryptSealedSecret`, see:

- https://github.com/bitnami-labs/sealed-secrets/blob/main/cmd/kubeseal/main.go#L316
- https://github.com/bitnami-labs/sealed-secrets/blob/main/cmd/kubeseal/main.go#L350

Couldn't do a similar assumption in `openCertCluster`?

juan131 (changes requested):
Thanks so much for the PR @sathieu !! Please check my comments

helm/sealed-secrets/templates/deployment.yaml
```diff
@@ -45,6 +45,7 @@ spec:
           args:
             - --key-prefix
             - {{ .Values.secretName | quote }}
+            - --controller-port-name=http
```
juan131, line 48: Please bump the patch version in the [Chart.yaml](https://github.com/bitnami-labs/sealed-secrets/blob/main/helm/sealed-secrets/Chart.yaml)

juan131 (changes requested):

helm/sealed-secrets/templates/deployment.yaml
juan131: Follow up question: you're implementing this flag for `kubeseal` but the `controller` ignores it. Am I right?

sathieu:
@juan131 Comments addressed. Please review again.

andoshin11:
This is indeed critical one. Appreciate the fix.

sathieu:
@juan131 Anything missing in this PR?

juan131 (approved):
LGTM! I can confirm this fix works

sathieu:
@juan131 Maybe this fix deserve a major version bump?

Also the CI is failing, is this related to this change?

juan131 (changes requested):
@sathieu I just realized that this fixes the chart.. but it breaks the compatibility with the `controller.yaml` manifests that we publish at:

- https://github.com/bitnami-labs/sealed-secrets/releases/tag/v0.17.1

These manifests don't have a "http" name in the service port.

> maybe this fix deserve a major version bump?

Why do you think so? I don't see that they break backwards compatibility.

juan131:
@sathieu I updated your branch with a suggestion that dynamically obtains the svc port name. Please let me know what you think

alvneiayu (approved):
Awesome work, LGTM

sathieu:
@juan131 Thanks.  This is great! lgtm!

linkvt:
Hello @juan131 , I'm addressing this here because I think your change broke the sealed-secrets-controller for k8s users with no permissions in kube-system - no blame intended.

As you can see in this line the proxier role only grants permissions to get `services/proxy` and not `services` https://github.com/bitnami-labs/sealed-secrets/blob/1490884d3b0d64c27900e63abfc0b268d49eaab2/helm/sealed-secrets/templates/role.yaml#L43-L45

I guess this is because `Services(namespace).ProxyGet` was replaced with `Services(namespace).Get`. Was there a reason for this? Otherwise I could create a PR to add `- services` to the Roles resources field to make it work again.

Thanks for checking.

juan131:
Hi @linkvt 

You're totally right!! That should be pretty straight forward to fix since it just requires to update the RBAC permissions. I'll send a PR to fix it.

> Was there a reason for this?

We still use a proxy to access the service (I didn't change that), but previously we access the services to obtain the service port name information.

chrisob:
@juan131 Coming off of @linkvt 's comment, is there no alternative that would support using the original `Services(namespace).ProxyGet`? Reason being, just simply allowing `get services` in RBAC as is the case in #715 punches a bigger hole in RBAC which may not be desired in multitenant environments (especially considering *any* authenticated user can get the entire service's manifest!).

Maybe I'm a little over cautious, but I feel that only allowing `services/proxy` RBAC was a nice solution which prevented random users from discovering more information about sealed-secrets-controller than they need, a-la the principle of least privilege.

chrisob:
Just brainstorming here, but one approach could be to control this "discover service endpoint" functionality via feature flag, but by default have kubeseal assume the default `http` port & not attempt any service lookup.

juan131:
Hi @chrisob 

Thanks for sharing your concerns, we really appreciate your opinion.  IMHO, the security risk is not that high. 

It limits the scope to the namespace where the SealedSecrets controller is installed, offering the opportunity to use a dedicated namespace for the controller so authenticated users can exclusively obtain the controller service metadata. Furthermore, this service is by default a "ClusterIP" service that exposes a single HTTP port, I don't think we add much value in terms of security by hiding this info from authenticated users.

mkmik:
My plan for fixing this was to create a ConfigMap with the public key 
and avoid all this mess with actively trying to talking to the controller.

PR #668, opened 2021-11-25
Move release automation to GH actions

### LINKED ISSUE #661: Release v0.17.0 automation is broken
The normal day to day PR CI has been migrated to GitHub Actions long time ago.
However, the release automation is still on travis (because it needed the tokens for quay and for gh I can no longer generate).

Today I cut release v0.17.0 but the travis automation apparently failed. 
I no longer have access to travis (now that it has migrated to travis-ci.com) so I don't know why it fails.

The GH CI integration tests are also broken (we cannot reproduce it locally); so it could be that the release CI also fails in the same way, or perhaps it fails because of a different reason.

CC: @beltran-rubo

juan131:

**Description of the change**

This PR is a proposal to move away from Travis to perform the release process and start using GH actions instead. Also, this PR proposes using [GoReleaser](https://goreleaser.com) to build/publish `controller` and `kubeseal` binaries using a multi-os & multi-arch approach.

**Benefits**

- Standardization: it doesn't make sense to use different CI/CD tools depending on what we want to do. Using GH actions both for testing contributions and publishing simplifies our processes.
- Reduce complexity: by relying on GoReleases we don't need to manage building multi-arch binaries & images in the Makefile, simplifying our automations.

**Possible drawbacks**

None

**Applicable issues**

  - fixes #661

**Additional information**

- Learn more about GoReleaser at https://goreleaser.com/how-it-works/
- This proposal in inspired in [this blog post](https://namiops.medium.com/golang-multi-arch-docker-image-with-github-action-b59a62c8d2bd).


mkmik:
Thanks. I'll take a look.

My original plan was to build the multiarch image with `ko` which would cross compile them for many archs, but as you know after January I couldn't access the admin pages and change the secrets since January so the CI migration got stuck with Travis. 

I don't think it's particularly interesting to have the conductor binaries as GH artifacts. We should have kubeseal binaries and the yaml manifests pointing to multiarch images

juan131:
Hi @mkmik 

> My original plan was to build the multiarch image with ko which would cross compile them for many archs, but as you know after January I couldn't access the admin pages and change the secrets since January so the CI migration got stuck with Travis.

Yes I know it wasn't possible to solve this after you lost the Quay credentials. I chose GoReleaser because it can manage both building the multiarch/multios binaries + handling the GH releases, but any feedback is welcomed if you think there are better alternatives.

> I don't think it's particularly interesting to have the conductor binaries as GH artifacts. We should have kubeseal binaries and the yaml manifests pointing to multiarch images

🤔  It makes sense.. That said, it's something we can configure in GoReleaser if we remove "controller" form the builds to package, see:

- https://goreleaser.com/customization/archive/

mkmik:
cool I didn't know goreleaser could also build docker images; I took a quick look here at https://goreleaser.com/customization/docker/ but I don't see where you enable that feature in your PR

juan131:
Yes @mkmik, but the feature to create Docker Manifests seems to be in beta, see https://goreleaser.com/customization/docker_manifest
That's why I decided to rely on [Docker Build & Push GH action](https://github.com/docker/build-push-action] to build/push the images.

mkmik:
Yes but the dockerfile doesn't build the controller, the binary is built by the Makefile and then copied in the image:

```
FROM gcr.io/distroless/static@sha256:c6d5981545ce1406d33e61434c61e9452dad93ecd8397c41e89036ef977a88f4
MAINTAINER Marko Mikulicic <mkmik@vmware.com>
COPY controller /usr/local/bin/

EXPOSE 8080
ENTRYPOINT ["controller"]
```


I think we have three options:

1) Keep the existing makefile and adapt it so it can run from GitHub Actions
2) use goreleaser for CLI and creating the release but using `ko` for actually building the image
3) using this new beta feature of goreleaser 

I think goreleaser and ko are orthogonal

alvneiayu (approved):

juan131:
@mkmik I'd rather go with option 2: `use goreleaser for CLI and creating the release but using ko for actually building the image`. The reasons: I'd like to avoid using beta features (option 3) and using custom logic to build the Docker manifests in the Makefile (option 1).

That said, do you have any strong opinion to use `ko`? I guess we can use the GH action below to setup `ko` and use it to build/publish the image:

- https://github.com/imjasonh/setup-ko 

But... I don't know what are the advantages of using it vs using `docker/build-push-action` GH action as I suggested:

```
      # Build & Publish multi-arch image
      - name: Login to Quay
        uses: docker/login-action@v1
        with:
          registry: quay.io
          username: ${{ secrets.QUAY_USER }}
          password: ${{ secrets.QUAY_PASS }}
      - name: Build and push
        id: docker_build
        uses: docker/build-push-action@v2
        with:
          push: true
          platforms: linux/amd64,linux/arm64
          tags: ${{ env.image_name }}:latest
```

mkmik:
the advantage of using ko is that it doesn't setup qemu emulation in the container and just builds the go binary using native cross compiling functionality. It's a very nice tool that produces multi-arch images off the shelf.

That said, honestly I don't care much about the exact mechanics as long as it works and we unblock our ability to make releases.

juan131:
> That said, honestly I don't care much about the exact mechanics as long as it works and we unblock our ability to make releases.

100% agreed! let's unblock this and I'll create an issue to explore `setup-ko` GH action right after.

PR #672, opened 2021-12-03
 Release v0.17.0 automation is broken (#661)

### LINKED ISSUE #661: Release v0.17.0 automation is broken

alvneiayu:
**Description of the change**

**Changes done to fix the integration tests:**
First round of fixes to solve the actual 27 errors:
- Replace the controller image to use quay official release instead of the gcr image removed
- Bump the Minikube version to reduce the errors in the CI GH Action

After update Minikube, I reproduced another error:
```
error: converting  to : type names don't match (Unknown, RawExtension), and no conversion 'func (runtime.Unknown, runtime.RawExtension) error' registered.
```
To fix it, the solution was:
- Bump client-go to 0.17.12. This is the first stable version that we can use to fix the integration tests without generate incompatibility issues.

**Benefits**

Integration tests are running in green
Screenshot from 2021-12-03 18-40-27

**Additional information**

Future work: Update client-go library to a more current version (right now the last stable version is 0.22.4)


agarcia-oss (approved):
LGTM!

go.mod
```diff
@@ -5,24 +5,25 @@ go 1.12
...
 	github.com/throttled/throttled v2.2.2+incompatible
 	golang.org/x/crypto v0.0.0-20200622213623-75b288015ac9
 	golang.org/x/text v0.3.3 // indirect
 	gopkg.in/yaml.v2 v2.3.0
-	k8s.io/api v0.16.8
-	k8s.io/apimachinery v0.16.8
-	k8s.io/client-go v0.16.8
+	k8s.io/api v0.17.11
+	k8s.io/apimachinery v0.17.11
+	k8s.io/client-go v0.17.11
```
juan131, line 26: This is just used for the integration tests, right? I mean, this isn't used by `kubeseal` nor the controller. Am I right? I  just want to double-check that we're not including changes that can affect backwards compatibility

juan131 (commented):
Thanks so much for this @alvneiayu !! Please check my comments, I just want to double-confirm we're not breaking anything.

go.mod
alvneiayu: hi @juan131 
client-go is used in:
- Controller
- Kubeseal
- Integration tests
The changes are working right now with 1.16.13, 1.17.11 and 1.20.4. So, in my opinion, I think that we are not breaking the backwards compatibility.

juan131 (approved):
LGTM

mkmik (approved):

PR #677, opened 2021-12-14
Use go modules without vendoring

mkmik:
Go modules are mature now and we don't need vendoring anymore


alvneiayu:
hi @mkmik 

I think that it is a really awesome idea. The compilation is failing. I think that we need to do some changes to compile without the -mod vendor

Thanks

Álvaro

mkmik:
I grepped quickly in .github but I didn't find mention of the mod vendor option; probably I grepped badly

alvneiayu:
I think that the problem is in the Makefile. We have a GO_FLAG adding -mod=vendor and I think that we need to remove it.

mkmik:
Fixed

alvneiayu (approved):
LGTM

PR #689, opened 2021-12-16
Remove unrequired files

juan131:

**Description of the change**

It seems these two files are exclusively used to ensure that the version available in the **version.yaml** is synced with the one available in the chart's **Chart.yaml** (`appVersion` field) and chart's **values.yaml** (`image.tag` field). However, I don't see much value on this since we're not using them in the CI/CD tests.

**Benefits**

Remove tech debt.

**Possible drawbacks**

None

**Applicable issues**

N/A

**Additional information**

N/A


alvneiayu (approved):
LGTM

agarcia-oss (approved):

mkmik (changes requested):
we should be using them. There used to be a GH action that ran this linter.

Where did it go?

mkmik:
added in https://github.com/bitnami-labs/sealed-secrets/pull/497

mkmik (approved):
uh so now we no longer sync the versions of helm charts and sealed-secrets?

well, that's a pity, because lots and lots of people were utterly confused when the chart version and the app version were not in sync.

But I don't care much honestly, as long as I'm not handling the support requests.

juan131:
> well, that's a pity, because lots and lots of people were utterly confused when the chart version and the app version were not in sync.
> But I don't care much honestly, as long as I'm not handling the support requests.

I'd be glad to explain the differences if this confuses users... But I don't see any value on keeping these versions synced. I mean, the chart and app versions are completely orthogonal.

mkmik:
> mean, the chart and app versions are completely orthogonal.

I know, but I noticed that most end users often don't grasp the difference, especially since the helm chart is produced by the same entity.

PR #704, opened 2022-01-08
Update README.md

LarsBingBong:
With correct label to be used when deleting the sealed-secrets pod. The current does not find/match the sealed-secrets Pod

PR template not strictly followed as this PR "only" entails to update the base ReadMe docs with a very simple change.

Thanks


mkmik (approved):

davidkarlsen (approved):

juan131 (approved):
Thanks for the fix! LGTM

PR #705, opened 2022-01-10
Bump k8s.io/client-go to v0.22.1

lrascao:
**Description of the change**

Bump k8s client api to v0.22.1, single change is propagating context across all functions in order to
reach k8s client api.

**Benefits**

Allows k8s controllers to use sealed secrets client API when running on newer more recent versions.

**Possible drawbacks**

Do not foresee any, single change is context propagation.


mkmik:
we're currently running the tests against `k8s: ["1.16.13", "1.17.11", "1.20.4"]`

Probably we should also add version `1.22` to the build matrix.

lrascao:
> we're currently running the tests against `k8s: ["1.16.13", "1.17.11", "1.20.4"]`
> 
> Probably we should also add version `1.22` to the build matrix.

roger, i'll add it here

mkmik:
@lrascao you probably need to bump also the Go version in `go.mod` to 1.16

> ../../../go/pkg/mod/k8s.io/client-go@v0.22.1/plugin/pkg/client/auth/exec/metrics.go:21:2: package io/fs is not in GOROOT (/opt/hostedtoolcache/go/1.15.15/x64/src/io/fs)
make: *** [Makefile:61: controller-static-linux-amd64] Error 1

mkmik:
I think we can drop support for go <1.16

We can build go 1.16 and 1.17. The go 1.18 release is behind the corner so very soon this will cover the last three versions anyway.

Yeah, let's drop Go 1.14 and 1.15

mkmik:
```
  container:
    name: Build Container
    runs-on: ubuntu-latest
    steps:
...
```

should we add a `actions/setup-go@v2` step there too in order to make sure we get a recent version of go? (I'd use the latest 1.17 there)

mkmik:
ok, the build is good now; now the problem appears to lie in the fact that the ingegration tests are behind the `// +build integration` tag; you have to update the API there too

lrascao:
> ok, the build is good now; now the problem appears to lie in the fact that the ingegration tests are behind the `// +build integration` tag; you have to update the API there too

yeah, on it now

mkmik (approved):
LGTM;

@juan131 this PR also changes the CI job names and requires updating the list of required CI jobs (e.g. remove `Build (1.14, macos-latest)`  and `Build (1.15, macos-latest)` and add the 1.6 and 1.7 ones)

juan131:
Awesome, thanks so much for this PR @lrascao 

> I think we can drop support for go <1.16
>  We can build go 1.16 and 1.17. The go 1.18 release is behind the corner so very soon this will cover the last three versions anyway.
>  Yeah, let's drop Go 1.14 and 1.15

I totally agree, it's time to drop it.

> Probably we should also add version 1.22 to the build matrix.

+1

.github/workflows/ci.yml
```diff
@@ -45,6 +45,12 @@ jobs:
         curl -sLf https://github.com/bitnami/kubecfg/releases/download/v0.16.0/kubecfg-linux-amd64 >~/bin/kubecfg
         chmod +x ~/bin/kubecfg
 
+    - name: Set up Go 1.x
+      uses: actions/setup-go@v2
+      with:
+        go-version: ^1.17
+      id: go
```
juan131, line 52: Why is this required? I don't thinks it's required for building the container.

.github/workflows/ci.yml
```diff
@@ -71,7 +77,7 @@ jobs:
     needs: container
     strategy:
       matrix:
-        k8s: ["1.16.13", "1.17.11", "1.20.4"]
+        k8s: ["1.16.13", "1.17.11", "1.20.4", "1.22.5"]
```
juan131, line 80: Awesome thanks!

.github/workflows/ci.yml
mkmik: Yes. We added that as a result of a build failure.

It appears that without this GH used a go compiler older than 1.16.

We choose to pin 1.17 because it's better to build the docker image with the latest version as opposed to the minimal supported version.

juan131:
@lrascao the Go version needs to be updated in the [release GH action](https://github.com/bitnami-labs/sealed-secrets/blob/main/.github/workflows/release.yaml#L21) too

.github/workflows/ci.yml
lrascao: yeah, there's no reference to a default version in [actions/setup-go](https://github.com/actions/setup-go#v2) so i'm guessing it's defaulting to whatever's installed in ubuntu 20.04.3

juan131: thanks for the context

juan131 (approved):
LGTM

juan131:
@mkmik @lrascao I plan to create a new release (`0.17.2`) of the binaries including these changes + the changes we're working on at https://github.com/bitnami-labs/sealed-secrets/pull/648 once they're merged

PR #737, opened 2022-02-05
Fix image pull secrets in chart

juan131:

**Description of the change**

The `image.pullSecrets` parameter is an array but we're wrongly treating it as a string. This PR address this.

**Benefits**

Users can set a custom a list of image pull secrets.

**Possible drawbacks**

None

**Applicable issues**

N/A

**Additional information**

N/A


agarcia-oss (approved):
Good catch, @juan131 👍

PR #738, opened 2022-02-09
Improve raw mode docs

wmcdona89:
**Description of the change**

Improves the raw mode docs to ensure users are aware that use of the namespace-wide or cluster-wide scope requires the SealedSecret, in which the encrypted value is placed, contain an annotation corresponding to the scope.

**Benefits**

**Possible drawbacks**

**Applicable issues**

- fixes #

**Additional information**

### DIFF
--- a/README.md
+++ b/README.md
@@ -387,20 +387,43 @@ echo -n baz | kubectl create secret generic mysecret --dry-run=client --from-fil
 
 Creating temporary Secret with the `kubectl` command, only to throw it away once piped to `kubeseal` can
 be a quite unfriendly user experience. We're working on an overhaul of the the CLI experience. In the meantime,
-we offer an alternative mode where kubeseal only cares about encrypting a value to stdout and it's your responsibility to put it inside a SealedSecret resource (not unlike any of the other k8s resources).
+we offer an alternative mode where kubeseal only cares about encrypting a value to stdout and it's your responsibility to put it inside a `SealedSecret` resource (not unlike any of the other k8s resources).
 
 It can also be useful as a building block for editor/IDE integrations.
 
 The downside is that you have to be careful to be consistent with the sealing scope, the namespace and the name.
-See [Scopes](#scopes):
+
+See [Scopes](#scopes)
+
+#### scrict scope (default)
 
 ```console
 $ echo -n foo | kubeseal --raw --from-file=/dev/stdin --namespace bar --name mysecret
 AgBChHUWLMx...
+```
+#### namespace-wide scope
+
+```console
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -387,21 +387,46 @@ echo -n baz | kubectl create secret generic mysecret --dry-run=client --from-fil
...
+#### strict scope (default)
 
 ```console
 $ echo -n foo | kubeseal --raw --from-file=/dev/stdin --namespace bar --name mysecret
 AgBChHUWLMx...
+```
+
+#### namespace-wide scope
+
+```console
```
agarcia-oss, line 407:
```suggestion
```bash
```
I think it'll be better to use the `bash` code decorator to be consistent with the rest of the document style.

README.md
```diff
@@ -387,21 +387,46 @@ echo -n baz | kubectl create secret generic mysecret --dry-run=client --from-fil
...
+Include the `sealedsecrets.bitnami.com/namespace-wide` annotation in the `SealedSecret`
+```yaml
+metadata:
+  annotations:
+    sealedsecrets.bitnami.com/namespace-wide: "true"
+```
+
+#### cluster-wide scope
+
+```console
```
agarcia-oss, line 420:
```suggestion
```bash
```
See comment above.

juan131: I don't agree @agarcia-oss. The rule we've been following is:
- Use "bash" when they're instructions intended for copy-pasting (to avoid the `$`)
- Use "console" when they're examples including the command's output.
In this case, we should use "console"

README.md
juan131: check my comment below

README.md
```diff
@@ -387,21 +387,46 @@ echo -n baz | kubectl create secret generic mysecret --dry-run=client --from-fil
...
 
 It can also be useful as a building block for editor/IDE integrations.
 
 The downside is that you have to be careful to be consistent with the sealing scope, the namespace and the name.
-See [Scopes](#scopes):
 
-```console
+See [Scopes](#scopes)
+
+#### strict scope (default)
```
juan131, line 398: Could you please run [doctoc](https://github.com/thlorenz/doctoc) to regenerate the TOC? You're including new headers that should be taken into account.
Thanks in advance

juan131 (changes requested):

README.md
wmcdona89: Thanks. I hadn't considered the TOC. This would add a 3rd level to the TOC and these don't really need to be headings. Updated to use formatting more similar to the `Scopes` section.

README.md
wmcdona89: Used "console"

juan131 (approved):
LGTM

alvneiayu (approved):
LGTM

PR #742, opened 2022-02-10
Add integration with VIB

juan131:
Signed-off-by: juan131 <juanariza@vmware.com>

**Description of the change**

This PR replaces the current GH action used to lint/validate the chart (based on [chart-testing](https://github.com/helm/chart-testing-action)) with VIB so we can:

- Test the chart using different platforms.
- Add tests (using [goss](https://github.com/aelsabbahy/goss)) to validate the solution.

**Benefits**

Using a more powerful tool to validate the Helm chart.

**Possible drawbacks**

None

**Applicable issues**

- fixes #741

**Additional information**

We need to update the "Actions secrets" of the project to include the `CSP_API_TOKEN` env. variable.

PR #743, opened 2022-02-11
Trivial PR to test VIB integration

juan131:
Test to ensure the validation we added at https://github.com/bitnami-labs/sealed-secrets/pull/742 works.


juan131:
I switched the validation method to use "health-check" instead of "goss" since the latter doesn't support distroless containers.

agarcia-oss (approved):
All good! thanks for the PR @juan131 👍

alvneiayu (approved):

PR #715, opened 2022-01-20
Add extra RBAC permissions

juan131:
Signed-off-by: juan131 <juanariza@vmware.com>

**Description of the change**

This PR adds extra permissions to the ClusterRole so it can list services and the panic error below is addressed:

```
panic: secrets is forbidden: User "system:serviceaccount:kube-system:sealed-secrets" cannot list resource "secrets" in API group "" in the namespace "kube-system"
```

It also adds extra permissions to the "service-proxier" Role since the changes introduced at https://github.com/bitnami-labs/sealed-secrets/pull/648 require getting the service to obtain the port name information dynamically.

**Benefits**

Sealed Secrets to be compatible with K8s environments with RBAC enabled.

**Possible drawbacks**

None

**Applicable issues**

- fixes #708

**Additional information**

N/A

PR #745, opened 2022-02-16
Added required resourceName to Role

jmederer:
**Description of the change**
Added necessary resourceName in Role when "RBAC create" is used.
Without this Change i got this error from kubeseal (client)
```
error: cannot fetch certificate: services "http:sealed-secrets-controller:http" is forbidden: User "XXXXXXX" cannot get resource "services/proxy" in API group "" in the namespace "sealed-secrets"
```
Eventually caused by 
- fixes #648

### DIFF
--- a/helm/sealed-secrets/Chart.yaml
+++ b/helm/sealed-secrets/Chart.yaml
@@ -16,4 +16,4 @@ maintainers:
     email: mmikulicic@gmail.com
 name: sealed-secrets
 type: application
-version: 2.1.2
+version: 2.1.3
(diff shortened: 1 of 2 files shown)


juan131:
Hi @jmederer, thanks for your PR!

I might be wrong but I don't think the issue is related to the resource name indicated in the Role... We did experience an issue related to RBAC after adding kubeseal the ability to auto-discover the svc portname but it was addressed at https://github.com/bitnami-labs/sealed-secrets/pull/715  

What version of `kubeseal` and the Sealed Secrets helm chart are you using?

jmederer:
Hi @juan131 

I'm using
Chart Version 2.1.2  and kubeseal 0.17.3

juan131 (approved):
I just confirmed you're 100% right, we need to include the port name in the RBAC rule.

alvneiayu (approved):
LGTM

PR #747, opened 2022-02-17
POC: kseal

### LINKED ISSUE #717: Kubeseal to use cobra library
**Which component**:

- kubeseal

The current CLI, kubeseal, is quite complex since every action this CLI can perform depends on a series of flags. In other words, there are no subcommands to help simplifying the CLI usage for certain use cases.

You can simply run the command below to realize there are **47 supported flags**!! 

```console
$ kubeseal --help
Usage of kubeseal:
(...)
```

This result in a poor UX that makes it quite difficult to new potential users to adopt the project.

Use [cobra library](https://github.com/spf13/cobra) to split the current supported use cases on several subcommands. For instance:

- `kubeseal fetch` in replacement of `--fetch-cert` flag.
- `kubeseal merge` in replacement of `--merge-into` flag.
- `kubeseal re-encrypt` in replacement of `--re-encrypt` flag.
- `kubeseal recovery` in replacement of `--recovery-unseal` flag.
- `kubeseal raw` in replacement of `--raw` flag.
- `kubeseal validate` in replacement of `--validate` flag.

Cobra also makes it very simple to create/manage a local config file for kubeseal (`~/.kubesea/config`) and create a `kubeseal config` subcommand to allow users to configure locally settings such as the controller name and the controller namespace. This can also help to reduce the number of required flags to use the CLI.

N/A

**Additional context**

Adopting the cobra interface is an important code refactoring that will require adapting all the docs and it'll require bumping the new major version. That said, we should prioritize not introducing incompatibilities in the controller side with kubeseal `0.x` series.

juan131:
**Description of the change**

At https://github.com/bitnami-labs/sealed-secrets/issues/717, I proposed using a completely new approach for the CLI to improve the UX: adopting the Cobra framework. This is just a very basic (and incomplete) POC so you have a better idea of my proposal.

I proposed you to fetch this PR changes and build the CLI in your own environment:

```bash
git fetch upstream "pull/747/head:pr-747"
git checkout "pr-747"
cd cmd/kseal
go build
```

Then, play with it and provide feedback:

```console
$ ./kseal --help
kseal is a CLI that uses asymmetric crypto to encrypt secrets that only the Sealed Secrets controller can decrypt.

Usage:
  kseal [flags]
  kseal [command]

Available Commands:
  config      Configure kseal
  create      Create a Sealed Secret
  help        Help about any command
  pubkey      Manage the controller public keys
  unseal      Unseal a Sealed Secret
  verify      Verify a Sealed Secret

Flags:
      --config string   config file (default is $HOME/.kseal/config)
  -h, --help            help for kseal

Use "kseal [command] --help" for more information about a command.
```

**Benefits**

- Simpler CLI
- Better UX

**Possible drawbacks**

None

**Applicable issues**

- fixes #717

**Additional information**

This new CLI doesn't attempt to replace `kubeseal` immediately. The idea is maintaining both `kseal` and `kubeseal` during a transition period.


cmd/kseal/cmd/fetch/fetch.go
```diff
@@ -0,0 +1,47 @@
...
+// NewCmdFetch creates a command object for the "fetch" action.
+func NewCmdFetch() *cobra.Command {
+	fetchCmd := &cobra.Command{
+		Use:   "fetch",
+		Short: "Fetch public key from the controller",
+		Long: `Fetch the public key to use to encrypt secrets from the Sealed Secrets controller
+
+Examples:
+
+    kseal fetch                  Fetch public key and write its content to stdout.
```
mkmik, line 35: I think calling this command "fetch" may be confusing.

Perhaps 

`kseal pubkey fetch` or `kseal pubkey get`?

the "pubkey" subcommand is a natural home for adding multikey support e.g. `kseal pubkey list` etc

cmd/kseal/cmd/encrypt/encrypt.go
```diff
@@ -0,0 +1,61 @@
...
+)
+
+// NewCmdEncrypt creates a command object for the "encrypt" action.
+func NewCmdEncrypt() *cobra.Command {
+	encryptCmd := &cobra.Command{
+		Use:   "encrypt",
+		Short: "Encrypt a secret",
+		Long: `Encrypt a secret using a public key.
+
+Examples:
```
mkmik, line 33: I think the main UX issue in `kubeseal` is the fact that we relied too much on people "feeding" in a k8s Secret as input to Seal.

I still think that's a very useful feature, but the new command should offer a more intuitive structure for creating and updating SealedSecret files.

One of the common pitfalls in `kubeseal` was that it was expecting data on `stdin` by default. That behavior fits very well in the unix philosophy but unfortunately a lot of users are not proficient in the unix way. 

Another source of confusion was the distinction between encrypting values and producing SealedSecret files (which contain encrypted items). Some people care about one aspect, some care about the other (hence the "--raw" API).

Proposal:

1. add a `create` command that mirrors closely the behaviour of `kubectl create secret`:

```
$ kseal create foo -n myns --from-literal=item1=value
$ kseal create foo -n myns --from-file=item1=file.txt --from-literal=item2=value
$ kseal create foo -n myns --from-env-file=.env
$ kseal create foo -n myns --from-literal=item1=value
$ kseal create foo -n myns --from-literal=item1=value --type=my-secret-type
$ kseal create foo --scope cluster-wide --from-literal=item1=value
```

2. Add the ability to create a sealed secrets from an existing k8s secret:

```
$ kseal create --from-secret=existing-secret.yaml
```

3. Consistent command for updating existing secrets (aka `--merge-into`):

```
$ kseal update sealed-secret.yaml --from-literal=item1=value
```

4. Ability to produce individual encrypted data items (tentative name scheme)

```
$ kseal data encrypt --sealed-secrets-context=sealed-secrets.yaml --from-file=foo.txt 
$ kseal data encrypt --sealed-secrets-context=sealed-secrets.yaml --from-literal=bar
$ kseal data encrypt --name foo -n myns --scope namespace-wide --from-literal=baz
```

cmd/kseal/cmd/fetch/fetch.go
juan131: Awesome feedback! Let me implement this.

cmd/kseal/cmd/encrypt/encrypt.go
juan131: Brilliant! I guess this would be also a replacement for the `--raw` mode, right?

mkmik: yes

cmd/kseal/cmd/fetch/fetch.go
mkmik: thanks! resolving this thread

alemorcuq:
Thanks for this PR. We are closing this as stale.

PR #749, opened 2022-02-17
Allow setting of the clusterrolename for multiple instances of sealed…

derjohn:
I use two instances in two different namespaces and RBAC creation fails for the instance in the second namespace. I use different service accounts, one for each namespace.
Overriding the RBAC cluster role name solves the issue.

### DIFF
--- a/helm/sealed-secrets/values.yaml
+++ b/helm/sealed-secrets/values.yaml
@@ -259,6 +259,9 @@ rbac:
   ## @param rbac.pspEnabled PodSecurityPolicy
   ##
   pspEnabled: false
+  ## @param rbac.clusterRoleName Name of the clusterrole. To run multiple instances of sealed secrets with create RBAC true you need to set differnet rolenames.
+  ##
+  clusterRoleName: secrets-unsealer
 
 ## @section Metrics parameters
 
--- a/helm/sealed-secrets/templates/cluster-role-binding.yaml
+++ b/helm/sealed-secrets/templates/cluster-role-binding.yaml
@@ -10,7 +10,7 @@ metadata:
 roleRef:
   apiGroup: rbac.authorization.k8s.io
   kind: ClusterRole
-  name: secrets-unsealer
+  name: {{ .Values.rbac.clusterRoleName }}
 subjects:
   - apiGroup: ""
     kind: ServiceAccount
--- a/helm/sealed-secrets/templates/cluster-role.yaml
+++ b/helm/sealed-secrets/templates/cluster-role.yaml
@@ -2,7 +2,7 @@
 apiVersion: rbac.authorization.k8s.io/v1
 kind: ClusterRole
 metadata:
-  name: secrets-unsealer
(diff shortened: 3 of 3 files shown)


juan131 (changes requested):
Thanks for bringing this up @derjohn 

IMHO it doesn't make sense to have N "ClusterRoles" that define the exact same rules. IMHO, we should make it possible to create N "ClusterRoleBindings" that associate the same "ClusterRole" with a different ServiceAccount.

What do you think?

derjohn:
Well,
I completely agree with you. No duplications, if they can be avoided. (DRY).

But how would you deal with two namespaces and two different rbac.serviceAccount.name=namepace-sa-one / rbac.serviceAccount.name=namepace-sa-two ?

juan131:
As I said, we can have N "ClusterRoleBindings" that associate the same "ClusterRole" with a N different ServiceAccounts. Doesn't it make sense?

rescDev:
Hey @juan131 ,

we basically have the same issue as we're migrating to FluxV2. We need to install the controller and all resources a second time on the same cluster and face the error of having this ClusterRole twice which doesn't work. If we follow your approach, the chart would need to be able to skip the ClusterRole creation (or support custom names and allowing it to be created N times as proposed in this PR) if I understood this correctly.

So, how do we proceed with this PR ? We would really appreciate this feature

helm/sealed-secrets/values.yaml
```diff
@@ -259,6 +259,9 @@ rbac:
   ## @param rbac.pspEnabled PodSecurityPolicy
   ##
   pspEnabled: false
+  ## @param rbac.clusterRoleName Name of the clusterrole. To run multiple instances of sealed secrets with create RBAC true you need to set differnet rolenames.
```
josvazg, line 262:
```suggestion
  ## @param rbac.clusterRoleName Name of the cluster role. To run multiple instances of sealed secrets with create RBAC true you need to set different role names.
```

alemorcuq:
> Hey @juan131 ,
> 
> we basically have the same issue as we're migrating to FluxV2. We need to install the controller and all resources a second time on the same cluster and face the error of having this ClusterRole twice which doesn't work. If we follow your approach, the chart would need to be able to skip the ClusterRole creation (or support custom names and allowing it to be created N times as proposed in this PR) if I understood this correctly.
> 
> So, how do we proceed with this PR ? We would really appreciate this feature

Hi, @relusc.

I agree that would be a great option -- allowing the Cluster Role creation to be skipped. I don't think it makes sense to create N different Cluster Roles. We will create a follow-up PR to support this.

alemorcuq:
We have merged #906 adding the `rbac.clusterRole` parameter to control (in combination with the `rbac.create` parameter) the creation of the ClusterRole. This should be enough to deploy several charts but have only one of them create the ClusterRole.

As such, I'm closing this PR. Thanks to everyone to participated on this!

PR #771, opened 2022-03-10
Fix linter errors running golangci-lint

alvneiayu:
**Description of the change**

Fix all the errors listed executing golangci-lint

- Remove unused functions disabledTestRoundTrip and ssecretFuzzerFuncs
- AddToScheme return is not checked
- use buf.String() instead of string(buf.Bytes())
- Error return values is not checked

**Benefits**

Prepare the code to have no linter errors


agarcia-oss (approved):

juan131 (approved):
LGTM

PR #773, opened 2022-03-15
Update doc label to delete sealed-controller pod

alerat:
**Description of the change**

Update the documentation to delete the sealed-secret-controller pod.
Change the label filter

### DIFF
--- a/README.md
+++ b/README.md
@@ -584,7 +584,7 @@ To restore from a backup after some disaster, just put that secrets back before 
 
 ```bash
 kubectl apply -f master.key
-kubectl delete pod -n kube-system -l name=sealed-secrets-controller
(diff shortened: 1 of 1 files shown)


alvneiayu (changes requested):
hi @alerat 

First of all, thanks a lot for the PR and your time.

I am a little bit confuse with your change. I generated the controller.yaml and I installed it and the labels are:
```
  labels:
    name: sealed-secrets-controller
   ...
```
I think that this change is related to the Helm chart. This README is focus in the controller generated compiling the project (controller.yaml). So this change doesn't apply here.

Thanks a lot again

Álvaro

alerat:
Hi !

You are right. I installed sealed secret via Helm. Sorry wasted your time.

Alexandre

PR #774, opened 2022-03-16
Document use of --controller-name for kubeseal

### LINKED ISSUE #767: Discrepancy of defaults for controller name between Helm chart and CLI needs resolved
This has been brought up several times, such as #424, #571, #626, #630, and #758. Instead of only commenting on those closed issues, it seemed that the best course of action is to open a new issue and propose a possible path.

The Helm chart, by default, installs the controller with the name `sealed-secrets`, while the command line interface (CLI) tries to access the controller with the name `sealed-secrets-controller`.

The helm chart could be updated with a `fullnameOverride` value, or the CLI user could always pass `--controller-name sealed-secrets`. But this is just adding friction to enjoying this project. Right off the bat, someone new has to start troubleshooting something that could be resolved by the project ahead of time.

Pick a preferred path and then output a warning in either the Helm chart or the CLI based on whichever one is wrong. Or the value could be made as required in either area or both areas, breaking, but with a clear error, those upgrading and not currently setting the value.

The CLI has the option of falling back and trying both names, while continuing to work for all users. It should print out a warning if it has to fall back to the non-preferred name (whatever is in the default Helm chart).

Also, this doesn't seem to be clearly addressed in the FAQ or really anywhere besides issues in this repo. The option, `--controller-name`, isn't even mentioned in the [readme](https://github.com/bitnami-labs/sealed-secrets/blob/main/README.md).

brsolomon-deloitte:
...particularly in regards to mismatch in the default
name installed by the helm chart versus that implicitly used
by kubeseal by default.

Closes #767

Signed-off-by: Brad Solomon <81818815+brsolomon-deloitte@users.noreply.github.com>

**Description of the change**

**Benefits**

**Possible drawbacks**

**Applicable issues**

- fixes #

**Additional information**

### DIFF
--- a/README.md
+++ b/README.md
@@ -289,6 +289,14 @@ affecting the static YAML manifests or the controller image itself.
 NOTE: the helm chart readme still contains a deprecation notice, but it's no longer reflects reality
 and will be removed upon next release.
 
+NOTE: the helm chart by default installs the controller with the name `sealed-secrets`, while the `kubeseal` command line interface (CLI) tries to access the controller with the name `sealed-secrets-controller`. You can explicitly pass `--controller-name` to the CLI:
+
+```bash
+kubeseal --controller-name sealed-secrets <args>
+```
+
+Alternatively, you can override `fullnameOverride` on the helm chart install.
+
 ### Operator Framework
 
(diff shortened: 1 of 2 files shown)


alvneiayu (approved):
LGTM

agarcia-oss (approved):
LGTM!

PR #789, opened 2022-03-29
Remove race condition in merge

mkmik:
@luhko in  #788 noticed that a security analysis tool signaled a bad practice in our use of the `0660` mask in `ioutil.WriteFile` in `sealMergingInto`.

`sealMergingInto` should just update existing files, and thus technically that parameter is ignored because the permissions of the existing file are re-used. However, due to a bug in the windows version of the Go stdlib back in the day we passed `0660` as a workaround (which assumed that the user had a proper `umask`).

Now, the issue has been fixed (https://go-review.googlesource.com/c/go/+/234686/4/src/syscall/syscall_windows.go), so technically we can just do `return ioutil.WriteFile(filename, out.Bytes(), 0)`.

However, at the cost of just a few more lines of code, we can sidestep the whole problem and just open the file once, read from it and write the updated content.

This also handled well the (rare) case in which the file gets renamed while kubeseal is executing, which could cause kubeseal to write a new file with bad (0) permissions.


alvneiayu (approved):
LGTM

alemorcuq (commented):
Shouldn't the file be closed after we are done with it?

cmd/kubeseal/main.go
```diff
@@ -467,13 +472,17 @@ func sealMergingInto(in io.Reader, filename string, codecs runtimeserializer.Cod
...
+	if err := f.Truncate(0); err != nil {
+		return err
+	}
+	if _, err := f.Seek(0, 0); err != nil {
+		return err
+	}
+	if _, err := io.Copy(f, &out); err != nil {
+		return err
+	}
+	return nil
```
alvneiayu, line 485: As @alemorcuq, we are missing to close it no?

alvneiayu (changes requested):

alvneiayu (approved):
LGTM

alemorcuq (approved):

PR #808, opened 2022-03-31
Use a specific VIB action version instead of the rolling main branch

alemorcuq:
**Description of the change**
Use a tagged version instead of main for the VIB action


agarcia-oss (approved):
Totally agree with the change. We should always use fixed version from every dependency/CI Action

agarcia-oss:
Since the vib-action is under active development, I think is more convenient not to used a pinned version and revert this PR changes.

PR #811, opened 2022-04-06
feat(kubernetes/chart): explicitly mount service-account-token in deployment

guethb:
**Description of the change**

Our environment has some restrictions on service-accounts: We're not allowed to use automountServiceAccountToken=true as default for a service-account but have to explicit enable the service-account-mount in each deployment.
It would be nice, if you could merge that change to make Reloader usable in such environments.

**Benefits**

Implementation for https://github.com/kubernetes/kubernetes/issues/57601

**Possible drawbacks**

no known drawbacks

**Applicable issues**

no issues
- fixes #

**Additional information**


helm/sealed-secrets/templates/deployment.yaml
```diff
@@ -40,6 +40,9 @@ spec:
       securityContext: {{- omit .Values.podSecurityContext "enabled" | toYaml | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ include "sealed-secrets.serviceAccountName" . }}
+      {{- if hasKey .Values "automountServiceAccountToken" }}
```
alemorcuq, line 43: I would change the condition to use the same approach across the whole chart:

```suggestion
      {{- if .Values.automountServiceAccountToken }}
```

The same applies to the ServiceAccount file.

helm/sealed-secrets/values.yaml
```diff
@@ -144,6 +144,10 @@ containerSecurityContext:
   readOnlyRootFilesystem: true
   runAsNonRoot: true
   runAsUser: 1001
+
+## @param automountServiceAccountToken whether to automatically mount the service account API-token to a particular pod
+# automountServiceAccountToken: true
```
alemorcuq, line 149:
```suggestion
automountServiceAccountToken: ""
```

helm/sealed-secrets/values.yaml
```diff
@@ -300,6 +304,8 @@ serviceAccount:
   ## If not set and create is true, a name is generated using the sealed-secrets.fullname template
   ##
   name: ""
+  ## @param serviceAccount.automountServiceAccountToken Specifies, whether to mount the service account API-token
+  # automountServiceAccountToken: true
```
alemorcuq, line 308:
```suggestion
  automountServiceAccountToken: ""
```

alemorcuq (changes requested):
Thanks for the PR! This is a useful feature to have. Could you please increase the chart version and add the new parameters to the README file?

alemorcuq (changes requested):
Thanks, @guethb!

Actually, when I asked to add the values to the README.md file, I was referring to the chart's README.md file located at `helm/sealed-secrets/README.md`. My fault for not being specific.

In that file, you will find some tables containing all the parameters in the chart. Could you please update them to add the new values?

helm/sealed-secrets/templates/deployment.yaml
guethb: Done

guethb:
@alemorcuq changes were committed, please review again.

alvneiayu (approved):
LGTM. Thanks a lot for your work @guethb

alvneiayu (approved):
LGTM

agarcia-oss (approved):

PR #828, opened 2022-04-21
fix: rbac permission

### LINKED ISSUE #827: helm RBAC should be more restrictive
**Which component**:
controller

**Describe the bug**
This PR https://github.com/bitnami-labs/sealed-secrets/pull/715 added the `get service` permission to `system:authenticated` which is needed to lookup the named port of the service.
While this works as expected I think the helm chart is missing the restriction to the resource name of the service. This in turn will allow every authenticated user to `get` any service in the same namespace as sealed-secrets is deployed.

**Expected behavior**
I think this should be limited to just "sealed-secrets-controller" like in the `controller.jsonnet` file in the same PR.

**Additional context**
I'll happily provide a PR fix for this :)

TheKangaroo:
**Description of the change**

add rbac name selector for get service permission in the helm chart.
I think this was missing in #715

Signed-off-by: Till Adam <lxmail@web.de>

**Benefits**

RBAC could be more restrictive

**Possible drawbacks**

none

**Applicable issues**

- fixes #827 

**Additional information**

none

### DIFF
--- a/helm/sealed-secrets/templates/role.yaml
+++ b/helm/sealed-secrets/templates/role.yaml
@@ -37,6 +37,8 @@ metadata:
 rules:
   - apiGroups:
       - ""
+    resourceNames:
+      - {{ include "sealed-secrets.fullname" . }}
     resources:
       - services
     verbs:
--- a/helm/sealed-secrets/Chart.yaml
+++ b/helm/sealed-secrets/Chart.yaml
@@ -16,4 +16,4 @@ maintainers:
     email: mmikulicic@gmail.com
 name: sealed-secrets
 type: application
-version: 2.1.7
+version: 2.1.8


helm/sealed-secrets/templates/role.yaml
```diff
@@ -37,6 +37,10 @@ metadata:
 rules:
   - apiGroups:
       - ""
+    resourceNames:
+      - 'http:{{ include "sealed-secrets.fullname" . }}:'
+      - 'http:{{ include "sealed-secrets.fullname" . }}:http'
```
alemorcuq, line 42: I think these two are not needed, only `{{ include "sealed-secrets.fullname" . }}`. This would align it with the `controller.jsonnet` file we mention.

`kubeseal` needs to fetch the service to get the port name (in our case `http`) that is later used through the proxy, that's why the `services/proxy` resource includes them.

alemorcuq (changes requested):
Thanks for the PR! Yes, RBAC can be more strict as you said. Could you check my comment?

Regarding the chart version, yes, please increase it.

helm/sealed-secrets/templates/role.yaml
TheKangaroo: I'm not sure why I put in the other two lines. I guess it was a copy paste mistake.
Fixed it and bumped the patch version for the chart

alvneiayu (approved):
LGTM

alemorcuq (approved):
Thanks!

PR #873, opened 2022-06-24
feat: override rate limit

### LINKED ISSUE #870: /v1/verify endpoint throttling
Hi there, 

We're trying to take advantage of [the `/v1/verify` endpoint](https://github.com/bitnami-labs/sealed-secrets/blob/b023eafcf510521dc950d959a7eab8e1d5869d29/cmd/controller/server.go#L50) by integrating the sealed-secrets validation during the CI/CD process. this way developers are informed before hitting deployment if something is uncorrectly sealed.

At the moment, there is [a built-in rate limiter](https://github.com/bitnami-labs/sealed-secrets/blob/b023eafcf510521dc950d959a7eab8e1d5869d29/cmd/controller/server.go#L128) with low hardcoded limits which makes it hard to use at scale.

Before trying any PR i have some questions :

 - do you think sealed-secret validation in CI makes sense or have other ideas ?
 - what about adding an `SEALED_SECRETS_VERIFY_RPS_LIMIT` env var (default 2) which allow to increase this ? with 0 it could disable the built-in rate-limiter and let the ingress take care of that?

Thanks !

devthejo:
fix #870

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -50,8 +50,8 @@ var (
 	namespaceAll         = flag.Bool("all-namespaces", true, "Scan all namespaces or only the current namespace (default=true).")
 	additionalNamespaces = flag.String("additional-namespaces", "", "Comma-separated list of additional namespaces to be scanned.")
 	labelSelector        = flag.String("label-selector", "", "Label selector which can be used to filter sealed secrets.")
-	rateLimitBurst       = flag.Int("rate-limit-burst", 2, "Number of requests per burst for verify endpoint")
-	rateLimitDelay       = flag.Int("rate-limit-delay", 2, "Number of seconds to sleep between bursts for verify endpoint")
+	rateLimitPerSecond   = flag.Int("rate-limit", 2, "Number of allowed sustained request per second for verify endpoint")
+	rateLimitBurst       = flag.Int("rate-limit-burst", 2, "Number of requests allowed to exceed the rate limit per second for verify endpoint")
 
 	oldGCBehavior = flag.Bool("old-gc-behaviour", false, "Revert to old GC behavior where the controller deletes secrets instead of delegating that to k8s itself.")
 
@@ -268,7 +268,7 @@ func main2() error {
 		return []*x509.Certificate{cert}, nil
 	}
 
-	server := httpserver(cp, controller.AttemptUnseal, controller.Rotate, *rateLimitBurst, *rateLimitDelay)
+	server := httpserver(cp, controller.AttemptUnseal, controller.Rotate, *rateLimitBurst, *rateLimitPerSecond)
 
 	sigterm := make(chan os.Signal, 1)
 	signal.Notify(sigterm, syscall.SIGTERM)
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -50,7 +50,7 @@ var (
 	namespaceAll         = flag.Bool("all-namespaces", true, "Scan all namespaces or only the current namespace (default=true).")
 	additionalNamespaces = flag.String("additional-namespaces", "", "Comma-separated list of additional namespaces to be scanned.")
 	labelSelector        = flag.String("label-selector", "", "Label selector which can be used to filter sealed secrets.")
-	rateLimitBurst       = flag.Int("rate-limit-burst", 2, "Number of requests per burst")
+	rateLimitBurst       = flag.Int("rate-limit-burst", 2, "Number of requests per burst for verify endpoint")
 	rateLimitDelay       = flag.Int("rate-limit-delay", 2, "Number of seconds to sleep between bursts")
 
 	oldGCBehavior = flag.Bool("old-gc-behaviour", false, "Revert to old GC behavior where the controller deletes secrets instead of delegating that to k8s itself.")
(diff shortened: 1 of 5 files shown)


cmd/controller/main.go
```diff
@@ -50,6 +50,8 @@ var (
 	namespaceAll         = flag.Bool("all-namespaces", true, "Scan all namespaces or only the current namespace (default=true).")
 	additionalNamespaces = flag.String("additional-namespaces", "", "Comma-separated list of additional namespaces to be scanned.")
 	labelSelector        = flag.String("label-selector", "", "Label selector which can be used to filter sealed secrets.")
+	rateLimitBurst       = flag.Int("rate-limit-burst", 2, "Number of requests per burst")
```
revolunet, line 53:
```suggestion
	rateLimitBurst       = flag.Int("rate-limit-burst", 2, "Number of requests per burst for verify endpoint")
```

cmd/controller/main.go
```diff
@@ -50,6 +50,8 @@ var (
...
+	rateLimitDelay       = flag.Int("rate-limit-delay", 2, "Number of seconds to sleep between bursts")
```
revolunet, line 54:
```suggestion
	rateLimitDelay       = flag.Int("rate-limit-delay", 2, "Number of seconds to sleep between bursts for verify endpoint")
```

helm/sealed-secrets/values.yaml
```diff
@@ -59,6 +59,12 @@ updateStatus: true
 keyrenewperiod: ""
 ## @param command Override default container command
 ##
+ratelimitburst: 2
+## @param command Override default container command
```
revolunet, line 63:
```suggestion
## @param ratelimitburst Number of requests per burst for verify endpoint
```

helm/sealed-secrets/values.yaml
```diff
@@ -59,6 +59,12 @@ updateStatus: true
...
+##
+ratelimitdelay: 2
+## @param command Override default container command
```
revolunet, line 66:
```suggestion
## @param ratelimitdelay Number of seconds to sleep between bursts for verify endpoint
```

helm/sealed-secrets/values.yaml
```diff
@@ -59,6 +59,12 @@ updateStatus: true
 keyrenewperiod: ""
 ## @param command Override default container command
```
revolunet, line 60:
```suggestion
## @param ratelimitburst Number of requests per burst for verify endpoint
```

alvneiayu:
hi @devthejo 

First of all, thanks a lot for the PR and your help.

The reason of the error that you are reproducing in our tests is that you need to split the patch in two of them. One for the feature included in the controller and the second one for the changes in the chart.

Please, remove the chart charges from this PR and create a new one.

Thanks a lot

Álvaro

cmd/controller/main.go
```diff
@@ -50,6 +50,8 @@ var (
 	namespaceAll         = flag.Bool("all-namespaces", true, "Scan all namespaces or only the current namespace (default=true).")
 	additionalNamespaces = flag.String("additional-namespaces", "", "Comma-separated list of additional namespaces to be scanned.")
 	labelSelector        = flag.String("label-selector", "", "Label selector which can be used to filter sealed secrets.")
+	rateLimitBurst       = flag.Int("rate-limit-burst", 2, "Number of requests per burst for verify endpoint")
+	rateLimitDelay       = flag.Int("rate-limit-delay", 2, "Number of seconds to sleep between bursts for verify endpoint")
```
alemorcuq, line 54: Unless I'm missing something, I think this isn't really a delay. Looking at the [documentation](https://pkg.go.dev/github.com/throttled/throttled#RateQuota):

> MaxRate specified the maximum sustained rate of requests and must be greater than zero. MaxBurst defines the number of requests that will be allowed to exceed the rate in a single burst and must be greater than or equal to zero.

> Rate{PerSec(1), 0} would mean that after each request, no more requests will be permitted for that client for one second. Rate{PerSec(2), 0} permits one request per 0.5 seconds rather than two requests in one second. In practice, you probably want to set MaxBurst >0 to provide some flexibility to clients that only need to make a handful of requests. In fact a MaxBurst of zero will *never* permit a request with a quantity greater than one because it will immediately exceed the limit.

alemorcuq (changes requested):

devthejo:
> hi @devthejo
> 
> First of all, thanks a lot for the PR and your help.
> 
> The reason of the error that you are reproducing in our tests is that you need to split the patch in two of them. One for the feature included in the controller and the second one for the changes in the chart.
> 
> Please, remove the chart charges from this PR and create a new one for the chart changes.
> 
> Thanks a lot
> 
> Álvaro

Hi @alvneiayu,

Thanks for your attention, got it, I will do that very soon

cmd/controller/main.go
devthejo: I was biased by the http response that is talking about burst and delay, OK, got it, I will change that, thanks

devthejo: it's changed

devthejo:
@alvneiayu I removed the chart changes from this PR and opened a separated PR for chart here  #876

alvneiayu (approved):
LGTM

PR #876, opened 2022-06-30
feat: override rate limit for chart

devthejo:
related to #873


helm/sealed-secrets/values.yaml
```diff
@@ -61,6 +61,12 @@ updateStatus: true
 ## keyrenewperiod: "720h30m"
 ##
 keyrenewperiod: ""
+## @param ratelimit Number of allowed sustained request per second for verify endpoint
+##
+ratelimit: 2
+## @param ratelimitburst Number of requests allowed to exceed the rate limit per second for verify endpoint
+##
+ratelimitburst: 2
```
alemorcuq, line 69: Could you please camel case both values?

I'd also set them empty by default, as that would also use the default rate limit from the controller.

alemorcuq (changes requested):
Thanks for the PR!

PR #885, opened 2022-07-14
feat(kubernetes/chart): explicitly mount service-account-token in deployment

guethb:
This PR has already been treated (https://github.com/bitnami-labs/sealed-secrets/pull/811/), but changes have not been merged correctly.

**Description of the change**
Our environment has some restrictions on service-accounts: We're not allowed to use automountServiceAccountToken=true as default for a service-account but have to explicit enable the service-account-mount in each deployment.
It would be nice, if you could merge that change to make Reloader usable in such environments.

**Benefits**
Implementation for https://github.com/kubernetes/kubernetes/issues/57601

**Possible drawbacks**

no known drawbacks

**Applicable issues**
no issues

- fixes #

**Additional information**
no additional information

### DIFF
--- a/helm/sealed-secrets/README.md
+++ b/helm/sealed-secrets/README.md
@@ -84,7 +84,7 @@ The command removes all the Kubernetes components associated with the chart and 
 
 | Name                                              | Description                                                                          | Value                               |
 | ------------------------------------------------- | ------------------------------------------------------------------------------------ | ----------------------------------- |
-| `image.registry`                                  | Sealed Secrets image registry                                                        | `docker.io`                           |
+| `image.registry`                                  | Sealed Secrets image registry                                                        | `docker.io`                         |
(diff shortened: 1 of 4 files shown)


helm/sealed-secrets/Chart.yaml
```diff
@@ -16,4 +16,4 @@ maintainers:
     email: mmikulicic@gmail.com
 name: sealed-secrets
 type: application
-version: 2.3.0
+version: 2.4.0
```
alemorcuq, line 19: Could you please remove this change? We are going to release a new version of the chart today (matching new release `v0.18.1`) so we will create a new PR increasing the version ourselves.

alemorcuq (changes requested):
Thanks! LGTM, but please, check my comment.

PR #897, opened 2022-08-01
Fix some minor issues in Readme

josvazg:
**Description of the change**

Minor fixes on the README.

**Benefits**

Reduce confusion and friction for newcomers.

**Possible drawbacks**

N/A

**Applicable issues**

- Fix the `kubeseal` install instructions to work with modern Go versions, like go 1.17 onward.
- Suggest a Helm Chart install command so that no env vars or extra arguments need to be used with the `kubeseal` CLI.
- Fix broken link to the GHCR registry.

### DIFF
--- a/README.md
+++ b/README.md
@@ -297,7 +297,11 @@ NOTE: the helm chart by default installs the controller with the name `sealed-se
 kubeseal --controller-name sealed-secrets <args>
 ```
 
-Alternatively, you can override `fullnameOverride` on the helm chart install.
+Alternatively, you can override `fullnameOverride` on the helm chart install. Note also that `kubeseal` assumes that the controller is installed within the `kube-system` namespace by default. So if you want to use the `kubeseal` CLI withut havving to pass the expected controller name and namespace you should install the Helm Chart like this:
+
+```bash
+helm install sealed-secrets -n kube-system --set-string fullnameOverride=sealed-secrets-controller sealed-secrets/sealed-secrets 
+```
 
 ### Homebrew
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -643,7 +653,7 @@ kubeseal <mysecret.json >mysealedsecret.json
 
 ### How to verify the images?
 
-Our images are being signed using [cosign](https://github.com/sigstore/cosign). The signatures have been saved in our [GitHub Container Registry](https://github.com/bitnami-labs/sealed-secrets/pkgs/container/sealed-secrets/signs).
+Our images are being signed using [cosign](https://github.com/sigstore/cosign). The signatures have been saved in our [GitHub Container Registry](https://ghcr.io/bitnami-labs/sealed-secrets-controller).
```
alemorcuq, line 656:
```suggestion
Our images are being signed using [cosign](https://github.com/sigstore/cosign). The signatures have been saved in our [GitHub Container Registry](https://ghcr.io/bitnami-labs/sealed-secrets-controller/signs).
```

There's a specific registry for the signatures

README.md
```diff
@@ -297,7 +297,11 @@ NOTE: the helm chart by default installs the controller with the name `sealed-se
 kubeseal --controller-name sealed-secrets <args>
 ```
 
-Alternatively, you can override `fullnameOverride` on the helm chart install.
+Alternatively, you can override `fullnameOverride` on the helm chart install. Note also that `kubeseal` assumes that the controller is installed within the `kube-system` namespace by default. So if you want to use the `kubeseal` CLI withut havving to pass the expected controller name and namespace you should install the Helm Chart like this:
```
alemorcuq, line 300:
```suggestion
Alternatively, you can set `fullnameOverride` when installing the chart to override the name . Note also that `kubeseal` assumes that the controller is installed within the `kube-system` namespace by default. So if you want to use the `kubeseal` CLI without having to pass the expected controller name and namespace you should install the Helm Chart like this:
```

alemorcuq (changes requested):
Thanks for this, I've added a couple of suggestions.

alemorcuq (approved):

PR #899, opened 2022-08-01
Qualify RSA replacement recommendation

josvazg:
**Description of the change**

Qualify the RSA replacement recommendation in the **crypto** document.

**Benefits**

Clarify what requirements need to happen first before RSA can be replaced in the Sealed Secrets code.

**Applicable issues**

Otherwise it might look like this is an imminent improvement when the reality is that the replacement algorithm has not been identified or selected yet and most probably does not have a reliable or usable Golang implementation at the moment. Without that, moving to a less susceptible algorithm or just increasing key size might be a better option in the middle term.

### DIFF
--- a/docs/developer/crypto.md
+++ b/docs/developer/crypto.md
@@ -166,7 +166,8 @@ Shor algorithm can be used to solve in a reasonable time 3 mathematical problems
 
 ### Recommendations
 
-Replace RSA as soon as possible. This recommendation must be the highest priority regarding the post-quantum security of Sealed Secrets. There are three serious candidates to use instead of RSA: LMS and XMSS, which are Lattice-based, and McEliece with random Goppa codes, which is code-based and relies on SDP (Syndrome Decoding Problem).
+Replace RSA whenever feasible. This recommendation must be the highest priority regarding the post-quantum security of Sealed Secrets.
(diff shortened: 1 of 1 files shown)


docs/developer/crypto.md
```diff
@@ -166,10 +166,16 @@ Shor algorithm can be used to solve in a reasonable time 3 mathematical problems
 
 ### Recommendations
 
-Replace RSA. This recommendation must be the highest priority regarding the post-quantum security of Sealed Secrets.
+Replace RSA whenever feasible. This recommendation must be the highest priority regarding the post-quantum security of Sealed Secrets.
 There are three serious candidates to use instead of RSA: LMS and XMSS, which are Lattice-based, and McEliece with random Goppa codes, which is code-based and relies on SDP (Syndrome Decoding Problem).
 Those three algorithms are serious candidates for RSA replacement and the choice must be done carefully, without forgetting to study other algorithms such as NTRU.
 
+It is important to qualify this recommendation with a couple of prerequisites:
+- An standard or clear recommended replacement Public Key Cryptography Algorithm emerges in the industry.
```
alemorcuq, line 174:
```suggestion
- A standard or clear recommended replacement Public Key Cryptography Algorithm emerges in the industry.
```

alemorcuq (changes requested):
Thanks! Just a minor typo.

alemorcuq (approved):

PR #900, opened 2022-08-02
Use gotestsum for better test reporting.

mpermar:
**Description of the change**

Experimenting with moving from plain `go test` to use [gotestsum](https://github.com/gotestyourself/gotestsum) which provides formatted test output but also a test summary. 

**Benefits**

Better developer experience on testing.

**Possible drawbacks**

Not sure if we might face any limitations with `gotestsum`. Their issues section has a few items but none look relevant.

Signed-off-by: Martin Perez <martinpe@vmware.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,6 +67,7 @@ jobs:
       matrix:
         go: ["1.17.11", "1.18.3"]
         os: [macos-latest, windows-latest, ubuntu-latest]
+        gotestsum: ["1.8.1"]
     steps:
 
     - name: Set up Go 1.x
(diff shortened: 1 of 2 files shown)


josvazg:
This LGTM, but just one question. Why not adding `gotestsum` also to the go.mod set of dependencies?

alemorcuq:
> This LGTM, but just one question. Why not adding `gotestsum` also to the go.mod set of dependencies?

My 2cents: because it is not a code dependency, it won't be used anywhere and `go mod tidy` will remove it as it's unused.

alemorcuq (approved):

josvazg:
Makes sense, then no further questions.

josvazg (approved):

josvazg:
Side note: we might want to work on auto installing or auto-detecting our dev tools.

PR #901, opened 2022-08-03
Remove clientconfig and related global variables

josvazg:
**Description of the change**

Remove the `clientConfig` and namespace wrapper func `namespaceFromClientConfig` global variables.

**Benefits**

Disentangle the code so we can make further enhancements, such as get rid of the `run` method with a gazillion unnamed arguments.

### DIFF
--- a/cmd/kubeseal/main.go
+++ b/cmd/kubeseal/main.go
@@ -116,7 +116,7 @@ func initClient(kubeConfigPath string, cfgOverrides clientcmd.ConfigOverrides, r
 	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
 	loadingRules.DefaultClientConfig = &clientcmd.DefaultClientConfig
 	loadingRules.ExplicitPath = kubeConfigPath
-	return clientcmd.NewInteractiveDeferredLoadingClientConfig(loadingRules, &cfgOverrides, w)
+	return clientcmd.NewInteractiveDeferredLoadingClientConfig(loadingRules, &cfgOverrides, r)
 }
 
 func initNamespaceFuncFromClient(clientConfig clientcmd.ClientConfig) namespaceFn {
--- a/cmd/kubeseal/main.go
+++ b/cmd/kubeseal/main.go
@@ -109,10 +109,10 @@ func initUsualKubectlFlags(flagset *flag.FlagSet) *clientcmd.ConfigOverrides {
 }
 
 func initClientFromFlags() clientcmd.ClientConfig {
-	return initClient(*kubeconfig, *globalOverrides, os.Stdout)
+	return initClient(*kubeconfig, *globalOverrides, os.Stdin)
 }
 
-func initClient(kubeConfigPath string, cfgOverrides clientcmd.ConfigOverrides, w io.Reader) clientcmd.ClientConfig {
+func initClient(kubeConfigPath string, cfgOverrides clientcmd.ConfigOverrides, r io.Reader) clientcmd.ClientConfig {
 	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
 	loadingRules.DefaultClientConfig = &clientcmd.DefaultClientConfig
 	loadingRules.ExplicitPath = kubeConfigPath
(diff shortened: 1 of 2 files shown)


josvazg:
I made a mistake wen solving conflicts...

cmd/kubeseal/main.go
```diff
@@ -108,6 +101,28 @@ func init() {
...
+func initUsualKubectlFlags(flagset *flag.FlagSet) *clientcmd.ConfigOverrides {
+	overrides := clientcmd.ConfigOverrides{}
+	kflags := clientcmd.RecommendedConfigOverrideFlags("")
+	clientcmd.BindOverrideFlags(&overrides, flagset, kflags)
+	return &overrides
+}
+
+func initClientFromFlags() clientcmd.ClientConfig {
+	return initClient(*kubeconfig, *globalOverrides, os.Stdout)
+}
```
alemorcuq, line 113: What are we expecting to read from `stdout`? I've noticed the original code reads from `stdin`, why this change?

cmd/kubeseal/main.go
```diff
@@ -108,6 +101,28 @@ func init() {
...
+
+func initClient(kubeConfigPath string, cfgOverrides clientcmd.ConfigOverrides, w io.Reader) clientcmd.ClientConfig {
```
alemorcuq, line 115:
```suggestion
func initClient(kubeConfigPath string, cfgOverrides clientcmd.ConfigOverrides, r io.Reader) clientcmd.ClientConfig {
```

Afaik, it is a common practice to use `r` for Readers and `w` for Writers.

cmd/kubeseal/main.go
```diff
@@ -71,17 +71,17 @@ var (
...
-	clientConfig clientcmd.ClientConfig
-
-	// testing hook for clientConfig.Namespace()
-	namespaceFromClientConfig = func() (string, bool, error) { return clientConfig.Namespace() }
 )
 
-func init() {
+type namespaceFn func() (string, bool, error)
+
+func initFlags() {
```
alemorcuq, line 84: Why are we replacing `init` with `initFlags`?

cmd/kubeseal/main.go
josvazg: This is a mistake. Will fix it

cmd/kubeseal/main.go
josvazg: `init()` is magic, it is always executed before the `main` function.

It means that such initialization is happening no matter what, and in this case that is NOT required. You can make the case for when you are, for instance, loading device or SQL driver implementations coming from dependencies.

Using `initFlags` instead we have full control on when the flags are parsed or not.

cmd/kubeseal/main.go
josvazg, line 113:
```suggestion
func initClientFromFlags() clientcmd.ClientConfig {
	return initClient(*kubeconfig, *globalOverrides, os.Stdin)
}
```

cmd/kubeseal/main.go
josvazg: Right! Thanks!

cmd/kubeseal/main.go
josvazg: Done

cmd/kubeseal/main_test.go
```diff
@@ -102,6 +93,21 @@ func tmpfile(t *testing.T, contents []byte) string {
...
+	err := flagset.Parse([]string{"-n", "default"})
+	if err != nil {
+		fmt.Printf("flagset parse err: %v\n", err)
+		os.Exit(1)
+	}
+	return testOverrides
+}
+
+func testClientConfig() clientcmd.ClientConfig {
+	return initClient("", *testConfigOverrides(), os.Stdout)
```
josvazg, line 108:
```suggestion
	return initClient("", *testConfigOverrides(), os.Stdin)
```

cmd/kubeseal/main.go
```diff
@@ -108,6 +101,28 @@ func init() {
...
+
+func initClientFromFlags() clientcmd.ClientConfig {
+	return initClient(*kubeconfig, *globalOverrides, os.Stdin)
+}
+
+func initClient(kubeConfigPath string, cfgOverrides clientcmd.ConfigOverrides, r io.Reader) clientcmd.ClientConfig {
+	loadingRules := clientcmd.NewDefaultClientConfigLoadingRules()
+	loadingRules.DefaultClientConfig = &clientcmd.DefaultClientConfig
+	loadingRules.ExplicitPath = kubeConfigPath
+	return clientcmd.NewInteractiveDeferredLoadingClientConfig(loadingRules, &cfgOverrides, w)
```
alemorcuq, line 119:
```suggestion
	return clientcmd.NewInteractiveDeferredLoadingClientConfig(loadingRules, &cfgOverrides, r)
```

alemorcuq (changes requested):

alemorcuq (approved):

PR #902, opened 2022-08-03
Rename package folders to sealedsecrets

josvazg:
**Description of the change**

Rename all package folders named `sealed-secrets` as `sealedsecrets`.

Go package names do not allow for the use of `-` and, [although the folder can be named differently to the package, it is discouraged](https://go.dev/doc/effective_go#package-names).

**Benefits**

Following standard Go package name conventions will allow for easier dependency upgrades. For instance, the newer [code-generator](https://github.com/kubernetes/code-generator) version seems to enforce the package names generated to match the directory, so for `sealed-secrets` it generates an error in the informer generation:

```
unable to format file "github.com/bitnami-labs/sealed-secrets/pkg/client/informers/externalversions/sealed-secrets/interface.go" (3:15: expected ';', found '-').
```

```bash
$ head -3 pkg/client/informers/externalversions/sealed-secrets/interface.go 
// Code generated by informer-gen. DO NOT EDIT.

package sealed-secrets
```


mkmik (approved):
If it makes it easier to regenerate the generated code I'm all for it

alemorcuq (approved):

PR #904, opened 2022-08-04
Update CI to support Go 1.19

alemorcuq:
**Description of the change**
- Run the CI using the latest versions from Go (1.18.5 and 1.19).
- Remove Go 1.17 from the CI as it is not maintained.
- Use Go 1.18.5 as the release version.
- Update several CI/CD tools.

**Benefits**
- Up to date CI/CD 
- More secure releases

### DIFF
--- a/pkg/flagenv/flagenv.go
+++ b/pkg/flagenv/flagenv.go
@@ -8,16 +8,16 @@
 // This example will make it possible to set the default of --my_flag also via the MY_PROG_MY_FLAG
 // env var:
 //
-//    var myflag = flag.String("my_flag", "", "some flag")
+//	var myflag = flag.String("my_flag", "", "some flag")
 //
-//    func init() {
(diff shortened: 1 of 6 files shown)


pkg/flagenv/flagenv.go
```diff
@@ -8,16 +8,16 @@
 // This example will make it possible to set the default of --my_flag also via the MY_PROG_MY_FLAG
 // env var:
 //
-//    var myflag = flag.String("my_flag", "", "some flag")
+//	var myflag = flag.String("my_flag", "", "some flag")
```
josvazg, line 11: Is this a tabs vs spaces thing?
why dd this change?

alemorcuq: This was automatic formatting by `goimports`, which is one of the linting tools we use.

josvazg: ok

josvazg (approved):

PR #905, opened 2022-08-04
Upgrade code-generator tool and update generated code

josvazg:
**Description of the change**

Upgrade `code-generator` tool to a newer version. Thanks to https://github.com/bitnami-labs/sealed-secrets/pull/902

**Benefits**

Not upgrading this tools is risky as it created dependency upgrading issues to keep up with other libraries. This upgrade was already not without issues.


Makefile
```diff
@@ -44,6 +44,11 @@ all: controller kubeseal
 
 generate: $(GO_FILES)
 	$(GO) mod vendor && $(GO) generate $(GO_PACKAGES)
+	@# code-generator insists in generating the file under diretory:
+	@# github.com/bitnami-labs/sealeds-secrets
+	@# instead of just updating ./pkg for that reason we generate at gentmp
+	@# and then move to ./pkg
+	cp -r gentmp/github.com/bitnami-labs/sealed-secrets/pkg . && rm gentmp/ -rf
```
alemorcuq, line 51: This looks tricky. I've been playing around and it seems to work, but I haven't found a cleaner way of doing this.

Maybe we should expend some time on this in the future.

alemorcuq (dismissed):

Makefile
josvazg: Let me add a TODO there...

Makefile
```diff
@@ -44,6 +44,11 @@ all: controller kubeseal
 
 generate: $(GO_FILES)
 	$(GO) mod vendor && $(GO) generate $(GO_PACKAGES)
+	@# code-generator insists in generating the file under diretory:
+	@# github.com/bitnami-labs/sealeds-secrets
+	@# instead of just updating ./pkg for that reason we generate at gentmp
+	@# and then move to ./pkg
```
josvazg, line 50:
```suggestion
	@# TODO: remove as soon as a proper way forward is found:
	@# code-generator insists in generating the file under directory:
	@# github.com/bitnami-labs/sealeds-secrets/...
	@# instead of just updating ./pkg 
	@# for that reason we generate at gentmp and then move it all to ./pkg
```

alemorcuq (approved):

PR #906, opened 2022-08-04
Allow skipping ClusterRole creation

alemorcuq:
**Description of the change**
Follow-up for #749 

Adds a parameter to skip the ClusterRole creation so that it doesn't fail to install on a cluster that already has the ClusterRole.

**Benefits**
Users can install the chart easily in several namespaces of the same cluster.


helm/sealed-secrets/templates/cluster-role.yaml
```diff
@@ -1,4 +1,4 @@
-{{ if .Values.rbac.create }}
+{{ if and .Values.rbac.create .Values.rbac.clusterRole }}
```
josvazg, line 1: TIL

josvazg (approved):

PR #911, opened 2022-08-10
Update README file

kiela:
**Description of the change**
It makes the README file clean and neat.

**Benefits**
Increase simplicity of documentation.

**Possible drawbacks**
None.

**Applicable issues**
None.

**Additional information**
None.


alvneiayu:
hi @kiela 

First of all, thanks a lot for the PR and the hard work.

We included a change in our README generating conflicts with this. Could you solve the conflict, please?

Thanks a lot

Álvaro

README.md
```diff
@@ -214,9 +214,9 @@ seal a secret for a namespace, it cannot be moved in another namespace and decry
...
 
-Furthermore, namespaces are not the only level at which RBAC configurations can decide who can see which secret. In fact, it's possible that users can access a secret called `foo` in a given namespace but not any other secret in the same namespace. We cannot thus by default let users freely rename SealedSecret resources otherwise a malicious user would be able to decrypt any SealedSecret for that namespace by just renaming it to overwrite the one secret she does have access to. We use the same mechanism used to include the namespace in the encryption key to also include the secret name.
+Furthermore, namespaces are not the only level at which RBAC configurations can decide who can see which secret. In fact, it's possible that users can access a secret called `foo` in a given namespace but not any other secret in the same namespace. We cannot thus by default let users freely rename `SealedSecret` resources otherwise a malicious user would be able to decrypt any SealedSecret for that namespace by just renaming it to overwrite the one secret he does have access to. We use the same mechanism used to include the namespace in the encryption key to also include the secret name.
```
alemorcuq, line 217:
```suggestion
Furthermore, namespaces are not the only level at which RBAC configurations can decide who can see which secret. In fact, it's possible that users can access a secret called `foo` in a given namespace but not any other secret in the same namespace. We cannot thus by default let users freely rename `SealedSecret` resources otherwise a malicious user would be able to decrypt any SealedSecret for that namespace by just renaming it to overwrite the one secret they do have access to. We use the same mechanism used to include the namespace in the encryption key to also include the secret name.
```

README.md
```diff
@@ -465,82 +465,82 @@ TL;DR:
...
 
 The *sealing key* renewal and SealedSecret rotation are **not a substitute** for rotating your actual secrets.
 
 A core value proposition of this tool is:
 
-> Encrypt your Secret into a SealedSecret, which *is* safe to store - even to a public repository.
+> Encrypt your Secret into a SealedSecret, which *is* safe to store - even inside a public repository.
 
-If you store anything in a version control storage, and in a public one in particular, you must assume
+If you store anything in a version control storage, and in a public one, in particular, you must assume
```
alemorcuq, line 494:
```suggestion
If you store anything in a version control storage, and in a public one in particular, you must assume
```

No need for this comma.

README.md
```diff
@@ -465,82 +465,82 @@ TL;DR:
...
 
-Sure, the content of those encrypted emails is not secure, as an attacker might now be able to decrypt them, but what's done is done. Your sudden loss of ability to read those emails surely doesn't undo the damage; if anything, it's worse because you no longer know for sure what secret the attacker got to know. What you really want to do is to make sure that your friend stops using your old key and that from now on all further communication is encrypted with a new key pair (i.e. your friend must know about that new key).
+Sure, the content of those encrypted emails is not secure, as an attacker might now be able to decrypt them, but what's done is done.Your sudden loss of the ability to read those emails surely doesn't undo the damage. If anything, it's worse because you no longer know for sure what secret the attacker got to know. What you really want to do is to make sure that your friend stops using your old key and that from now on all further communication is encrypted with a new key pair (i.e. your friend must know about that new key).
```
alemorcuq, line 519:
```suggestion
Sure, the content of those encrypted emails is not secure, as an attacker might now be able to decrypt them, but what's done is done. Your sudden loss of the ability to read those emails surely doesn't undo the damage. If anything, it's worse because you no longer know for sure what secret the attacker got to know. What you really want to do is to make sure that your friend stops using your old key and that from now on all further communication is encrypted with a new key pair (i.e. your friend must know about that new key).
```

README.md
```diff
@@ -557,7 +557,7 @@ in your version control system (`kubeseal --re-encrypt` doesn't update the in-cl
 
 Currently, old keys are not garbage collected automatically.
 
-It's a good idea to periodically re-encrypt your SealedSecrets. But as mentioned above, don't lull yourself in a false sense of security: you must assume the old version of the SealedSecret (the one encrypted with a key you think of as dead) is still potentially around and accessible to attackers. I.e. re-encryption is not a substitute for periodically rotating your actual secrets.
+It's a good idea to periodically re-encrypt your SealedSecrets. But as mentioned above, don't lull yourself in a false sense of security: you must assume the old version of the `SealedSecret` resouece (the one encrypted with a key you think of as dead) is still potentially around and accessible to attackers. I.e. re-encryption is not a substitute for periodically rotating your actual secrets.
```
alemorcuq, line 560:
```suggestion
It's a good idea to periodically re-encrypt your SealedSecrets. But as mentioned above, don't lull yourself in a false sense of security: you must assume the old version of the `SealedSecret` resource (the one encrypted with a key you think of as dead) is still potentially around and accessible to attackers. I.e. re-encryption is not a substitute for periodically rotating your actual secrets.
```

README.md
```diff
@@ -581,26 +581,26 @@ More details about crypto can be found [here](docs/developer/crypto.md).
 
 ## Developing
 
-More details about crypto can be found [in the Developer Guide](docs/developer/README.md).
+Developing guidelines can be found [in the Developer Guide](docs/developer/README.md).
```
alemorcuq, line 584: Good catch 😅

alemorcuq (changes requested):
Thank you for this! I've requested some minor changes, but this is great.

README.md
kiela: The text is about a situation in which a **user** could rename a SealedSecret to which the **user** has no access to and name it as a SealedSecret the **user** has access to so we do not allow anyone (users) to rename any SealedSecret.

In that case, how about:
```suggestion
Furthermore, namespaces are not the only level at which RBAC configurations can decide who can see which secret. In fact, it's possible that users can access a secret called `foo` in a given namespace but not any other secret in the same namespace. We cannot thus by default let users freely rename `SealedSecret` resources otherwise a malicious user would be able to decrypt any SealedSecret for that namespace by just renaming it to overwrite the one secret user does have access to. We use the same mechanism used to include the namespace in the encryption key to also include the secret name.
```
?

README.md
kiela: Anytime 😉

kiela:
@alvneiayu @alemorcuq @davidkarlsen @gliptak PR has already been idle for some time. Can I ask for a review?

josvazg:
> @alvneiayu @alemorcuq @davidkarlsen @gliptak PR has already been idle for some time. Can I ask for a review?

Sorry we missed this updates!
We'll get to this next week!

README.md
josvazg: Replacing `he` by `user` in this context LGTM!

josvazg (approved):
LGTM!

PR #920, opened 2022-08-22
Rebase of Refactor flag handling

josvazg:
**Description of the change**

This is basically a rebase of #440 with some extra `Config` struct to wrap everything and avoid all remaining globals.

The run method was too long and it was impossible to figure out what each parameter meant.
Furthermore a few flags were still accessed via global variables making it hard to reason about
and to test properly.

This is the first pass that just mechanically puts all the flag variables into a struct.

**Benefits**

No more global variables, easier to pass flags around and more readable and maintainable code.

### DIFF
--- a/cmd/kubeseal/main.go
+++ b/cmd/kubeseal/main.go
@@ -376,7 +376,8 @@ func seal(cfg *Config, in io.Reader, out io.Writer, codecs runtimeserializer.Cod
 	return nil
 }
 
-func validateSealedSecret(cfg *Config, in io.Reader, namespace, name string) error {
+func validateSealedSecret(cfg *Config, in io.Reader) error {
+	flags := cfg.flags
 	conf, err := cfg.clientConfig.ClientConfig()
 	if err != nil {
 		return err
@@ -385,7 +386,7 @@ func validateSealedSecret(cfg *Config, in io.Reader, namespace, name string) err
 	if err != nil {
 		return err
 	}
-	portName, err := getServicePortName(cfg.ctx, restClient, namespace, name)
+	portName, err := getServicePortName(cfg.ctx, restClient, flags.controllerNs, flags.controllerName)
 	if err != nil {
 		return err
 	}
@@ -396,10 +397,10 @@ func validateSealedSecret(cfg *Config, in io.Reader, namespace, name string) err
 	}
 
 	req := restClient.RESTClient().Post().
-		Namespace(namespace).
+		Namespace(flags.controllerNs).
 		Resource("services").
 		SubResource("proxy").
-		Name(net.JoinSchemeNamePort("http", name, portName)).
+		Name(net.JoinSchemeNamePort("http", flags.controllerName, portName)).
 		Suffix("/v1/verify")
 
 	req.Body(content)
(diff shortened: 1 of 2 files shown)


josvazg:
I think I fixed the linter issues.

Will now run the integration tests locally. Seems the unit tests missed something on `kubeseal` CLI behavior that might be broken in this PR.

cmd/kubeseal/main.go
```diff
@@ -700,28 +752,28 @@ func run(clientConfig clientcmd.ClientConfig, ctx context.Context, w io.Writer,
...
 	}
-	if len(privKeys) != 0 && isatty.IsTerminal(os.Stderr.Fd()) {
+	if len(flags.privKeys) != 0 && isatty.IsTerminal(os.Stderr.Fd()) {
 		fmt.Fprintf(os.Stderr, "warning: ignoring --recovery-private-key because unseal command not chosen with --recovery-unseal\n")
 	}
 
-	if validateSecret {
-		return validateSealedSecret(clientConfig, ctx, input, controllerNs, controllerName)
+	if flags.validateSecret {
+		return validateSealedSecret(cfg, input, flags.controllerNs, flags.controllerName)
```
alvneiayu, line 763: Looking the code, we are passing the flags as parameters. Maybe one future improvement is to use the flag struct as variable and then use the values inside each function. We can reduce the parameters in the functions. WDYT?

alvneiayu (changes requested):
First of all, thanks @josvazg for your awesome work. I included a comment and a future suggestion.

cmd/kubeseal/main.go
josvazg: Need to review this carefully.

In some places the call used the flag values as fallbacks and the explicit arguments as the overrides. Not that I like that, but it seems something for a follow up PR.

In this particular case, it seems we could remove those arguments and just pass the config (contains the flags) and the input. As I said  think that is better for a follow up PR. Specially as if we move this code around into a library we might not want to pass cfg or flags directly.

josvazg: I guess we can already fix this particular case here and worry about further refactors later.

Will fix.

cmd/kubeseal/main.go
```diff
@@ -700,28 +752,28 @@ func run(clientConfig clientcmd.ClientConfig, ctx context.Context, w io.Writer,
 		w = f
 	}
 
-	if unseal {
-		return unsealSealedSecret(w, input, scheme.Codecs, privKeys)
+	if flags.unseal {
+		return unsealSealedSecret(flags, w, input, scheme.Codecs, flags.privKeys)
```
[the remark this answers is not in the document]
alvneiayu: sounds awesome

cmd/kubeseal/main.go
josvazg: Done, I also fixed `reEncryptSealedSecret` which was NOT using overrides as well.

cmd/kubeseal/main.go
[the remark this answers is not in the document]
josvazg: Done

alvneiayu (approved):
LGTM

PR #921, opened 2022-08-23
Publish kubeseal binary in a Docker image

### LINKED ISSUE #916: Publish kubeseal binary in a Docker image
**Which component**:
`kubeseal`

We have an internal docker image that is built and published automatically via Gitlab CI and that it containing the `kubeseal` binary.
The Gitlab CI runners by default do not have internet access so we have added a static version of the kubeseal file and we have to update it manually every time there is a new release

Instead of that, we would like to have the possibility to dynamically fetch the latest version or take advantage of Renovate to upgrade the version used as soon as a new one is released  (Docker images are synched to a private registry )

Ideally, I would like to be able to grab an official docker image that contains the `kubeseal` self-contained binary and just copy it to my image

```
COPY --from=bitnami/kubeseal:0.18.1 /usr/bin/kubeseal-linux-amd64 /usr/bin/
```

I would not specifically request one docker image per dist/arch. It can be one image that contains all different combinations and we can choose which binary to grab

tewfik-ghariani:
**Description of the change**

New official docker images containing the `kubeseal` binary

**Benefits**

Users may copy the binary directly while building their own docker images

**Possible drawbacks**

Too many images, maybe?

**Applicable issues**

- fixes #916

**Additional information**

1) <strike>Question about the name of the image, should it be `kubeseal` or `sealed-secrets-kubeseal`</strike> -> `sealed-secrets-kubeseal`

2) Another question about the dockerfiles, there does not seem to be a straightforward way to have only one Dockerfile and use arguments to specify the tool ( kubeseal vs controller ) mainly due to the entrypoint, is it okay to keep both files?

3) I have just tested the build phase via `make` but not sure about the push/release process

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -11,7 +11,15 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        gotestsum: ["1.8.1"]
+        include:
+          - dockerfile: ./docker/controller.Dockerfile
+            image: docker.io/bitnami/sealed-secrets-controller
+          - dockerfile: ./docker/controller.Dockerfile
+            image: ghcr.io/bitnami-labs/sealed-secrets-controller
+          - dockerfile: ./docker/kubeseal.Dockerfile
+            image: docker.io/bitnami/sealed-secrets-kubeseal
+          - dockerfile: ./docker/kubeseal.Dockerfile
+            image: ghcr.io/bitnami-labs/sealed-secrets-kubeseal
     env:
       dockerhub_image_name: docker.io/bitnami/sealed-secrets-controller
       ghcr_image_name: ghcr.io/bitnami-labs/sealed-secrets-controller
@@ -31,7 +39,7 @@ jobs:
 
       - name: Install dependencies
         run: |
-          go install gotest.tools/gotestsum@v${{ matrix.gotestsum }}
+          go install gotest.tools/gotestsum@v1.8.1
 
       # Run tests
       - name: Tests
@@ -87,14 +95,13 @@ jobs:
         id: meta
         uses: docker/metadata-action@98669ae865ea3cffbcbaa878cf57c20bbf1c6c38
         with:
-          images: |
-            ${{ env.dockerhub_image_name }}
-            ${{ env.ghcr_image_name }}
+          images: ${{ matrix.image }}
       - name: Build and push
         id: docker_build
         uses: docker/build-push-action@v2
         with:
           context: .
+          file: ${{ matrix.dockerfile }}
           platforms: linux/amd64,linux/arm64,linux/arm
           push: true
           tags: ${{ steps.meta.outputs.tags }}
--- a/Makefile
+++ b/Makefile
@@ -14,7 +14,7 @@ CONTROLLER_GEN ?= controller-gen
 
 REGISTRY ?= docker.io
 CONTROLLER_IMAGE = $(REGISTRY)/bitnami/sealed-secrets-controller:latest
-KUBESEAL_IMAGE = $(REGISTRY)/bitnami/kubeseal:latest
+KUBESEAL_IMAGE = $(REGISTRY)/bitnami/sealed-secrets-kubeseal:latest
 INSECURE_REGISTRY = false # useful for local registry
 IMAGE_PULL_POLICY = Always
 KUBECONFIG ?= $(HOME)/.kube/config
@@ -52,7 +52,7 @@ generate: $(GO_FILES)
 	@# github.com/bitnami-labs/sealeds-secrets/...
 	@# instead of just updating ./pkg 
 	@# for that reason we generate at gentmp and then move it all to ./pkg
-	cp -r gentmp/github.com/bitnami-labs/sealed-secrets/pkg . && rm gentmp/ -rf
+	cp -r gentmp/github.com/bitnami-labs/sealed-secrets/pkg . && rm -rf gentmp/
 
 manifests:
 	$(CONTROLLER_GEN) crd paths="./pkg/apis/..."  output:crd:artifacts:config=helm/sealed-secrets/crds/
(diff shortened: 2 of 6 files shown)


.github/workflows/release.yaml
```diff
@@ -10,8 +10,10 @@ jobs:
   build:
     runs-on: ubuntu-latest
     env:
-      dockerhub_image_name: docker.io/bitnami/sealed-secrets-controller
-      ghcr_image_name: ghcr.io/bitnami-labs/sealed-secrets-controller
+      dockerhub: docker.io/bitnami
+      ghcr: ghcr.io/bitnami-labs
+      controller_image_name: sealed-secrets-controller
+      kubeseal_image_name: kubeseal
```
alvneiayu, line 16: Please, change it to sealed-secrets-kubeseal. To follow the same format as the controller.

Makefile
```diff
@@ -12,6 +12,7 @@ DOCKER = docker
 GINKGO = ginkgo -p
 
 CONTROLLER_IMAGE = docker.io/bitnami/sealed-secrets-controller:latest
+KUBESEAL_IMAGE = docker.io/bitnami/kubeseal:latest
```
alvneiayu, line 15: DITTO

alvneiayu (changes requested):
hi @tewfik-ghariani 

First of all, thanks a lot for sharing your solution. Let me test the patch. I think that we are going to have problem with the release process. Let me verify it and I am going to come back with news or possible solution if the release process fails. 

Thanks again

josvazg:
> Another question about the dockerfiles, there does not seem to be a straightforward way to have only one Dockerfile and use arguments to specify the tool ( kubeseal vs controller ) mainly due to the entrypoint, is it okay to keep both files?

I would say we want 2 separate Dockerfiles. Each have different purposes and run on different environments. You do not usually need the CLI inside the controller or container but outside.

IMHO the main `Dockerfile` is the controller one, as it **must** be a container image to be deployed in Kubernetes. The CLI container is yet another packaging solution that might be convenient on certain use cases.

.github/workflows/release.yaml
tewfik-ghariani: Done, thanks ^^

Makefile
tewfik-ghariani: Done as well

tewfik-ghariani:
👋 Hello
Are there any updates about this please?

alvneiayu:
hi @tewfik-ghariani 

I am working on it. This patch will break the release process. I am going to share a solution to include it in your patch.

Thanks a lot and sorry for the delay.

Álvaro

tewfik-ghariani:
@alvneiayu No worries thanks a lot for the response ^^

.github/workflows/release.yaml
```diff
@@ -13,8 +13,10 @@ jobs:
       matrix:
```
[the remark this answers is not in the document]
alvneiayu: After the change, the code will look like:

Screenshot from 2022-09-09 14-03-08

.github/workflows/release.yaml
```diff
@@ -88,8 +90,10 @@ jobs:
...
           images: |
-            ${{ env.dockerhub_image_name }}
-            ${{ env.ghcr_image_name }}
+            ${{ env.dockerhub }}/${{ env.controller_image_name }}
+            ${{ env.dockerhub }}/${{ env.kubeseal_image_name }}
+            ${{ env.ghcr }}/${{ env.controller_image_name }}
+            ${{ env.ghcr }}/${{ env.kubeseal_image_name }}
       - name: Build and push
         id: docker_build
         uses: docker/build-push-action@v2
```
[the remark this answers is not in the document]
alvneiayu: After the change, the code will look like:
Screenshot from 2022-09-09 13-58-15

alvneiayu (changes requested):
Here we go, I think that with my suggestions the release will work as expected.

Again, thanks a lot for your PR and your great work.

Álvaro

.github/workflows/release.yaml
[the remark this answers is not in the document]
tewfik-ghariani: Just to double check, is the "file:" line supposed to be **before** the "context:" or **after** ?

.github/workflows/release.yaml
```diff
@@ -41,7 +46,7 @@ jobs:
       - name: K8s manifests
         run: |
           export PATH=~/bin:$PATH
-          make CONTROLLER_IMAGE=${{ env.dockerhub_image_name }}:${{ github.ref_name }} controller.yaml controller-norbac.yaml
+          make CONTROLLER_IMAGE=${{ env.dockerhub }}/${{ env.controller_image_name }}:${{ github.ref_name }} controller.yaml controller-norbac.yaml
```
tewfik-ghariani, line 49: We will need to adapt this value

.github/workflows/release.yaml
```diff
@@ -87,13 +92,12 @@ jobs:
...
-            ${{ env.ghcr_image_name }}
+          images: ${{ matrix.image }}
       - name: Build and push
         id: docker_build
         uses: docker/build-push-action@v2
         with:
+          file: ${{ matrix.dockerfile }}
           context: .
           platforms: linux/amd64,linux/arm64,linux/arm
           push: true
```
tewfik-ghariani, line 103: Same for the ghcr signing

```
            COSIGN_REPOSITORY: ${{ env.ghcr_image_name }}/signs
```

.github/workflows/release.yaml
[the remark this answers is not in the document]
alvneiayu: before or after, it will work ^^ I included it after the "context:"

.github/workflows/release.yaml
alvneiayu: good catch. Totally agree

.github/workflows/release.yaml
alvneiayu: good catch. Totally agree

tewfik-ghariani: How do you suggest to remediate that? Should I keep the env variable? We would not want to loop through the matrix again just for the ghrc/controller right?

alvneiayu: keep the env variable, yes. I tried to use the env variable in the matrix but I reproduced errors. So keep that matrix as I shared with you and keep the env variables , please. Thanks again

tewfik-ghariani: ok done, thanks a lot for the review!

alvneiayu (approved):
LGTM

alvneiayu (approved):
LGTM

PR #924, opened 2022-08-24
Simplify controller testing

josvazg:
**Description of the change**

Add a `controller-tests` Makefile rule that will automatically setup everything to run all tests required for validating the controller. 

This includes the integration tests and requires that a given Kubernetes configuration context, environment variable `K8S_CONTEXT`, is available to deploy all controller artifacts and run the integration tests.

There are also environment variables to:
- Customize the `REGISTRY` to which the controller images compiled will we pushed to and used from the tested code.
- Specify the `OS` and architecture (`ARCH`) of the target test cluster.

**Benefits**

I noticed I was failing to properly run the numerous steps to test the controller on my local `kind` cluster. For instance, I was not setting the registry properly, so it was probably testing a stale older image of the controller.

This new single step avoids an error prone sequence of steps. Not only it makes things easier, it also ensures you are running the test against the expected cluster context.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -11,7 +11,9 @@ KUBECFG = kubecfg
 DOCKER = docker
 GINKGO = ginkgo -p
 
-CONTROLLER_IMAGE = docker.io/bitnami/sealed-secrets-controller:latest
+REGISTRY ?= docker.io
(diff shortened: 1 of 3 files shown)


alvneiayu (approved):
LGTM

PR #933, opened 2022-08-25
Fix Dockerfile to align with new GoReleaser Path format

alvneiayu:
**Description of the change**

GoReleaser has changed the path format to generate the binaries. Modifying the Dockerfile, we are changing the copy path to generate the images.

**Benefits**

Fix the Release process.


josvazg (approved):
LGTM

Maybe explain in the description that the issue comes exactly because `GoReleaser` is adding the architecture version suffix to the arch bit? (and maybe with an example)

Issue #934, opened 2022-08-29
Raw mode does not work on windows: open /dev/stdin: The system cannot find the path specified.


matbarker:
**Which component**:
Kubeseal v0.18.2

**Describe the bug**

Expected to be able to use --raw option using the example command from the readme but got the error:
`error: open /dev/stdin: The system cannot find the path specified.`

**To Reproduce**

In a Powershell or Command window:
Run `echo -n "mypassword" | kubeseal --raw --from-file=/dev/stdin --namespace default --name my-test-secret --controller-name=sealed-secrets       --controller-namespace=kube-system`

outputs: `error: open /dev/stdin: The system cannot find the path specified.`

The command succeeds in a windows bash shell, but the encrypted password ends up being (base 64 decoded): 
`!<symlink>/�p�r�o�c�/�s�e�l�f�/�f�d�/�0���` instead of `mypassword`

**Expected behavior**
Expected output of an encoded version of 'mypassword' string.

**Version of Kubernetes**:
Docker desktop kubernetes v1.24.2

- Output of `kubectl version`:

```
WARNING: This version information is deprecated and will be replaced with the output from kubectl version --short.  Use --output=yaml|json to get the full version.
Client Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.2", GitCommit:"f66044f4361b9f1f96f0053dd46cb7dce5e990a8", GitTreeState:"clean", BuildDate:"2022-06-15T14:22:29Z", GoVersion:"go1.18.3", Compiler:"gc", Platform:"windows/amd64"}
Kustomize Version: v4.5.4
Server Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.0", GitCommit:"4ce5a8954017644c5420bae81d72b09b735c21f0", GitTreeState:"clean", BuildDate:"2022-05-03T13:38:19Z", GoVersion:"go1.18.1", Compiler:"gc", Platform:"linux/amd64"}
```

**Additional context**

josvazg:
When running a command such as:
```shell
$ echo -n "mypassword" | kubeseal --raw --from-file=/dev/stdin --name my-test-secret ...
```

The tool is encrypting the string "mypassword" using the key from the cluster. The result from standard output is a _base64_ encoded **encrypted value**. This result is different each time you run the command and should not match "mypassword", but instead the Kubernetes cluster should be able to use its key to decrypt encrypted string back into "mypassword".

[...]

PR #935, opened 2022-08-29
Remove /dev/stdin mentions in Readme

josvazg:
**Description of the change**

As per https://github.com/bitnami-labs/sealed-secrets/issues/934 we noticed users might run into issues when testing the README raw commands that mention `/dev/stdin` and will not work in Windows.

As that flag is not required and does not work in Windows seems better to remove it from the examples.

**Benefits**

Simpler documentation, less confusion and friction for Windows users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -434,14 +434,14 @@ See [Scopes](#scopes)
 `strict` scope (default):
 
 ```console
-$ echo -n foo | kubeseal --raw --from-file=/dev/stdin --namespace bar --name mysecret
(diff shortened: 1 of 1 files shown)


mkmik (approved):

alvneiayu (approved):
LGTM

josvazg:
As per #934 we noticed users might run into issues when testing the README raw commands that mention /dev/stdin and will not work in Windows.

As that flag is not required and does not work in Windows seems better to remove it from the examples.

PR #937, opened 2022-08-30
Fix missing push and readme on new tests rules

josvazg:
**Description of the change**

The push was missing and some instructions also regarding the push rule were inaccurate, missing the `K8S_CONTEXT` env var.


alvneiayu (approved):
LGTM

PR #939, opened 2022-08-30
Make the kubeseal CLI have a skinny main

josvazg:
**Description of the change**

The CLI had way too much code within the main package.

This change moves most of the kubeseal CLI code to a package "pkg/kubeseal", leaving a simple "main" which is only concerned with CLI flags and other command line perks.

Note that the changes are minimal, and only functions affected by the split of code between main and the package are really affected. The rest is code copied to another package.

Other changes included:
- Flag fields needed to become public, and this are upper case now.
- There is a new NewConfig() function now.
- VERSION handling is handled by main completely and is out of the Flags struct now.

**Benefits**

Code in the new pkg is accesible from other packages and can be further refactored and improved.

**Possible drawbacks**

This is still a mostly mechanical code move, it does **not yet address making the `kubeseal` code look and feel like a proper library** to be consumed by code outside of a CLI context. This was left out of scope intentionally to make progress on smaller incremental steps.

**Additional information**

Once this code is merged, the intent going forward is to start refactoring things out of the big `Run` function in small steps. The expectation of the final result is that a smaller (and skinny) version of such "run" might go back to the `main` code but it will only decide what operations or behaviors of `kubeseal` will be invoked by each CLI execution. Many of those behaviors or operations will become new functions within `kubeseal`. Hopefully we can get rid of using the `Flags` struct within `kubeseal`.


josvazg:
Tip: The diff might be more confusing that actually looking at the resulted `main.go` and `main_test.go` now to have a sense of what has remained there. The "pkg/kubeseal" code is mostly copy& paste from main except for those bits that could not be moved.

alvneiayu (dismissed):
LGTM

josvazg:
I found a weird bug on the refactor and it ended up being because `goimports` had changed the importes in the moved code and chose the wrong import for `scheme`, choosing the local package over a dependency.

In order to detect other instances of goimports making wrong choices, I did a comparison of old and new imports. I expect that new imports on the `main` AND `kubeseal` be the same as the old `main`. Same for tests, new test files combined should import the same things as the old `main_test` code did.

```shell
$ cat /tmp/imports-new-main.txt /tmp/imports-new-kubeseal.txt | sort -u | grep -v '^$' > /tmp/new-code-imports
$ sort -u < /tmp/imports-old-main.txt |grep -v '^$' > /tmp/expected-code-imports
$ diff /tmp/expected-code-imports /tmp/new-code-imports 
13a14
> 	"github.com/bitnami-labs/sealed-secrets/pkg/kubeseal"
28c29
< 	"k8s.io/klog/v2"
---
> 	"k8s.io/klog"
```
This happened because `goimports` knew it needed to import a `klog` package ,but it did not know it had to be version 2.

Note the `github.com/bitnami-labs/sealed-secrets/pkg/kubeseal` is expected on the new main and not the old, that is the code we moved away.

Test code showed no differences, so we are fine there:
```shell
$ cat /tmp/imports-new-main_test.txt /tmp/imports-new-kubeseal_test.txt | sort -u | grep -v '^$' > /tmp/new-test-imports
$ sort -u < /tmp/imports-old-main_test.txt |grep -v '^$' > /tmp/expected-test-imports
$ diff /tmp/expected-test-imports /tmp/new-test-imports 
$ 
```

Fix for `klog/v2` coming up...

pkg/kubeseal/kubeseal.go
```diff
@@ -0,0 +1,743 @@
...
+	}
+	sec, err := ss.Unseal(codecs, privKeys)
+	if err != nil {
+		return err
+	}
+
+	return resourceOutput(w, flags, codecs, v1.SchemeGroupVersion, sec)
+}
+
+func Run(w io.Writer, cfg *Config) (err error) {
```
alemorcuq, line 617: I have my doubts about this function being part of the `kubeseal` pkg. IMO, the `kubeseal` pkg should provide an API that can later be used, for example, to create a CLI or a HTTP API for `kubeseal`. This function is basically running the CLI, but the `kubeseal` pkg shouldn't know anything about a CLI. It should be independent, agnostic.

josvazg: You are right, this `Run` should not be here, in fact it should not exist as is. But this PR is only a first step to move most of the `kubeseal` specific code away from the `main`, so it can be easily reused as needed.

In this shape right now you could do all the CLI does from as a library. It is not a proper library API, true, but it can be used. Whereas in the `main` it could not be reused at all from other code, only as a binary.

Note that what remains in `main` now is trully CLI dependent, whereas in `kubeseal` we have a `Flags` struct that might invoke a CLI meaning but other than that is basically not CLI aware, it does not parse flags, for instance.

Going forward, the idea is to keep refactoring and dismembering this Run into pieces. Some code will go back to `main` and probably we will get rid of the `Flags` type. Ideally, each of the behaviors of the CLI will map to functions in `kubeseal` with specific parameters, instead of a flags bag.

The plan is for smaller incremental changes, to avoid big changes that are difficult to review and may insert bugs without notice. In fact, I expect that dismembering `Run` will take more than a single PR.

alemorcuq (approved):

alvneiayu (approved):
LGTM

PR #940, opened 2022-08-31
Remove magic init funcs and global flags

josvazg:
**Description of the change**

Like with the CLI, the controller main code relies too much on global flags that get used more deeply in the code than needed. Also the use of `func init()` "magic" initializations would make moving code out of the main package problematic. So:
- All flags but the `printVersion` one are packed into a single `Flags` struct, similarly to what was done with the CLI recently. That removes most global variables in main right now.
- Both `func init()` functions are replaced by explicit calls with explicit parameters.

**Benefits**

The code is now ready to be easily moved to a package while keeping the command line concerns in main.

### DIFF
--- a/cmd/controller/keyregistry.go
+++ b/cmd/controller/keyregistry.go
@@ -47,8 +47,8 @@ func NewKeyRegistry(client kubernetes.Interface, namespace, keyPrefix, keyLabel 
 	}
 }
 
-func (kr *KeyRegistry) generateKey(ctx context.Context) (string, error) {
-	key, cert, err := generatePrivateKeyAndCert(kr.keysize)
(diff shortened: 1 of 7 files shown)


josvazg:
Note that one of the most noise changes here is that `generateKey` and `initKeyRenewal` were relying on glabal variables set by flags to avoid having to pass 2 more parameters: `validFor Duration` and `cn string` 

Internally, those end up calling `generatePrivateKeyAndCert` also not passing those extra parameters, and that file was using the globals directly.

I was thinking about some cert generation config type, but seemed just easier to propagate the parameters as they were.

alvneiayu (approved):
LGTM

alemorcuq (approved):

Issue #505, opened 2021-01-21
Broken SealedSecret causes failure in processing


bbockelm:
We had an individual accidentally upload a broken sealed secret to the cluster (it's not quite obvious what was wrong with it); it appears that, once this secret was considered by the sealed-secrets operator the operator failed and stopped processing other items.

Here's the error message that was printed about once a second:

```
E0121 18:14:24.255540       1 reflector.go:123] github.com/bitnami-labs/sealed-secrets/cmd/controller/controller.go:164: \
Failed to list *v1alpha1.SealedSecret: v1alpha1.SealedSecretList.Items: \
[]v1alpha1.SealedSecret: v1alpha1.SealedSecret.Spec: v1alpha1.SealedSecretSpec.EncryptedData: \
ReadString: expects " or n, but found {, error found in #10 byte of \
  ...|etadata":{"name":"ju|..., bigger context ...|ec":{"encryptedData":{"kind":"Secret","metadata":{"name":"jupyter-hub-cookies"},"proxy.token":"AgBa1|...
```

(line breaks added for readability).  Note the original YAML version of the object looked fine but something in the JSON representation went awry it seems.

Once this particular object was fixed the operator continued on to process everything successfully.

bbockelm:
It seems to be reproducible if you just feed it any garbage data in the encryptedData.  From a separate test:
```
spec:
  encryptedData:
    client_id: xxx
    client_secret: xxx
    metadata:
      name: test-object
      namespace: default
```
(literally use `xxx` as the encrypted data)

This is fairly frustrating as a failure in any one namespace results in the operator from working in any namespace.

smercier74:
We have the same issue

windmark:
I just experienced the same issue with a single failed SealedSecret causing a blockade of all other secrets. Can this be mitigated somehow?

stefancoetzee-xneelo:
Same here. Feeding it clear text, rather than encrypted data causes the controller to get stuck trying to decrypt the same sealed secret over and over, rather than marking it as failed and moving on with the rest of the sealed secrets.

burkhat:
We got last week the same issue. Any updates here?

denniskaulbars:
Same here. We need a fix for this asap :-/

bbockelm:
@alvneiayu - I see you handled the latest release for 0.17.3.  Do you happen to know if there's any interest in fixing this issue?

[...]

PR #941, opened 2022-09-01
Add CRD schema and fix JSON errors

### LINKED ISSUE #82: Add/restore openapi schema for CRD
See discussion during PR #73.  We had to remove the schema due to upstream k8s limitations that left us with no option to keep the schema _and_ allow an object with arbitrary keys.  This issue is to track finding a workaround and/or upstream fix that allows us to have key/value objects _and_ a CRD schema.

snorwin:
The lack of proper schema validation can lead to errors when using libraries that rely on K8s API responses (JSON) to match Go structs. In particular, if there are errors in the unmarshaling  a  SealedSecret resource, the controller will stop watching and processing all the resources of that kind. To avoid such problems, an OpenAPI scheme was introduced wherever possible. 

**Description of the change**
- OpenAPI schema generation using controller-gen
- preserve unknown fields only for `.spec.encryptedData`
- ignore errors when unmarshaling `.spec.encryptedData` as the structure isn't ensured by any schema

**Benefits**
- controller still valid processes sealed secrets even if there are any invalid ones
- OpenAPI schema guides the user to create SealedSecret and prevents structural errors

**Possible drawbacks**
None.

**Applicable issues**
Fixes #82 #505


mkmik:
awesome I was meaning to do this since forever but I was lazy; thank you!

mkmik:
Unfortunately I can no longer approve the workflows :-(

snorwin:
As we now enable validation maybe it would make sense to set `Data`  in `.spec.template` as optional too, what are you thinking?
```
	// Keys that should be templated using decrypted data
	// +optional
	Data map[string]string json:"data"
```

mkmik:
yeah making that optional seems reasonable

josvazg:
Hi @snorwin, thanks for this fix

I will review on our side and we can hopefully have it merged soon

josvazg (approved):
LGTM!

Thanks a lot for this contribution!

stefancoetzee-xneelo:
Shouldn't we have updated the patch version as we're touching the schema?

PR #943, opened 2022-09-02
Extend controller-tests support and docs to minikube

josvazg:
**Description of the change**

When making tests with `minikube` locally in the OSS project I noticed the new `make controller-tests` was not working as expected: In the case of `minikube` we need to skip the docker push that is required for kind.

I also noticed a race condition with the apply manifests rule as they did not wait for the deployment to be stable, so tests could be run too soon.

The documentation of this can be improved by adding more details on the `kind` vs `minikube` setups and the parameters they need to use.

**Benefits**

More reliable `make controller-test` supporting `minikube` and better documentation on both recommended setups: `kind` & `minikube`.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -137,10 +137,15 @@ check-k8s:
 
 push-controller: clean check-k8s controller.image.$(OS)-$(ARCH)
 	docker tag $(CONTROLLER_IMAGE)-$(OS)-$(ARCH) $(CONTROLLER_IMAGE)
+ifeq ($(REGISTRY),docker.io)
+  echo "Skip push: docker.io registy means minikube"
+else
 	docker push $(CONTROLLER_IMAGE)
+endif
 
 apply-controller-manifests: clean check-k8s controller.yaml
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -137,10 +137,15 @@ check-k8s:
 
 push-controller: clean check-k8s controller.image.$(OS)-$(ARCH)
 	docker tag $(CONTROLLER_IMAGE)-$(OS)-$(ARCH) $(CONTROLLER_IMAGE)
+ifeq ($(REGISTRY),docker.io)
+  echo "Skip push: docker.io registy means minikube"
```
alvneiayu, line 141: TYPO here, registry

alvneiayu (approved):
LGTM, Fix the typo that I included in my comments.

Makefile
josvazg: Right!
```suggestion
  echo "Skip push: docker.io registry means minikube"
```

alvneiayu (approved):
LGTM

PR #945, opened 2022-09-05
Allow additional volumes deployment

suetema:
**Description of the change**

This PR makes within the helm chart the following changes:
 - Allow for extra volumes to be added to the Sealed Secrets controller Deployment.
 - Allow for extra volumeMounts to be added to the Sealed Secrets controller container.
 - Possibility to add annotations only to the service account.

**Benefits**

This would allow Sealed secrets to pick up certificates/secrets installed by the secret store csi driver(https://secrets-store-csi-driver.sigs.k8s.io)

I ran into the issue when combining sealed secrets with ASCP(https://docs.aws.amazon.com/secretsmanager/latest/userguide/integrating_csi_driver.html). I would bootstrap the EKS cluster with an instance of sealed secrets and to get a previously defined certificate as a secret in the cluster, I use the secret store csi driver with the AWS provider. CSI driver has option to reflect a secret volume as a k8s secret object, only after the volume has been mounted by a pod.

Without this change we would have to deploy an additional pod to the same namespace mounting the secret volume. so that then the controller can pick the secret up.

edit: forgot to mention why the SA change is needed.
In EKS pods get linked to an IAM role using an Annotation in the following form: `eks.amazonaws.com/role-arn: arn:aws:iam::XXXXXXXXXXXX:role/eksctl-sandbox-addon-iamserviceaccount-defau-xxx-xxx` .
I don't want the annotation to be added anywhere else, so commonAnnotations is not an option.

### DIFF
--- a/helm/sealed-secrets/Chart.yaml
+++ b/helm/sealed-secrets/Chart.yaml
@@ -14,4 +14,4 @@ maintainers:
     url: https://github.com/bitnami-labs/sealed-secrets
 name: sealed-secrets
 type: application
-version: 2.6.1
+version: 2.6.2
--- a/helm/sealed-secrets/README.md
+++ b/helm/sealed-secrets/README.md
@@ -96,6 +96,8 @@ The command removes all the Kubernetes components associated with the chart and 
 | `rateLimit`                                       | Number of allowed sustained request per second for verify endpoint                   | `""`                                |
 | `rateLimitBurst`                                  | Number of requests allowed to exceed the rate limit per second for verify endpoint   | `""`                                |
 | `additionalNamespaces`                            | List of namespaces used to manage the Sealed Secrets                                 | `[]`                                |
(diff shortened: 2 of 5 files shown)


alvneiayu:
hi @suetema 

Could you include the new variables inside the helm chart README, please?

Thanks a lot

Álvaro

josvazg (changes requested):
Marking as request changes waiting on the readme bits

suetema (commented):
Entries for the new parameters have been added to the helm README

alvneiayu (approved):
LGTM

josvazg (approved):
LGTM

Thanks for this new feature!

suetema:
I now see that the chartversion had already increased to 2.6.2 prior to my pr.
How should I handle this in the future? Should I be the one bumping the chartversion or is this done by the maintainers?

alvneiayu:
No problem, it was my fault. I expected conflicts with our branch. I am going to fix it.

suetema:
Super, thank you!

PR #947, opened 2022-09-07
Make the controller main skinny

josvazg:
**Description of the change**

The Controller has way too much code within the main package.

- Things in main cannot be referred to or used from other code.
- Refactors and code separation are harder when busines logic is in the main package.

The first step to allow easy modularity is to move most code outside of main, where we can further slice it and separate it.

**Benefits**

Code moves to a package where it is easier to further refactor.


alvneiayu (approved):
LGTM

agarcia-oss (approved):
Great refactor, @josvazg !

PR #959, opened 2022-09-15
Make imagePullPolicy default to Always only if using :latest image

### LINKED ISSUE #958: imagePullPolicy is Always in release controller.yaml
**Which component**:

controller.yaml v0.18 (and older)

**Describe the bug**

The release `controller.yaml` includes

```yaml
image: "docker.io/bitnami/sealed-secrets-controller:v0.18.0"
imagePullPolicy: "Always"
```

"Always" here is surprising, and I think, incorrect.  In particular, this means that an in-place pod restart requires that the docker registry is available, even though the correct image is already available locally.

**Expected behavior**

```yaml
imagePullPolicy: "IfNotPresent"
```

(or absent, which would have the same default value)

**Additional context**

I believe the helm chart handles this correctly, and the issue is just the value set in `Makefile` and used when generating a release manifest.

mkmik:
Fixes #958

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -16,7 +16,7 @@ REGISTRY ?= docker.io
 CONTROLLER_IMAGE = $(REGISTRY)/bitnami/sealed-secrets-controller:latest
 KUBESEAL_IMAGE = $(REGISTRY)/bitnami/sealed-secrets-kubeseal:latest
 INSECURE_REGISTRY = false # useful for local registry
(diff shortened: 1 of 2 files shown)


alemorcuq (approved):
LGTM!

alvneiayu (approved):
LGTM

josvazg:
Thanks @mkmik !

PR #960, opened 2022-09-15
Upgrade Go versions and dependencies

alemorcuq:
**Description of the change**

- Upgrade Go to the latest versions available: `1.18.6` and `1.19.1`
- Upgrade dependencies
- Fix [CVE-2022-27664](https://avd.aquasec.com/nvd/cve-2022-27664)

New scan:

```
Total: 0 (UNKNOWN: 0, LOW: 0, MEDIUM: 0, HIGH: 0, CRITICAL: 0)
```


.github/workflows/ci.yml
```diff
@@ -15,7 +15,7 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        go: ["1.19", "1.18.5"]
+        go: ["1.19.1", "1.18.6"]
```
josvazg, line 18: Can't we use variables for all the Gitlab yamls so we do not need to update these in so many places?
We risk being inconsistent

I assume we can at least do it per file, right?

agarcia-oss: Agree with that, @josvazg but it looks unrelated to the current issue..

agarcia-oss (approved):
LGTM

.github/workflows/ci.yml
alemorcuq: I agree with @agarcia-oss. Let's tackle that as a separate issue.

josvazg: Yes, separate issue makes sense.

PR #963, opened 2022-09-15
Support to recreate a deleted secret generated by the controller

### LINKED ISSUE #224: React to deleted secrets by re-creating them
Currently when a secret is deleted the controller doesn't do anything until something triggers a new run, which can happen either because there is a change in the sealed secret input or because the controller restarts.

K8s users are used to controllers recreating dependent objects, or at least the deployment->pod relationship sets a precedent.

I think it makes sense for sealed-secrets to re-create deleted secrets because it would do it eventually anyway. If we don't re-create the secret immediately, users might think they deleted the secret only to have it reappear at a random time in the future (e.g. when the controller reboots). This would be a surprising behavior.

A more practical use case is when users delete the secret because something goes wrong (e.g. they change the secret type and the sealed secret controller can perform such an update) and they want to cleanup; at least one such user has reported that the current sealed-secrets controller behavior (to not re-create the secret) is surprising.

### LINKED ISSUE #952: deleted secrets not being recreated
**Which component**:
The name (and version) of the affected component (controller or kubeseal)

bitnami/sealed-secrets-controller:v0.12.1

**Describe the bug**

I wanted to recreate a secret I accidentally deleted from a file. Kubectl says `sealedsecret.bitnami.com/mysecret unchanged` but I can't find the secret in any namespace.

**To Reproduce**

1. delete secret --> kubectl -n myns delete secret mysecret
2. recreate secret from file --> kubectl -n myns apply -f mysecret.yaml
3. output --> sealedsecret.bitnami.com/mysecret unchanged
4. list secrets in namespace --> kubectl -n myns get secrets
5. mysecret is missing

NOTE:
```
cat mysecret.yaml
kind: SealedSecret
metadata:
  annotations:
    sealedsecrets.bitnami.com/cluster-wide: "true"
  creationTimestamp: null
  name: munge-secret
  namespace: myns
spec:
  encryptedData:
    munge.key: -- REDACTED --
  template:
    metadata:
      annotations:
        sealedsecrets.bitnami.com/cluster-wide: "true"
      creationTimestamp: null
      name: munge-secret
      namespace: myns

```

**Expected behavior**
I would expect to see munge-secret secret listed after creation

**Version of Kubernetes**:

```
Client Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.13", GitCommit:"2444b3347a2c45eb965b182fb836e1f51dc61b70", GitTreeState:"clean", BuildDate:"2021-11-17T13:05:33Z", GoVersion:"go1.15.15", Compiler:"gc", Platform:"linux/amd64"}
Server Version: version.Info{Major:"1", Minor:"20", GitVersion:"v1.20.13", GitCommit:"2444b3347a2c45eb965b182fb836e1f51dc61b70", GitTreeState:"clean", BuildDate:"2021-11-17T13:00:29Z", GoVersion:"go1.15.15", Compiler:"gc", Platform:"linux/amd64"}
```

**Additional context**

I don't understand why `sealedsecret.bitnami.com/mysecret unchanged` if the secret has not been created.

alvneiayu:
**Description of the change**

Recreate secrets generated by the Sealed Secrets when they were deleted.

**Benefits**

Secrets generated by the Sealed Secrets will always be existed

**Applicable issues**

- fixes #224
- fixes #952

### DIFF
--- a/integration/controller_test.go
+++ b/integration/controller_test.go
@@ -276,7 +276,7 @@ var _ = Describe("create", func() {
 				err:= c.Secrets(ns).Delete(ctx, secretName, metav1.DeleteOptions{})
 				Expect(err).NotTo(HaveOccurred())
 			})
-			It("should the Secret exists", func() {
+			It("should recreate the secret", func() {
 				expected := map[string][]byte{
 					"foo": []byte("bar"),
 				}
--- a/integration/controller_test.go
+++ b/integration/controller_test.go
@@ -267,6 +267,28 @@ var _ = Describe("create", func() {
 				}, Timeout, PollingInterval).Should(WithTransform(getFirstOwnerName, Equal(ss.GetName())))
 			})
 		})
+
+		Context("Secret recreation", func() {
+			BeforeEach(func() {
+				c.Secrets(ns).Delete(ctx, secretName, metav1.DeleteOptions{})
+			})
+			It("should the Secret exists", func() {
+				expected := map[string][]byte{
+					"foo": []byte("bar"),
+				}
+				Eventually(func() (*v1.Secret, error) {
+					return c.Secrets(ns).Get(ctx, secretName, metav1.GetOptions{})
+				}, Timeout, PollingInterval).Should(WithTransform(getData, Equal(expected)))
+				Eventually(func() (*v1.EventList, error) {
+					return c.Events(ns).Search(scheme.Scheme, ss)
+				}, Timeout, PollingInterval).Should(
+					containEventWithReason(Equal("Unsealed")),
+				)
+				Eventually(func() (*v1.Secret, error) {
+					return c.Secrets(ns).Get(ctx, secretName, metav1.GetOptions{})
+				}, Timeout, PollingInterval).Should(WithTransform(getFirstOwnerName, Equal(ss.GetName())))
+			})
+		})
 	})
 
 	Describe("Same name, wrong key", func() {
--- a/integration/controller_test.go
+++ b/integration/controller_test.go
@@ -267,8 +267,10 @@ var _ = Describe("create", func() {
 				}, Timeout, PollingInterval).Should(WithTransform(getFirstOwnerName, Equal(ss.GetName())))
 			})
 		})
+	})
 
-		Context("With secret being deleted", func() {
+	Describe("Secret Recreation", func() {
+		Context("With owned secret", func() {
 			JustBeforeEach(func() {
 				Eventually(func() (*v1.Secret, error) {
 					return c.Secrets(ns).Get(ctx, secretName, metav1.GetOptions{})
@@ -291,6 +293,53 @@ var _ = Describe("create", func() {
 				Eventually(func() (*v1.Secret, error) {
 					return c.Secrets(ns).Get(ctx, secretName, metav1.GetOptions{})
 				}, Timeout, PollingInterval).Should(WithTransform(getFirstOwnerName, Equal(ss.GetName())))
+			})
+		})
+
+		Context("With unowned secret with managed annotation", func() {
+			BeforeEach(func() {
+				s.Annotations = map[string]string{
+					ssv1alpha1.SealedSecretManagedAnnotation: "true",
+				}
+				c.Secrets(ns).Create(ctx, s, metav1.CreateOptions{})
+			})
+			JustBeforeEach(func() {
+				err:= c.Secrets(ns).Delete(ctx, secretName, metav1.DeleteOptions{})
+				Expect(err).NotTo(HaveOccurred())
+			})
+			It("should recreate the secret", func() {
+				expected := map[string][]byte{
+					"foo": []byte("bar"),
+				}
+				Eventually(func() (*v1.Secret, error) {
+					return c.Secrets(ns).Get(ctx, secretName, metav1.GetOptions{})
+				}, Timeout, PollingInterval).Should(WithTransform(getData, Equal(expected)))
+				Eventually(func() (*v1.EventList, error) {
+					return c.Events(ns).Search(scheme.Scheme, ss)
+				}, Timeout, PollingInterval).Should(
+					containEventWithReason(Equal("Unsealed")),
+				)
+				Eventually(func() (*v1.Secret, error) {
+					return c.Secrets(ns).Get(ctx, secretName, metav1.GetOptions{})
+				}, Timeout, PollingInterval).Should(WithTransform(getFirstOwnerName, Equal(ss.GetName())))
+			})
+		})
+
+		Context("With unowned secret without managed annotation", func() {
+			BeforeEach(func() {
+				s.Annotations = map[string]string{
+				}
+				c.Secrets(ns).Create(ctx, s, metav1.CreateOptions{})
+			})
+			JustBeforeEach(func() {
+				err:= c.Secrets(ns).Delete(ctx, secretName, metav1.DeleteOptions{})
+				Expect(err).NotTo(HaveOccurred())
+			})
+			It("should not recreate the secret", func() {
+				Consistently(func() error {
+					_, err := c.Secrets(ns).Get(ctx, secretName, metav1.GetOptions{})
+					return err
+				}).Should(WithTransform(errors.IsNotFound, Equal(true)))
 			})
 		})
 	})
--- a/integration/controller_test.go
+++ b/integration/controller_test.go
@@ -269,8 +269,12 @@ var _ = Describe("create", func() {
 		})
 
 		Context("Secret recreation", func() {
-			BeforeEach(func() {
-				c.Secrets(ns).Delete(ctx, secretName, metav1.DeleteOptions{})
+			JustBeforeEach(func() {
+				Eventually(func() (*v1.Secret, error) {
+					return c.Secrets(ns).Get(ctx, secretName, metav1.GetOptions{})
+				}, Timeout, PollingInterval).Should(WithTransform(getFirstOwnerName, Equal(ss.GetName())))
+				err:= c.Secrets(ns).Delete(ctx, secretName, metav1.DeleteOptions{})
+				Expect(err).NotTo(HaveOccurred())
 			})
 			It("should the Secret exists", func() {
 				expected := map[string][]byte{
--- a/integration/controller_test.go
+++ b/integration/controller_test.go
@@ -268,7 +268,7 @@ var _ = Describe("create", func() {
 			})
 		})
 
-		Context("Secret recreation", func() {
+		Context("With secret being deleted", func() {
 			JustBeforeEach(func() {
 				Eventually(func() (*v1.Secret, error) {
 					return c.Secrets(ns).Get(ctx, secretName, metav1.GetOptions{})
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -56,8 +56,8 @@ const (
 // Controller implements the main sealed-secrets-controller loop.
 type Controller struct {
 	queue       workqueue.RateLimitingInterface
-	informer    cache.SharedIndexInformer
-	infsecret   cache.SharedIndexInformer
+	ssInformer    cache.SharedIndexInformer
+	sInformer   cache.SharedIndexInformer
 	sclient     v1.SecretsGetter
 	ssclient    ssv1alpha1client.SealedSecretsGetter
 	recorder    record.EventRecorder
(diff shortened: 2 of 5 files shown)


josvazg (dismissed):

pkg/controller/controller.go
```diff
@@ -102,8 +105,28 @@ func NewController(clientset kubernetes.Interface, ssclientset ssclientset.Inter
...
+				if err == nil {
+					ssecret, err := ssclientset.BitnamiV1alpha1().SealedSecrets(ns).Get(context.Background(), name, metav1.GetOptions{})
+					if err == nil {
+						key, err := cache.MetaNamespaceKeyFunc(ssecret)
+						if err == nil {
+							queue.Add(key)
+						}
+					}
+				}
+			}
```
josvazg, line 123: Nitpick:
Not a fan of nested **ifs** "foreva"
I'd rather fail fast and keep the happy path on the "left lane", straight down.

alemorcuq: I agree.

integration/controller_test.go
```diff
@@ -267,6 +267,28 @@ var _ = Describe("create", func() {
 				}, Timeout, PollingInterval).Should(WithTransform(getFirstOwnerName, Equal(ss.GetName())))
 			})
 		})
+
+		Context("Secret recreation", func() {
+			BeforeEach(func() {
```
alemorcuq, line 272: Not sure about this. I think this is being executed before the SealedSecret is created?
The SealedSecret is created in a `JustBeforeEach` function.

pkg/controller/controller.go
```diff
@@ -102,8 +105,28 @@ func NewController(clientset kubernetes.Interface, ssclientset ssclientset.Inter
 		},
 	})
 
+	infsecret := informers.NewSharedInformerFactory(clientset, 0).Core().V1().Secrets().Informer()
```
alemorcuq, line 108: Why are we creating this informer in a different way as the Sealed Secret informer?

pkg/controller/controller.go
```diff
@@ -102,8 +105,28 @@ func NewController(clientset kubernetes.Interface, ssclientset ssclientset.Inter
...
+						}
+					}
+				}
+			}
+		},
+	})
+
 	return &Controller{
 		informer:    informer,
+		infsecret:   infsecret,
```
alemorcuq, line 129: Not a fan of this name. Perhaps it would be better to have `ssInformer` for the SealedSecrets informer and `sInformer` for the Secret informer?

alemorcuq (changes requested):

pkg/controller/controller.go
alvneiayu: it is not the same informer. This is an informer generated directly by the client-go API. This is why I needed to include the library.

integration/controller_test.go
alvneiayu: let me check, you are right. Thanks ^^

pkg/controller/controller.go
alvneiayu: Let me check the impact of changing the two of them. Of course, I am going to modify the new one, I like your proposal.

pkg/controller/controller.go
alvneiayu: Modified. Let me know if you think that it is ok for you. Thanks

integration/controller_test.go
```diff
@@ -267,6 +267,32 @@ var _ = Describe("create", func() {
 				}, Timeout, PollingInterval).Should(WithTransform(getFirstOwnerName, Equal(ss.GetName())))
 			})
 		})
+
+		Context("Secret recreation", func() {
```
alemorcuq, line 271:
```suggestion
		Context("With secret being deleted", func() {
```

integration/controller_test.go
```diff
@@ -267,6 +267,32 @@ var _ = Describe("create", func() {
...
+			JustBeforeEach(func() {
+				Eventually(func() (*v1.Secret, error) {
+					return c.Secrets(ns).Get(ctx, secretName, metav1.GetOptions{})
+				}, Timeout, PollingInterval).Should(WithTransform(getFirstOwnerName, Equal(ss.GetName())))
+				err:= c.Secrets(ns).Delete(ctx, secretName, metav1.DeleteOptions{})
+				Expect(err).NotTo(HaveOccurred())
+			})
+			It("should the Secret exists", func() {
```
alemorcuq, line 279:
```suggestion
			It("should recreate the secret", func() {
```

alemorcuq (changes requested):

pkg/controller/controller.go
```diff
@@ -102,8 +105,36 @@ func NewController(clientset kubernetes.Interface, ssclientset ssclientset.Inter
 		},
 	})
 
+	sInformer := sinformer.Core().V1().Secrets().Informer()
+	sInformer.AddEventHandler(cache.ResourceEventHandlerFuncs{
+		DeleteFunc: func(obj interface{}) {
+			key, err := cache.MetaNamespaceKeyFunc(obj)
+			if err != nil {
+				return
```
josvazg, line 113: We should log the error at the very least.

```suggestion
			if err != nil {
			        log.Printf("failed to fetch key: %v", err)
				return
```

Otherwise we expose ourselves to long debugging hours for no good reason.

Same for the other returns.

pkg/controller/controller.go
```diff
@@ -102,8 +105,36 @@ func NewController(clientset kubernetes.Interface, ssclientset ssclientset.Inter
...
+			}
+
+			key, err = cache.MetaNamespaceKeyFunc(ssecret)
+			if err != nil {
+				return
+			}
+
+			queue.Add(key)
```
josvazg, line 131: If this `key` is different form the one above I would use a different name for it.

josvazg (changes requested):
I think we should address those errors being shallower silently.

pkg/controller/controller.go
alvneiayu: Let me check the messages. I am going to review the other returns but several of them are used only to check that the Secret has a Sealed Secret. Include messages to those parts are going to generate more noise than help.

pkg/controller/controller.go
```diff
@@ -102,8 +105,36 @@ func NewController(clientset kubernetes.Interface, ssclientset ssclientset.Inter
...
+
+			ssecret, err := ssclientset.BitnamiV1alpha1().SealedSecrets(ns).Get(context.Background(), name, metav1.GetOptions{})
+			if err != nil {
+				return
+			}
+
+			key, err = cache.MetaNamespaceKeyFunc(ssecret)
+			if err != nil {
+				return
+			}
```
josvazg, line 129: I understand these are more checks that usual sources of errors, but still the fact that the function returns a `error` type and not a boolean means we need to check what error we are getting.

If the error is the expected, like `not found` then we can silently return. If the error is something else we should log it as it is unexpected.

pkg/controller/controller.go
alemorcuq: I agree with @alvneiayu. This code is checking if a Secret is managed by a certain SealedSecret, I don't think we should polute the logs with messages about Secrets not managed by a Sealed Secret.

josvazg: I am totally fine with skipping expected  error results, such as `not found`. But we should still log if the error was something else we did not expect to happen.

pkg/controller/controller.go
```diff
@@ -102,8 +115,47 @@ func NewController(clientset kubernetes.Interface, ssclientset ssclientset.Inter
...
+		DeleteFunc: func(obj interface{}) {
+			skey, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
+			if err != nil {
+				log.Printf("failed to fetch Secret key: %v", err)
+				return
+			}
+
+
+			if (!findOwnerReferenceSS(obj)) {
+				if (!isAnnotatedToBeManaged(obj.(*corev1.Secret))) {
```
josvazg, line 129: This could be a simple && expression right?

Not a big deal anyways... I am not fan on nesting I guess, specially when not needed.

alvneiayu: totally :')

pkg/controller/controller.go
```diff
@@ -102,8 +115,47 @@ func NewController(clientset kubernetes.Interface, ssclientset ssclientset.Inter
...
+			if err != nil {
+				log.Printf("failed to find the Sealed Secret associated: %v", err)
+				return
+			}
+
+			sskey, err := cache.MetaNamespaceKeyFunc(ssecret)
+			if err != nil {
+				log.Printf("failed to fetch Sealed Secret key: %v", err)
+				return
+			}
```
josvazg, line 150: I am not sure about the second case but the first one seems like not found is perfectly normal, so I would do this instead:

```suggestion
			if err != nil {
			        if !errors.Is(err, NotFound) { // Only log if the error is unexpected at this point
				  log.Printf("failed to find the Sealed Secret associated: %v", err)
				}
				return
			}

			sskey, err := cache.MetaNamespaceKeyFunc(ssecret)
			if err != nil {
				log.Printf("failed to fetch Sealed Secret key: %v", err)
				return
			}
```

alvneiayu: With the isAnnotatedToBeManaged and the findOwnerReferenceSS, you are doing the checks previous to this code block. At this point, your secret must have always a sealed secret, if not you are having something not aligned in your cluster.

josvazg (approved):
LGTM

Just a nitpick left, but is fine

josvazg (approved):

pkg/controller/controller.go
```diff
@@ -102,8 +106,46 @@ func NewController(clientset kubernetes.Interface, ssclientset ssclientset.Inter
...
+
+			ns, name, err := cache.SplitMetaNamespaceKey(skey)
+			if err != nil {
+				log.Printf("failed to split metadatada namespace: %v", err)
+				return
+			}
+
+			ssecret, err := ssclientset.BitnamiV1alpha1().SealedSecrets(ns).Get(context.Background(), name, metav1.GetOptions{})
+			if err != nil {
+				if !strings.Contains(err.Error(), "not found") {
```
alemorcuq, line 126:
```suggestion
				if !errors.IsNotFound(err) {
```

pkg/controller/controller.go
```diff
@@ -102,8 +106,46 @@ func NewController(clientset kubernetes.Interface, ssclientset ssclientset.Inter
...
+		DeleteFunc: func(obj interface{}) {
+			skey, err := cache.DeletionHandlingMetaNamespaceKeyFunc(obj)
+			if err != nil {
+				log.Printf("failed to fetch Secret key: %v", err)
+				return
+			}
+
+			ns, name, err := cache.SplitMetaNamespaceKey(skey)
+			if err != nil {
+				log.Printf("failed to split metadatada namespace: %v", err)
```
alemorcuq, line 120:
```suggestion
				log.Printf("failed to get namespace and name from key: %v", err)
```

pkg/controller/controller.go
```diff
@@ -102,8 +106,46 @@ func NewController(clientset kubernetes.Interface, ssclientset ssclientset.Inter
...
+				return
+			}
+
+			ssecret, err := ssclientset.BitnamiV1alpha1().SealedSecrets(ns).Get(context.Background(), name, metav1.GetOptions{})
+			if err != nil {
+				if !strings.Contains(err.Error(), "not found") {
+					log.Printf("failed to find the Sealed Secret associated: %v", err)
```
alemorcuq, line 127: This log is only printed if the error is not "not found", implying something went wrong when getting the SealedSecret. 

```suggestion
					log.Printf("failed to get SealedSecret: %v", err)
```

pkg/controller/controller.go
```diff
@@ -102,8 +106,46 @@ func NewController(clientset kubernetes.Interface, ssclientset ssclientset.Inter
...
+				return
+			}
+
+			if !metav1.IsControlledBy(obj.(*corev1.Secret), ssecret) && !isAnnotatedToBeManaged(obj.(*corev1.Secret)) {
+				return
+			}
+
+			sskey, err := cache.MetaNamespaceKeyFunc(ssecret)
+			if err != nil {
+				log.Printf("failed to fetch Sealed Secret key: %v", err)
```
alemorcuq, line 138:
```suggestion
				log.Printf("failed to fetch SealedSecret key: %v", err)
```

alemorcuq (changes requested):

alemorcuq (approved):

josvazg (approved):

mateuszdrab:
Hi @josvazg 
Thanks for merging this, I need this feature as without it - my flux deployment doesn't work due to the secret not being recreated.
Can you advise in which release will this feature appear?

josvazg:
Hi @mateuszdrab 
We are working on a new formal monthly schedule for our releases, see #992

PR #965, opened 2022-09-19
fix: CRDs should not include unnecessary dashes

juan131:
Signed-off-by: juan131 <juan.ariza.1311993@gmail.com>

### Description of the change

The Sealed Secrets chart's CRD start with 3 dashes that are not necessary since these are only required as a separator when N directives are included in a single YAML file (which is not the case). Indeed, if you render the templates including the CRDs (e.g. running  `helm template sealed-secrets sealed-secrets/sealed-secrets --include-crds > foo.yaml` and inspect the output, you'll find blocks such as the one below:

```
---
# Source: crds/bitnami.com_sealedsecrets.yaml
---
```

This is interpreted as an empty directive which doesn't make sense.

### Benefits

Charts compatible with `helm template --include-crds ...`.

### Possible drawbacks

None

### Applicable issues

N/A

### Additional information

This bug was introduced at https://github.com/bitnami-labs/sealed-secrets/commit/7d6ae6e2cc013f3f760ca8748057f68195f5d833

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ generate: $(GO_FILES)
 	cp -r gentmp/github.com/bitnami-labs/sealed-secrets/pkg . && rm -rf gentmp/
 
 manifests:
-	$(CONTROLLER_GEN) crd paths="./pkg/apis/..."  output:crd:artifacts:config=helm/sealed-secrets/crds/
+	$(CONTROLLER_GEN) crd paths="./pkg/apis/..." output:stdout | tail -n +2 > helm/sealed-secrets/crds/bitnami.com_sealedsecrets.yaml
 
 controller: $(GO_FILES)
 	$(GO) build -o $@ $(GO_FLAGS) -ldflags "$(GO_LD_FLAGS)" ./cmd/controller
--- a/helm/sealed-secrets/Chart.yaml
+++ b/helm/sealed-secrets/Chart.yaml
@@ -14,4 +14,4 @@ maintainers:
     url: https://github.com/bitnami-labs/sealed-secrets
 name: sealed-secrets
 type: application
-version: 2.6.5
+version: 2.6.6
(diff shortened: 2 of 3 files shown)


alvneiayu (dismissed):
LGTM

josvazg (changes requested):
We should include a fix for the code generation itself.

josvazg:
Seems this would do the trick:
```bash
$ git diff
diff --git Makefile Makefile
index 7a95a5b..33e440b 100644
--- Makefile
+++ Makefile
@@ -55,7 +55,7 @@ generate: $(GO_FILES)
        cp -r gentmp/github.com/bitnami-labs/sealed-secrets/pkg . && rm -rf gentmp/
 
 manifests:
-       $(CONTROLLER_GEN) crd paths="./pkg/apis/..."  output:crd:artifacts:config=helm/sealed-secrets/crds/
+       $(CONTROLLER_GEN) crd paths="./pkg/apis/..." output:stdout | tail -n +2 > helm/sealed-secrets/crds/bitnami.com_sealedsecrets.yaml
 
 controller: $(GO_FILES)
        $(GO) build -o $@ $(GO_FLAGS) -ldflags "$(GO_LD_FLAGS)" ./cmd/controller
```

juan131:
Feel free to amend this PR @josvazg , I'm not familiar with the automations around the CRD included in the chart.

josvazg:
> Feel free to amend this PR @josvazg , I'm not familiar with the automations around the CRD included in the chart.

Thanks @juan131 I will take care of this. And thanks for reporting!

Not an expert with controller-gen, not 100% sure there is no flag or tag that would do this for us. So for now it seems that just removing the unwanted prefix after generation would work for us.

josvazg:
Looking good:
```bash
$ make manifests
controller-gen crd paths="./pkg/apis/..." output:stdout | tail -n +2 > helm/sealed-secrets/crds/bitnami.com_sealedsecrets.yaml

$ git status
On branch fix/helm-crd-dashes
Your branch is up to date with 'juan131/fix/helm-crd-dashes'.

nothing to commit, working tree clean
```

josvazg (approved):
LGTM as we patched the re-generation rule

alemorcuq (approved):
Great, thanks!

juan131:
Thanks guys!!

Issue #976, opened 2022-09-23
Controller.yaml manifest doesn't contain image name


ViBiOh:
**Which component**:

`controller.yaml` file in the release of 0.18.4

**Describe the bug**

There is no image name in the `controller.yaml` asset in [the release](https://github.com/bitnami-labs/sealed-secrets/releases/tag/v0.18.4)

**To Reproduce**

Download the file and find that the yaml contains `image: :v0.18.4`

**Expected behavior**

Download the file and find that the yaml contains `image: somewhere/bitnami/sealed-secrets-controller:v0.18.4`

agarcia-oss:
Hi @ViBiOh you're absolutely right; there's an issue affecting the current `controller.yaml` file. We'll generate a new patch release shortly.

alemorcuq:
`v0.18.5` was released with a valid [controller.yaml](https://github.com/bitnami-labs/sealed-secrets/releases/download/v0.18.5/controller.yaml)

ViBiOh:
Thank you for quick feedback ;)

PR #977, opened 2022-09-23
Fix controller image reference

alemorcuq:
**Description of the change**
Use the right environment variable for the controller image name so that the `controller.yaml` file is properly generated.

Reported in #976


alvneiayu (approved):
LGTM

agarcia-oss (approved):
Env variables seem to be Ok now

PR #984, opened 2022-09-27
WIP: Enforce go versions from versions.env

josvazg:
**Description of the change**

Define built or used Go versions from a single main definition file. And use a check script to enforce all CI usages comply with the versions defined there.

**Benefits**

The Go versions are defined in as few places as possible and any inconsistencies are detected early on.

**Additional information**

As I did not find a way for Github CI definitions to accept external inputs on env vars, I cannot use a single source of truth. So I have to resort to a extra check to enforce consistency.


josvazg:
Note that I expect this first version to fail, as I made the CI variables and the desired valued to purposefully differ with each other.

josvazg:
`Error: Unable to find Go version '$GO_VERSION' for platform linux and architecture x64.`

This error was not expected, it seems GitHub action matrices do not accept env vars?

josvazg:
https://github.com/bitnami-labs/sealed-secrets/actions/runs/3136137420/jobs/5092670158 is showing the expected errors:
```
./scripts/check-versions.sh
Fix .github/workflows/ci.yml workflow environment variables to set GO_VERSION=1.18.6 (instead of 1.18.5)
Fix .github/workflows/ci.yml workflow matrix to be 'go: ["1.18.6", "1.19.1"]' instead of 'go:'
Fix .github/workflows/release.yaml workflow environment variables to set GO_VERSION=1.18.6 (instead of 1.18.7)
Found 3 version mistmatchs between CI settings and main version definition file: versions.env
```
Will fix them now and expect all to pass.

josvazg:
Moving to test this in my own repo, to avoid all the noise.

PR #985, opened 2022-09-28
Use a single versioning source

josvazg:
**Description of the change**

Enforce a single place to control CI Go versioning.

The new `version.env` now controls all CI Go version settings.

**Benefits**

Avoids possible mistakes or inconsistencies when upgrading go versioning.


alemorcuq (approved):
This is unrelated to the change, but I would remove the complexity of having the `go_version_list` parameter. IMO, there's not much value in running the CI with several versions of Go.

agarcia-oss:
> This is unrelated to the change, but I would remove the complexity of having the go_version_list parameter. IMO, there's not much value in running the CI with several versions of Go.

Good suggestion, @alemorcuq but I think that can be tackle with another task.

agarcia-oss (approved):
LGTM! Maybe is worth mentioning that, since we're simplifying the CI job, we no longer evaluate the code using multiple `golang` versions (only with the target version)

alvneiayu (approved):
LGTM

alemorcuq:
> Good suggestion, @alemorcuq but I think that can be tackle with another task.

But why add it now if we are not using it? We are adding complexity but we are not taking advantage of it.

josvazg:
> > This is unrelated to the change, but I would remove the complexity of having the go_version_list parameter. IMO, there's not much value in running the CI with several versions of Go.
> 
> Good suggestion, @alemorcuq but I think that can be tackle with another task.

The list still required AFAIK:
```shell
GO_VERSION_LIST="[\"$GO_VERSION\"]"
```

Note that even when we have a single entry, like in `golangci-lint: ["1.49.0"]` be need to be explicit about it and we also need to make it an array.

This version is already using `go: ["1.18.6"]` a single version and you only need to update the version in one single line, the first one in `versions.env`:
```shell
GO_VERSION=1.18.6
GO_VERSION_LIST="[\"$GO_VERSION\"]"
```

alemorcuq:
> > > This is unrelated to the change, but I would remove the complexity of having the go_version_list parameter. IMO, there's not much value in running the CI with several versions of Go.
> > 
> > 
> > Good suggestion, @alemorcuq but I think that can be tackle with another task.
> 
> The list still required AFAIK:
> 
> ```shell
> GO_VERSION_LIST="[\"$GO_VERSION\"]"
> ```
> 
> Note that even when we have a single entry, like in `golangci-lint: ["1.49.0"]` be need to be explicit about it and we also need to make it an array.

I don't see any issue on removing those single value lists from the matrices.

PR #995, opened 2022-10-06
Remove binary

alemorcuq:
Signed-off-by: Alejandro Moreno <amorenoc@vmware.com>

**Description of the change**
Remove controller binary as it should not be there.


josvazg (approved):
LGTM

Let's check .gitignore as well

PR #1003, opened 2022-10-11
Fix solving GO_VERSION from env

josvazg:
Signed-off-by: Jose Luis Vazquez Gonzalez <josvaz@vmware.com>

**Description of the change**

This was wrong since #985, but we did not notice as releases are not tested in the PR process.

Just noticed it while working on the upcoming release.


alvneiayu (approved):
LGTM

PR #1006, opened 2022-10-13
Helm chart: field 'rules' removed from 'templates/psp-clusterrolebind…

spreger5:
…ing.yaml'

Signed-off-by: Gerhard Sprenger <spreger5@yahoo.de>

**Description of the change**
The template psp-clusterrolebinding.yaml contains an empty filed 'rules:'. This field is not supported on kind 'ClusterRoleBinding' according to 'kubectl explain clusterrolebinding'. Manifest validators like kubeconform fail on this ressource. It might be a copy-paste-error, sincs it's a valid field for [cluster-]roles.

**Benefits**

**Possible drawbacks**

**Applicable issues**

- fixes #

**Additional information**


alemorcuq (approved):
Thanks!

PR #1011, opened 2022-10-13
Upgrade node 12 actions to node 16 versions

mpermar:
**Description of the change**

GitHub has deprecated Node.js 12 actions. This change upgrades the two actions that use node 12 to their most recent versions which are node.js 16 based.

**Possible drawbacks**

There shouldn't be any drawbacks but we need to see if this works.

Signed-off-by: Martin Perez <martinpe@vmware.com>


alemorcuq (approved):

alvneiayu (approved):
LGTM

PR #1012, opened 2022-10-13
Try to fix release 0.19.0 tags in dockerhub

josvazg:
**Description of the change**

Try to fix DockerHub tags


josvazg:
Permissions issues, better do it properly another day.

PR #1014, opened 2022-10-14
Fix container image release tag

josvazg:
**Description of the change**

Fix the image container release tags.

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -44,7 +44,9 @@ jobs:
       - name: K8s manifests
         run: |
           export PATH=~/bin:$PATH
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yaml
```diff
@@ -44,7 +44,9 @@ jobs:
       - name: K8s manifests
         run: |
           export PATH=~/bin:$PATH
-          make CONTROLLER_IMAGE=${{ env.controller_dockerhub_image_name }}:${{ github.ref_name }} controller.yaml controller-norbac.yaml
```
josvazg, line 47: `github.ref_name` ends up being tag `release-vX.Y.Z` otherwise.

alvneiayu (approved):
LGTM

PR #1029, opened 2022-10-18
Remove namespaceFn

josvazg:
**Description of the change**

This configurable function behavior was only used on a test and otherwise making the production code more complex for no good reason.

This change removes the function definition and instead uses an interface to overload the behavior of the `clientConfig` to ensure the test was still behaving the same way.

**Benefits**

Simpler code that is easier to be further evolved and inproved.


alvneiayu (approved):
LGTM

agarcia-oss (approved):

PR #1030, opened 2022-10-18
Decouple the kubeseal CLI from the kubeseal library

alemorcuq:
Signed-off-by: Alejandro Moreno <amorenoc@vmware.com>

**Description of the change**
Decouple the kubeseal CLI specific code from its library.

Note: **The currently exposed `kubeseal` interface must be considered unstable, it will change going forward**.

**Benefits**
Makes the library easier to reuse.

**Possible drawbacks**
As with all refactors, bugs could be introduced. We will ensure however our test suite continues to pass.

### DIFF
--- a/cmd/kubeseal/main.go
+++ b/cmd/kubeseal/main.go
@@ -36,35 +36,35 @@ var (
 	VERSION = buildinfo.DefaultVersion
 )
 
-type Flags struct {
-	CertURL        string
-	ControllerNs   string
-	ControllerName string
-	OutputFormat   string
-	OutputFileName string
-	InputFileName  string
-	Kubeconfig     string
-	DumpCert       bool
-	AllowEmptyData bool
-	ValidateSecret bool
-	MergeInto      string
-	Raw            bool
-	SecretName     string
-	FromFile       []string
-	SealingScope   ssv1alpha1.SealingScope
-	ReEncrypt      bool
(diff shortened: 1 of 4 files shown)


josvazg:
I believe it should now pass tests.

josvazg (dismissed):
LGTM, but lets get some more approvals here

josvazg:
BTW this will conflict once we merge #1029

josvazg:
will check the conflicts locally instead

josvazg:
Special mention: Recent removal of the `NamespaceFn` (see #1029)
That function was only needed for a single test (`TestRaw`) that required to customize the namespace used in such test. The way it was done was using a custom Interface `ClientConfig` that only container a couple of the methods from `clientcmd.ClienConfig` used by the kubeseal code here. It was ten trivial to use a wrapper object that re-implemented the `Namespace()` method for our test.

For the point of view if the current code, it might not even be needed to use the type wrapper trick above. But this change has tried to preserve that and leaves further refactor improvements for follow up PRs.

josvazg (dismissed):

josvazg (approved):

alvneiayu (approved):
LGTM

bakito:
@alemorcuq finally, thank you very much for this decoupling I was waiting for this for along time. 🥳

PR #1044, opened 2022-10-27
Fix empty Group Version Kind

josvazg:
**Description of the change**

Fix empty Group Version Kind fields which can happen at times. See:
https://github.com/operator-framework/operator-sdk/issues/727

Some typos have also been opportunistically fixed.

**Benefits**

The Object is completed at generation/cast time, so it can be relied to be completed later on.


alvneiayu (approved):
LGTM

PR #1081, opened 2022-12-21
Create secret informers only when needed

### LINKED ISSUE #1064: Watch `Secret` permission introduced by PR #963 breaks dynamic low-privilege multi-tenant set-ups
## Which component
controller

## Describe the bug
Due to the implementation of #963 , versions of the sealed secrets controller >= `v0.19.0` are no longer installable in dynamic low-privilege multi-tenant environments. The properties of the environment are:

- **multi-tenant**: Namespaces are isolated "per tenant"
- **low-privilege**: The sealed secret controller must only access resources of kind `Secret` in specific namespaces; No `ClusterRoleBinding` can be used.
- **dynamic**: Namespaces are added/removed in a high frequency.

The implementation of #568 already allows watching only specific namespaces. However, this static list must be adapted whenever a namespace is added / removed.

### Dynamically create `Secret` for namespaces that contain `SealedSecret` resources

Whenever the sealed secret controller picks up a new `SealedSecret` resource, it creates a watch for `Secret` resources within the same namespace, unless a watch already exists. Whenever a `SealedSecret` resource is deleted, the watch for `Secret` resources within the same namespace is removed, if it was the last `SealedSecret` in the same namespace.

### Watch Namespaces for specific labels to include

Instead of maintaining a static list of namespaces when deploying the controller, the controller could watch namespace resources for a specific label (e.g. `sealed-secrets: enabled`). If the label is present, the controller creates watches for `Secret` (and `SealedSecret`?) resources.

### Feature flag for #963 

Another option could be a feature flag to disable the behavior introduced by #963 .

alemorcuq:
**Description of the change**
Secret informers are now created as required, needing less privileges as only namespaces with sealed secrets will be watched.

**Applicable issues**
- fixes #1064

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -25,7 +25,8 @@ import (
 	"k8s.io/client-go/util/workqueue"
 	"k8s.io/klog"
 
-	"k8s.io/client-go/informers"
+	corev1informers "k8s.io/client-go/informers/core/v1"
+	"k8s.io/client-go/informers/internalinterfaces"
 
 	ssv1alpha1 "github.com/bitnami-labs/sealed-secrets/pkg/apis/sealedsecrets/v1alpha1"
 	ssclientset "github.com/bitnami-labs/sealed-secrets/pkg/client/clientset/versioned"
@@ -64,7 +65,7 @@ var (
 type Controller struct {
 	queue       workqueue.RateLimitingInterface
 	ssInformer  cache.SharedIndexInformer
-	sInformer   cache.SharedIndexInformer
+	sInformers  map[string]*secretInformer
 	sclient     v1.SecretsGetter
 	ssclient    ssv1alpha1client.SealedSecretsGetter
 	recorder    record.EventRecorder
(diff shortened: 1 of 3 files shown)


pkg/controller/controller.go
```diff
@@ -93,6 +100,63 @@ func NewController(clientset kubernetes.Interface, ssclientset ssclientset.Inter
 			key, err := cache.MetaNamespaceKeyFunc(obj)
 			if err == nil {
 				queue.Add(key)
+				ns, _, err := cache.SplitMetaNamespaceKey(key)
+				if err != nil {
+					log.Printf("failed to get namespace and name from key: %v", err)
+					return
+				}
+				if _, ok := sInformers[ns]; !ok {
+					// chache.Indexers intitialization: https://github.com/kubernetes/client-go/blob/e7cd4ba474b5efc2882e377362c9aa8b407428d9/informers/core/v1/secret.go#L81
```
josvazg, line 109: This is too big now, I would move it to a regular function...

```go
AddFunc: addHandler

...

func addHandler(obj interface{}) {
...
```

pkg/controller/controller.go
```diff
@@ -93,6 +100,63 @@ func NewController(clientset kubernetes.Interface, ssclientset ssclientset.Inter
...
+					inf := corev1informers.NewFilteredSecretInformer(clientset, ns, 0, cache.Indexers{cache.NamespaceIndex: cache.MetaNamespaceIndexFunc}, tweakopts)
+					stopCh := make(chan struct{})
+
+					_, err = inf.AddEventHandler(cache.ResourceEventHandlerFuncs{
+						DeleteFunc: func(obj interface{}) {
```
josvazg, line 114: Inception! (add handled setting delete handler)

I would add a regular function here instead. It will be more readable.

armingerten:
Any updates on this? 😉

PR #1098, opened 2023-01-25
feat: add drop all capabilities

JuanPabloSGU:
**Description of the change**

Added drop all capabilities to the container security context. 

**Benefits**

Capabilities permit privileged actions without giving full root access. All capabilities should be dropped from the container, with only those required added back.

**Possible drawbacks**
None

**Applicable issues**

- fixes #1060 

**Additional information**

PR #1116, opened 2023-02-22
feat: add drop all capabilities

### LINKED ISSUE #1060: [feat] Add Drop ALL Capabilities
**Which component**:
This would affect controller (v0.19.2).

Currently it seems that 

```YAML
capabilities:
  drop: 
    - ALL
```

is missing from the container security context. If the container does not require those capabilities, would it be possible to drop them by default.

The solution would have drop all capabilities by default in the controller if possible.

NONE

**Additional context**

JuanPabloSGU:
**Description of the change**
Added drop all capabilities to the container security context.

**Benefits**
Capabilities permit privileged actions without giving full root access. All capabilities should be dropped from the container, with only those required added back.

**Possible drawbacks**
None

**Applicable issues**

- fixes #1060 

**Additional information**
This is a redo of a pr that has been already created... see pr #1098

### DIFF
--- a/helm/sealed-secrets/Chart.yaml
+++ b/helm/sealed-secrets/Chart.yaml
@@ -14,4 +14,4 @@ maintainers:
     url: https://github.com/bitnami-labs/sealed-secrets
 name: sealed-secrets
 type: application
-version: 2.7.4
+version: 2.7.5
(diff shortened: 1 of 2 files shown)


alemorcuq (approved):
Thank you very much!

FlorisFeddema:
This breaks our setup which sealed-secrets. We have a namespace with the label `pod-security.kubernetes.io/enforce: restricted`. When deploy this version of sealed-secrets we get an error that says the drop should be `- ALL` instead of  `- all` .

PR #1120, opened 2023-03-02
Run VIB chart checks only after PR approval

josvazg:
**Description of the change**

Run VIB chart validation once a PR was validated.

**Benefits**

Avoid running those expensive tests on each PR change.


agarcia-oss (approved):
LGTM

PR #1124, opened 2023-03-03
change drop capabilities to uppercase

### LINKED ISSUE #1123: PR #1116 adds capabilities drop all in lowercase
**Which component**:
Chart version: 2.7.5

**Describe the bug**
This breaks our setup which sealed-secrets. We have a namespace with the label `pod-security.kubernetes.io/enforce: restricted`.
When deploy this version of sealed-secrets we get an error that says the drop should be `- ALL` instead of `- all` .

**To Reproduce**

1. Create a namespace with label  `pod-security.kubernetes.io/enforce: restricted`
2. Deploy the Sealed-secrets helm chart in that namespace.

**Expected behavior**
The applications runs without problems.

**Version of Kubernetes**:
1.25, but will probably not work with other versions.

**Additional context**
[Pull request #1116 ](https://github.com/bitnami-labs/sealed-secrets/pull/1116) adds this behaviour.

FlorisFeddema:
**Description of the change**
Changes the drop all capability to uppercase to make it correctly work with the `pod-security.kubernetes.io/enforce: restricted` namespace label. 

**Benefits**
Kubernetes needs it to be uppercase instead of lowercase, while most deployments will not break it might break in some cases.

**Possible drawbacks**
None

**Applicable issues**

- fixes #1123

### DIFF
--- a/helm/sealed-secrets/Chart.yaml
+++ b/helm/sealed-secrets/Chart.yaml
@@ -14,4 +14,4 @@ maintainers:
     url: https://github.com/bitnami-labs/sealed-secrets
 name: sealed-secrets
 type: application
-version: 2.7.5
+version: 2.7.6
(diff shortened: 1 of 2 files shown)


alemorcuq:
Thanks!

PR #1128, opened 2023-03-06
Allow automountServiceAccountToken to be set to false

### LINKED ISSUE #1125: Wrong behavior of automountServiceAccountToken
**Which component**:

Controller, v2.7.5

**Describe the bug**

The behavior of automountServiceAccountToken introduced by https://github.com/bitnami-labs/sealed-secrets/pull/885 is like this: 

```
{{- if .Values.automountServiceAccountToken }}
      automountServiceAccountToken: {{ .Values.automountServiceAccountToken }}
{{- end }}
```
See: https://github.com/bitnami-labs/sealed-secrets/blob/main/helm/sealed-secrets/templates/deployment.yaml#L50

Very similar, its also in the service account: 

```
{{- if .Values.serviceAccount.automountServiceAccountToken }}
automountServiceAccountToken: {{ .Values.serviceAccount.automountServiceAccountToken }}
{{- end }}
```
See: https://github.com/bitnami-labs/sealed-secrets/blob/main/helm/sealed-secrets/templates/service-account.yaml

I believe this is wrong. In restricted environments we need to *specifically disable* SA automounting and thus the deployment yaml as well as the SA yaml must containt the following: 

```
automountServiceAccountToken: false
```

It unfortunately doesnt work just by ommiting the automountServiceAccountToken altogether. I believe the logic surrounding `automountServiceAccountToken` should be changed in both cases to allow for a explicit `false`. 

**To Reproduce**

Just add this to your Values: 

```
# This changes the deployment
automountServiceAccountToken: false
serviceAccount:
  # This changes the Service Account
  automountServiceAccountToken: false
```

**Expected behavior**

We expect to find `automountServiceAccountToken: false` in both Deployment as well as in the Service Account.

alemorcuq:
**Description of the change**
This fixes the logic so that `automountServiceAccountToken` can be explicitly set to false to opt out of this feature, as documented here: https://kubernetes.io/docs/tasks/configure-pod-container/configure-service-account/#opt-out-of-api-credential-automounting

**Applicable issues**
- fixes #1125

### DIFF
--- a/helm/sealed-secrets/templates/deployment.yaml
+++ b/helm/sealed-secrets/templates/deployment.yaml
@@ -46,9 +46,7 @@ spec:
       securityContext: {{- omit .Values.podSecurityContext "enabled" | toYaml | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ include "sealed-secrets.serviceAccountName" . }}
-      {{- if .Values.automountServiceAccountToken }}
       automountServiceAccountToken: {{ .Values.automountServiceAccountToken }}
-      {{- end }}
       {{- if .Values.hostNetwork }}
       hostNetwork: true
       {{- end }}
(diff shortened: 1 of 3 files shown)


josvazg (approved):

alvneiayu (approved):
LGTM

PR #1140, opened 2023-03-07
Run VIB Helm chart validations on push to main

josvazg:
**Description of the change**

Run VIB validations only when merging or pushing into `main`.

### DIFF
--- a/.github/workflows/helm-vib-lint.yaml
+++ b/.github/workflows/helm-vib-lint.yaml
@@ -1,0 +1,23 @@
+name: Lint Helm Chart
+on:
+  workflow_dispatch:
+  push:
+    paths:
+       - 'helm/**'
+
+env:
+  CSP_API_URL: https://console.cloud.vmware.com
+  CSP_API_TOKEN: ${{ secrets.CSP_API_TOKEN }}
+  VIB_PUBLIC_URL: https://cp.bromelia.vmware.com
+
+jobs:
+  # make sure chart is linted/safe
+  vib-validate:
+    runs-on: ubuntu-latest
+    name: Lint chart
+    steps:
+      - uses: actions/checkout@v3.1.0
+        with:
+          ref: ${{github.event.pull_request.head.ref}}
+          repository: ${{github.event.pull_request.head.repo.full_name}}
+      - uses: vmware-labs/vmware-image-builder-action@0.4.7
--- a/.github/workflows/helm-vib-lint.yaml
+++ b/.github/workflows/helm-vib-lint.yaml
@@ -2,6 +2,9 @@ name: Lint Helm Chart
 on:
   workflow_dispatch:
   pull_request_target:
+    branches:
+      - main
+      - bitnami-labs:main
     paths:
        - 'helm/**'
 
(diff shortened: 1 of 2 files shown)


.github/workflows/helm-vib.yaml
```diff
@@ -1,7 +1,7 @@
 name: Verify Helm Chart
 on:
   workflow_dispatch:
-  pull_request_target:
+  push:
```
alvneiayu, line 4: I am a little bit confuse, why we are changing from pull_request_target to push? I mean, we are going to launch this VIB action every time that we are pushing, no? Am I wrong?

josvazg: No, it only triggers when you **push** to `main`, and that ONLY happens on merges, as `main` is protected.

.github/workflows/helm-vib-lint.yaml
```diff
@@ -0,0 +1,23 @@
+name: Lint Helm Chart
```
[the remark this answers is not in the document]
josvazg: You have a point there, maybe using `pull_request_target` is enough as we have tested this pre-merge.

.github/workflows/helm-vib.yaml
alemorcuq: I think this is a leftover:

```
bitnami-labs:main
```

Just:

```
branches:
  - main
```

.github/workflows/helm-vib-lint.yaml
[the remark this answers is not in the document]
josvazg: changes to lint only on PRs

.github/workflows/helm-vib.yaml
```diff
@@ -1,7 +1,7 @@
...
     branches:
       - main
       - bitnami-labs:main
```
josvazg, line 7:
```suggestion
```

.github/workflows/helm-vib.yaml
josvazg: fixed

.github/workflows/helm-vib-lint.yaml
```diff
@@ -0,0 +1,23 @@
...
+on:
+  workflow_dispatch:
+  pull_request_target:
```
alemorcuq, line 4:
```suggestion
  pull_request_target:
    branches:
      - main
      - bitnami-labs:main
```

This is how it was before and I think it is fine. Only run on PRs whose target is `main` or `bitnami-labs:main`

josvazg: ok

.github/workflows/helm-vib-lint.yaml
```diff
@@ -0,0 +1,26 @@
...
+    branches:
+      - main
+      - bitnami-labs:main
+    paths:
+       - 'helm/**'
```
alemorcuq, line 9:
```suggestion
  pull_request_target:
    branches:
       - main
       - bitnami-labs:main
    paths:
       - 'helm/**'
```

Alignment

josvazg: You are right, but the misalignment is on the paths I think.

```suggestion
  pull_request_target:
    branches:
      - main
      - bitnami-labs:main
    paths:
      - 'helm/**'
```

alemorcuq (approved):

alvneiayu (approved):
LGTM

Issue #1151, opened 2023-03-19
Usage of skipRecreate causes pods Crash loops with Go Panic 


tewfik-ghariani:
**Which component**:
controller

**Describe the bug**
Hii! Thanks a lot for adding the `skipRecreate` flag per https://github.com/bitnami-labs/sealed-secrets/pull/1118

However, after enabling it, the controller pod unfortunately fell into a crash loop with the following Go panic

```
2023/03/19 21:23:15 Starting sealed-secrets controller version: 0.20.1
2023/03/19 21:23:15 Searching for existing private keys
2023/03/19 21:23:15 ----- sealed-secrets-pub-key-15465
E0319 21:23:15.188417       1 runtime.go:79] Observed a panic: "invalid memory address or nil pointer dereference" (runtime error: invalid memory address or nil pointer dereference)
goroutine 127 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic({0x1731ec0?, 0x27ee9d0})
	/home/runner/go/pkg/mod/k8s.io/apimachinery@v0.26.2/pkg/util/runtime/runtime.go:75 +0x99
k8s.io/apimachinery/pkg/util/runtime.HandleCrash({0x0, 0x0, 0x0?})
	/home/runner/go/pkg/mod/k8s.io/apimachinery@v0.26.2/pkg/util/runtime/runtime.go:49 +0x75
panic({0x1731ec0, 0x27ee9d0})
	/opt/hostedtoolcache/go/1.19.4/x64/src/runtime/panic.go:884 +0x212
github.com/bitnami-labs/sealed-secrets/pkg/controller.(*Controller).Run(0xc00014a230, 0xc000152660)
	/home/runner/work/sealed-secrets/sealed-secrets/pkg/controller/controller.go:210 +0x136
created by github.com/bitnami-labs/sealed-secrets/pkg/controller.Main
	/home/runner/work/sealed-secrets/sealed-secrets/pkg/controller/main.go:210 +0x445
2023/03/19 21:23:15 HTTP server serving on :8080
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
	panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x0 pc=0x1568256]

goroutine 127 [running]:
k8s.io/apimachinery/pkg/util/runtime.HandleCrash({0x0, 0x0, 0x0?})
[...]

PR #1152, opened 2023-03-20
Fix panic when skip-recreate is enabled

alemorcuq:
**Description of the change**
Fixes a panic caused when the skip recreate feature is enabled due to the secrets informer being nil.

**Applicable issues**
- #1151


alvneiayu (approved):
LGTM

PR #1174, opened 2023-04-10
ReProcess only on spec changes

josvazg:
**Description of the change**

Right now unsealing happens on every change on the Sealed Secret. That means the secret is sometimes updated for no reason, as the sealed secret changes were in bits that do NOT affect the managed secret, such as. Sample of those updates are Status only changes, which seems to happen after every sealed secret creation.

**Benefits**

Log observers will now see that unseal events happen only after actual changes were detected in the sealed secret spec.


alvneiayu (approved):
LGTM

Issue #1171, opened 2023-04-05
Controller stops unsealing when an invalid SealedSecret object is deployed


FSchumacher:
**Which component**:
controller version 0.20.2

**Describe the bug**
The controller stops processing unsealing SealedSecrets, when an invalid SealedSecret is deployed.

In our case, the `template.metadata.labels` has invalid value type (`bool`) instead of expected type (`str`).

**To Reproduce**

1. Deploy the following invalid object definition:
```yaml
apiVersion: bitnami.com/v1alpha1
kind: SealedSecret
metadata:
  name: broken-template-secret
spec:
  encryptedData: {}
  template:
    metadata:
      labels:
        "not.a.string": true
```

2. Go to the controller and watch the logs.

    The controller stops processing and produces the following log lines (of which the last two lines get repeated about every 30s). 
```
W0405 09:08:25.440944 1 reflector.go:347] pkg/mod/k8s.io/client-go@v0.26.2/tools/cache/reflector.go:169: watch of *v1alpha1.SealedSecret ended with: an error on the server ("unable to decode an event from the watch stream: unable to decode watch event: json: cannot unmarshal bool into Go struct field ObjectMeta.spec.template.metadata.labels of type string") has prevented the request from succeeding
W0405 09:08:26.596910 1 reflector.go:424] pkg/mod/k8s.io/client-go@v0.26.2/tools/cache/reflector.go:169: failed to list *v1alpha1.SealedSecret: json: cannot unmarshal bool into Go struct field ObjectMeta.items.spec.template.metadata.labels of type string
E0405 09:08:26.596944 1 reflector.go:140] pkg/mod/k8s.io/client-go@v0.26.2/tools/cache/reflector.go:169: Failed to watch *v1alpha1.SealedSecret: failed to list *v1alpha1.SealedSecret: json: cannot unmarshal bool into Go struct field ObjectMeta.items.spec.template.metadata.labels of type string
```

**Expected behavior**

Log a message about the broken object and *continue* to process the valid sealed secrets.

**Version of Kubernetes**:

- Output of `kubectl version`:

```
Kubernetes Version: v1.24
```

cloud-pushers:
We are facing the same issue.
One faulty sealed-secret blocks the whole controller.

agarcia-oss:
Thanks for the report, @FSchumacher . There is an issue with the controller while dealing with broken Sealed Secrets definition. We'll start working on a fix asap.

PR #1177, opened 2023-04-13
Generate embedded ObjectMeta in CRD

alemorcuq:
**Description of the change**
Add ObjectMeta fields to the CRD to validate fields such as `spec.template.metadata.labels`. This is generated by `controller-gen` by setting `crd:generateEmbeddedObjectMeta=true`.

This prevents Kubernetes from accepting invalid values in these metadata fields, such as reported in #1171. When trying to apply a resource with such an invalid value, an error like this will be shown:

```
The SealedSecret "my-secret" is invalid: spec.template.metadata.labels.hi: Invalid value: "boolean": spec.template.metadata.labels.hi in body must be of type string: "boolean"
```

It also replaces the deprecated tag `+kubebuilder:validation:XPreserveUnknownFields` with the newer `+kubebuilder:pruning:PreserveUnknownFields`.

**Applicable issues**
- #1171

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -55,7 +55,7 @@ generate: $(GO_FILES)
 	cp -r gentmp/github.com/bitnami-labs/sealed-secrets/pkg . && rm -rf gentmp/
 
 manifests:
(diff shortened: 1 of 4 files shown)


alvneiayu (approved):
LGTM

PR #1182, opened 2023-04-17
Fix kubeseal image signing for cosign v2

alemorcuq:
**Description of the change**
Update signing command for the Kubeseal image to be compatible with Cosign v2


agarcia-oss (approved):
LGTM!

PR #1192, opened 2023-04-19
Fix go generate to use module dep instead of vendor dir

mkmik:
**Description of the change**

We used to vendor our dependencies but then we got rid of vendoring in favour of using modern go modules and module cache.

However the `go generate` targets were still referencing the vendoring dir in order to execute code generation scripts for sealed-secrets CRD.

This PR fixes the go:generate stanza so that it locates the script from go module directory.

(Forcing a `go download` before `go list` is necessary to ensure that the `Dir` field is populated. If the cache is warm go download is a no-op).

**Benefits**

`go generate ./...` now works

**Possible drawbacks**

**Applicable issues**

- fixes #

**Additional information**


alvneiayu (approved):
LGTM, thanks a lot @mkmik

josvazg (approved):
Thanks!

Issue #396, opened 2020-04-19
INFO logs are sent to stderr


yegorka6:
It seems that many INFO logs are sent to stderr instead of stdout. Please see attached example where successful message is sent to stderr, we receive this on all successful unseal operations.
We do not have any log parsing mechanism between our cluster in GCP and the Datadog, so the logs are streaming directly and the log level is not altered. 

![image](https://user-images.githubusercontent.com/19820931/79682183-e4f84500-8228-11ea-9de2-ae4f0b8d2cc0.png)

PR #1195, opened 2023-04-24
Enable logging info to stdout

cyrgim:
**Description of the change**

This commit introduces a logging package that utilizes https://pkg.go.dev/log as before, but now includes `Infof` and `Errorf` methods. The controller's code has been updated to use these methods. When the `log-info-stdout` flag is enabled, `Infof` logs will be directed to stdout instead of stderr. By default, the controller continues to log information to stderr to maintain compatibility with existing behavior.

**Known limitations**

- You may prefer to switch to a different logging package rather than using a custom one.
- By default, informational messages are still logged to stderr.

**Benefits**

The native Golang log library logs by default to stderr with Println, which is suitable for most messages logged by the controller, as they are usually warnings or errors. However, in some instances, informational messages are logged, which also go to stderr. This can cause issues when using log management tools.

**Applicable issues**

#396

### DIFF
--- a/cmd/controller/main.go
+++ b/cmd/controller/main.go
@@ -49,6 +49,8 @@ func bindControllerFlags(f *controller.Flags, fs *flag.FlagSet) {
 
 	fs.BoolVar(&f.SkipRecreate, "skip-recreate", false, "if true the controller will skip listening for managed secret changes to recreate them. This helps on limited permission environments.")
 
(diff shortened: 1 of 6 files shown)


alvneiayu (approved):
LGTM

alvneiayu:
thanks a lot for your contribution @cyrgim

Issue #1199, opened 2023-04-29
An option to keep Secret when SealedSecret is removed


longquan0104:
**Which component**:
Controller@latest

**Describe the solution you'd like**
Add an annotation that k8s gc knows not to remove the Secret when SealedSecret is removed.

agarcia-oss:
Hi @longquan0104  I'm going to proceed to close this issue since it was already addressed on the #1200 PR.

PR #1200, opened 2023-04-29
Feature allow to skip set owner references

longquan0104:
**Description of the change**
I met an edged use case that I need to keep the _**Secret**_ resource when I delete the _**SealedSecret**_ resource.
Add an annotation that allows a user to skip set **_ownerReferences_**.  

**Benefits**

This prevent the k8s GC to remove the Secret when the _**SealedSecret**_ is removed

**Possible drawbacks**

**Applicable issues**

- adds #1199

**Additional information**

### DIFF
--- a/README.md
+++ b/README.md
@@ -449,6 +449,11 @@ only change from existing Kubernetes is that the *contents* of the
 
 If you want `SealedSecret` controller to take management of an existing `Secret` (i.e. overwrite it when unsealing a `SealedSecret` with the same name and namespace), then you have to annotate that `Secret` with the annotation `sealedsecrets.bitnami.com/managed: "true"` ahead applying the [Usage](#usage) steps.
 
+
+### Seal secret which can skip set owner references
+
+If you want `SealedSecret` and the `Secret` to be independent (which mean when you delete the `SealedSecret` the `Secret` won't disappear with). then you have to annotate that Secret with the annotation `skip-set-owner-references/skip-set-owner-references: "true"` ahead applying the Usage steps. This still add `sealedsecrets.bitnami.com/managed: "true"` to your `Secret` so that your secret will be updated when `SealedSecret` is updated. 
+
 ### Update existing secrets
 
 If you want to add or update existing sealed secrets without having the cleartext for the other items,
--- a/README.md
+++ b/README.md
@@ -452,7 +452,7 @@ If you want `SealedSecret` controller to take management of an existing `Secret`
 
 ### Seal secret which can skip set owner references
 
-If you want `SealedSecret` and the `Secret` to be independent (which mean when you delete the `SealedSecret` the `Secret` won't disappear with). then you have to annotate that Secret with the annotation `skip-set-owner-references/skip-set-owner-references: "true"` ahead applying the Usage steps. This still add `sealedsecrets.bitnami.com/managed: "true"` to your `Secret` so that your secret will be updated when `SealedSecret` is updated. 
+If you want `SealedSecret` and the `Secret` to be independent, which mean when you delete the `SealedSecret` the `Secret` won't disappear with it, then you have to annotate that Secret with the annotation `sealedsecrets.bitnami.com/skip-set-owner-references: "true"` ahead of applying the Usage steps. You still may also add `sealedsecrets.bitnami.com/managed: "true"` to your `Secret` so that your secret will be updated when `SealedSecret` is updated. 
 
 ### Update existing secrets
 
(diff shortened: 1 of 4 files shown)


josvazg:
Hi @longquan0104 could you explain in a bit more detail why would this be useful?
What kind of use case justifies this requirement that you want the secret not be tied to its generating sealed secret?

longquan0104:
Hello @josvazg, I'm meeting the case that when my hook function runs, it removes the old SealedSecret before adding a new SealedSecret to update the Secret. When removing the sealed secret, the secret is removed cause some other resource that uses values in the Secret that can not work anymore.

josvazg:
> Hello @josvazg, I'm meeting the case that when my hook function runs, it removes the old SealedSecret before adding a new SealedSecret to update the Secret. When removing the sealed secret, the secret is removed cause some other resource that uses values in the Secret that can not work anymore.

Why does your hook need to remove the prior sealed secret?

longquan0104:
> > Hello @josvazg, I'm meeting the case that when my hook function runs, it removes the old SealedSecret before adding a new SealedSecret to update the Secret. When removing the sealed secret, the secret is removed cause some other resource that uses values in the Secret can not work anymore.
> 
> Why does your hook need to remove the prior sealed secret?

Because before any actions, Hooks always remove previous resources as a default policy which is mentioned here.
https://helm.sh/docs/topics/charts_hooks/#hook-deletion-policies
So I think this will help Secret to have the option to make itself orphaned to prevent being deleted with SealedSecret.

josvazg:
So this seems an issue when using helm hooks. In such case maybe this will help without changes to sealed secrets helm chart?
https://helm.sh/docs/topics/charts_hooks/#hook-resources-are-not-managed-with-corresponding-releases

> The resources that a hook creates are currently not tracked or managed as part of the release. Once Helm verifies that the hook has reached its ready state, it will leave the hook resource alone. Garbage collection of hook resources when the corresponding release is deleted may be added to Helm 3 in the future, so any **hook resources that must never be deleted should be annotated with `helm.sh/resource-policy: keep`.**

Would that solve your issue?

longquan0104:
helm.sh/resource-policy: keep does not work. I tried :(
Thank you. 😆

josvazg:
I did not get it working, BTW could you add a small section to the README explaining how to use it?

I got this for instance:
```yaml
apiVersion: v1
data:
  foo: YmFy
kind: Secret
metadata:
  annotations:
    sealedsecrets.bitnami.com/skip-set-owner-references: "true"
  creationTimestamp: "2023-05-22T17:10:47Z"
  name: mysecret
  namespace: default
  ownerReferences:
    controller: true
    kind: SealedSecret
    name: mysecret
    uid: 5d73d76a-09b1-432a-92ab-758143b01d70
  resourceVersion: "3266"
  uid: 477bb745-e252-4ce8-b039-fa7ec058fb5d
type: Opaque
```

The annotation is there, yet the ownership reference is also there.

I believe the reason might be that if you place the annotation in the template/original secret, it will be moved under `template/metadata` in the sealedsecret, so your code will not find it there.

I suggest documenting the feature and making sure that it also works as well if you annotate the `secret` rather than the `sealedsecret` itself: Some users might be expecting that. For instance, they might use a secret template from which they expand a temporary secret that they use to create the sealed secret from. In that case they should be able to annotate the input secret and leave the `sealedsecret` as generated by `kubeseal`.

josvazg:
BTW, one thing to note is if you do this then the secret is not owned, obviously, which means that changes in the sealed secrets will not update the secrets.

If you would go ahead we would ask you to document also the relationship with the _managed_ annotation.

Also, as an alternative you can just edit the secret and remove the ownership, before the hook does its thing.

README.md
```diff
@@ -449,6 +449,11 @@ only change from existing Kubernetes is that the *contents* of the
 
 If you want `SealedSecret` controller to take management of an existing `Secret` (i.e. overwrite it when unsealing a `SealedSecret` with the same name and namespace), then you have to annotate that `Secret` with the annotation `sealedsecrets.bitnami.com/managed: "true"` ahead applying the [Usage](#usage) steps.
 
+
+### Seal secret which can skip set owner references
+
+If you want `SealedSecret` and the `Secret` to be independent (which mean when you delete the `SealedSecret` the `Secret` won't disappear with). then you have to annotate that Secret with the annotation `skip-set-owner-references/skip-set-owner-references: "true"` ahead applying the Usage steps. This still add `sealedsecrets.bitnami.com/managed: "true"` to your `Secret` so that your secret will be updated when `SealedSecret` is updated.
```
josvazg, line 455:
```suggestion
If you want `SealedSecret` and the `Secret` to be independent, which mean when you delete the `SealedSecret` the `Secret` won't disappear with it, then you have to annotate that Secret with the annotation `sealedsecrets.bitnami.com/skip-set-owner-references: "true"` ahead of applying the Usage steps. You still may also add `sealedsecrets.bitnami.com/managed: "true"` to your `Secret` so that your secret will be updated when `SealedSecret` is updated. 
```

longquan0104: So `sealedsecrets.bitnami.com/managed: "true"` should be added by the user instead of automatically when Unsealing? If so I should make an update for this.

josvazg: I would keep both tags separate and explicit, yes.

josvazg:
Thanks a lot for for your change!

josvazg (approved):

PR #1227, opened 2023-05-25
Fix doc generated code directory

josvazg:
**Description of the change**

Avoid the logic of copying the generated code to the proper directory. Instead use the flag `--trim-path-prefix` to generate the code in the right place and without the unwanted prefix directories.

https://github.com/kubernetes/code-generator/blob/db4dff4b8168dd37724b89025de5f5be71db6a4f/cmd/go-to-protobuf/protobuf/cmd.go#L99

I also fixed some typos.

**Benefits**

Now `go generate ./...` just does the whole thing properly, including leaving files at the right place.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -37,7 +37,6 @@
 /docker/controller
 *.iml
 .idea
-gentmp/
 
 # GoReleaser output dir
 dist/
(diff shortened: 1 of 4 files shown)


josvazg:
Seems generating the code again would add some extra files and include some changes, I would like to add those as a follow up PR.

alvneiayu (approved):
LGTM

PR #1228, opened 2023-05-29
Update generated code

josvazg:
**Description of the change**

Result of `make generate` / `go generate ./...`.

Seems latest code generation library version produces code for `applyconfiguration`. As we do not use or need it, we moved our generation spec to explicitly enumerate the generators we require: `deepcopy,client,informer,lister`.

**Benefits**

Generated code is in line with the latest code generation library.

Signed-off-by: Jose Luis Vazquez Gonzalez <josvaz@vmware.com>

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -20,6 +20,7 @@ require (
 	k8s.io/code-generator v0.27.2
 	k8s.io/klog v1.0.0
 	k8s.io/klog/v2 v2.100.1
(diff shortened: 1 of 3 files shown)


alemorcuq (approved):

alvneiayu (dismissed):
LGTM

alemorcuq (approved):

Issue #1142, opened 2023-03-09
Add option to replace Cluster Roles by namespaced Roles


josvazg:
**Which component**:
Controller Helm Chart

**Is your feature request related to a problem? Please describe.**

Right now the following [capabilities are defined by default as Cluster wide Roles](https://github.com/bitnami-labs/sealed-secrets/blob/main/helm/sealed-secrets/templates/cluster-role.yaml):
- Monitor (get/list/watch) sealed secrets.
- Update sealed secrets' status.
- Fully manage (get/list/create/update/delete/watch) secrets.
- Emit (create/patch) events.

Cluster wide roles are problematic, as not all environments allow end users to rely on them.

For instance, multi-tenant environments do not allow for Cluster Roles or impose important restrictions to include custom permissions as one off special requests.

**Describe the solution you'd like**

For such environments we want instead the controller to have regular Roles on each of the managed namespaces.

**Additional context**

This was brought up by https://github.com/bitnami-labs/sealed-secrets/issues/1064 and replaces https://github.com/bitnami-labs/sealed-secrets/issues/1092

alvneiayu:
solved with #1240

PR #1240, opened 2023-06-15
Add namespaced roles support to Helm chart

kvandenhoute:
Signed-off-by: Karel Vanden Houte <karel.vandenhoute@gmail.com>

**Description of the change**
This PR solves https://github.com/bitnami-labs/sealed-secrets/issues/1142
Add support for using Roles instead of ClusterRoles, based on additionalNamespaces

**Benefits**
Cluster wide roles are problematic, as not all environments allow end users to rely on them.

For instance, multi-tenant environments do not allow for Cluster Roles or impose important restrictions to include custom permissions as one off special requests.

This change allows to *not* use ClusterRoles, but Roles for each of the additionalNamespaces instead
**Possible drawbacks**
None, default values still have the same behavior as before.

**Applicable issues**
#1142
- fixes #

**Additional information**


alvneiayu (approved):
this is an incredible improvement and contribution @kvandenhoute. Thanks a lot for your contribution and help.

Issue #1205, opened 2023-05-08
Option to not-clobber existing secret, opting to patch it instead. 


avo-sepp:
**Which component**:
Controller

**Is your feature request related to a problem? Please describe.**
Yes, my problem is that I have a case where I need to use Kustomize to PATCH an exiting Kubernetes secret for ArgoCD. ArgoCD has other values in this secret that it populates at runtime. I have Kustomize and the patches saved in Git, but I cannot commit the secret to Git because it is in plaintext. 

I thought I could use Kustomize to instead apply a sealed-secret. However, sealed-secret's default is to clobber/overwrite any existing secret. 

**Describe the solution you'd like**
An annotation/label where I could tell the sealed-secret controller to use a PATCH operation and NOT overwrite any existing secret with the same name. 

**Describe alternatives you've considered**
Alternatives I've considered store the secret outside of Git and apply it with some SIG tool or script.

alemorcuq:
It's an interesting suggesion, @avo-sepp. We will look into it.

alemorcuq:
Hey, @avo-sepp. Could you show us a concrete example of what you have in mind to help us understand this issue? I understand you want to Sealed Secrets to partially update an already existing secret.

tvanderka:
I recently had the same use-case. I want to create a secret with sealedsecret from git, and later have additional keys added by other controllers/operators at runtime. I also do need an option to keep the secret when sealedsecret is deleted.

Possible solution:
sealedsecrets.bitnami.com/managed: "false" -> do NOT recreate&own the existing secret, only update/patch keys from ss
(skip if no "managed" annotation)

For an example look at https://github.com/argoproj/argo-cd/blob/master/docs/operator-manual/argocd-secret.yaml, there are a few "autogenerated" keys I could leave undefined, use sealedsecret only for webhook.github.secret, or add additional accounts.* later. This is currently not possible as ss overwrites it completely.

avo-sepp:
I would like to add another, related, but different option for `sealedsecrets.bitnami.com/clobber: "false"` which would modify the patch operation such that it does not overwrite a value that already exists. In this case the end result of valueB would be `valueB: "bcd"` instead of `"123"`. Which might be a nice addition to this PATCH functionality.

agarcia-oss:
Hi @avo-sepp @tvanderka Thanks for your feedback and examples! We're going to include the `patch` proposal in the project backlog. 
[...]

PR #1259, opened 2023-06-29
Add option to patch secrets instead of clobbering them

alemorcuq:
**Description of the change**
Add a new annotation `sealedsecrets.bitnami.com/patch: "true"` to secrets so that instead of replacing the whole secret with the unencrypted sealed secret, only the keys that are present in the sealed secret are added to the secret while the existing keys in the secret remain there.

In case of keys present in both secret and sealed secret, the second one will take precedence and the original value in the secret will be replaced with the one in the sealed secret. 

When annotated with this annotation, the `SealedSecret` won't take ownership of the `Secret`. When combined with the `sealedsecrets.bitnami.com/managed: "true"` annotation, the `SealedSecret` will take ownership of the `Secret` while the patching behavior is maintained.

**Benefits**
Sealed Secrets can control just some of the keys that are present in the Secret. For example, it can be used to manage the webhook secrets in [argocd-secret](https://github.com/argoproj/argo-cd/blob/master/docs/operator-manual/argocd-secret.yaml)

**Possible drawbacks**
 
**Applicable issues**
- #1205

### DIFF
--- a/README.md
+++ b/README.md
@@ -453,6 +453,8 @@ If you want the Sealed Secrets controller to manage an existing `Secret`, you ca
 
 ### Patching existing secrets
 
+> New in v0.23.0
+
 There are some use cases in which you don't want to replace the whole `Secret` but just add or modify some keys from the existing `Secret`. For this, you can annotate your `Secret` with `sealedsecrets.bitnami.com/patch: "true"`. Using this annotation will make sure that secret keys, labels and annotatations in the `Secret` that are not present in the `SealedSecret` won't be deleted, and those present in the `SealedSecret` will be added to the `Secret` (secret keys, labels and annotations that exist both in the `Secret` and the `SealedSecret` will be modified by the `SealedSecret`).
 
 This annotation does not make the `SealedSecret` take ownership of the `Secret`. You can add both the `patch` and `managed` annotations to obtain the patching behaviour while also taking ownership of the `Secret`.
--- a/README.md
+++ b/README.md
@@ -455,7 +455,7 @@ If you want the Sealed Secrets controller to manage an existing `Secret`, you ca
 
 > New in v0.23.0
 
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -448,8 +449,15 @@ only change from existing Kubernetes is that the *contents* of the
...
+If you want the Sealed Secrets controller to manage an existing `Secret`, you can annotate your `Secret` with the `sealedsecrets.bitnami.com/managed: "true"` annotation. The existing `Secret` will be overwritten when unsealing a `SealedSecret` with the same name and namespace, and the `SealedSecret` will take ownership of the `Secret` (so that when the `SealedSecret` is deleted the `Secret` will also be deleted).
 
+### Patching existing secrets
+
+> New in v0.23.0
+
+There are some use cases in which you don't want to replace the whole `Secret` but just add or modify some keys from the existing `Secret`. For this, you can annotate your `Secret` with `sealedsecrets.bitnami.com/patch: "true"`. Using this annotation will make sure that secret keys, labels and annotatations in the `Secret` that are not present in the `SealedSecret` won't be deleted, and those present in the `SealedSecret` will be added to the `Secret` (secret keys, labels and annotations that exist both in the `Secret` and the `SealedSecret` will be modified by the `SealedSecret`).
```
alvneiayu, line 458:
```suggestion
There are some use cases in which you don't want to replace the whole `Secret` but just add or modify some keys from the existing `Secret`. For this, you can annotate your `Secret` with `sealedsecrets.bitnami.com/patch: "true"`. Using this annotation will make sure that secret keys, labels and annotations in the `Secret` that are not present in the `SealedSecret` won't be deleted, and those present in the `SealedSecret` will be added to the `Secret` (secret keys, labels and annotations that exist both in the `Secret` and the `SealedSecret` will be modified by the `SealedSecret`).
```

alvneiayu (approved):
LGTM

PR #1275, opened 2023-07-20
Introduce KUBESEAL_VERSION for Linux installation

lindhe:
**Description of the change**

This change introduces a variable `KUBESEAL_VERSION` in the README's example of how to install the `kubeseal` tools on Linux. 

**Benefits**

This makes it quicker and easier to just copy-paste the commands in order to install a certain version.

**Additional information**

If you are CI ninjas it might be worth considering some sort of check to make sure that `KUBESEAL_VERSION` stays up to date with the latest stable release number.

> 💡 A good way to inspect this commit is:
> 
> ```shell
> git show --color-words='.'
> ```

### DIFF
--- a/README.md
+++ b/README.md
@@ -379,7 +379,7 @@ nix-env -iA nixpkgs.kubeseal
 The `kubeseal` client can be installed on Linux, using the below commands:
 
 ```bash
-KUBESEAL_VERSION='0.23.0'
+KUBESEAL_VERSION='' # Set this to, for example, KUBESEAL_VERSION='0.23.0'
 wget "https://github.com/bitnami-labs/sealed-secrets/releases/download/v${KUBESEAL_VERSION:?}/kubeseal-${KUBESEAL_VERSION:?}-linux-amd64.tar.gz"
 tar -xvzf kubeseal-${KUBESEAL_VERSION:?}-linux-amd64.tar.gz kubeseal
 sudo install -m 755 kubeseal /usr/local/bin/kubeseal
--- a/README.md
+++ b/README.md
@@ -379,8 +379,9 @@ nix-env -iA nixpkgs.kubeseal
 The `kubeseal` client can be installed on Linux, using the below commands:
 
 ```bash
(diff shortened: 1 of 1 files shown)


lindhe:
Ping

alvneiayu:
thanks a lot for your contribution @lindhe and sorry for the stale message. Reviewing it and thanks again.

alvneiayu (approved):
LGTM

README.md
```diff
@@ -379,8 +379,9 @@ nix-env -iA nixpkgs.kubeseal
 The `kubeseal` client can be installed on Linux, using the below commands:
 
 ```bash
-wget https://github.com/bitnami-labs/sealed-secrets/releases/download/<release-tag>/kubeseal-<version>-linux-amd64.tar.gz
-tar -xvzf kubeseal-<version>-linux-amd64.tar.gz kubeseal
+KUBESEAL_VERSION='0.23.0'
```
alvneiayu, line 382: I only have a concern here. We are setting up to 0.23.0 (it is the actual one) but with the next releases the documentation will be confused and older. Could you set this value to something like '<release-tag>'  or '<latest-release-tag>' just to avoid this situation?

alvneiayu (changes requested):

README.md
lindhe: That's absolutely a valid concern! Stale documentation is no bueno.

I could either change it to a non-value like you suggest, e.g. like `KUBESEAL_VERSION=''` or `KUBESEAL_VERSION='CHANGEME'`.

Or I could expand the example to one where it always fetches the latest version and one where it fetches a pinned version.

Which one do you prefer?

alvneiayu: KUBESEAL_VERSION='CHANGEME' is a good option. thanks a lot @lindhe

alvneiayu (approved):
LGTM

README.md
lindhe: I changed my mind. I think `KUBESEAL_VERSION=''` is better than `KUESEAL_VERSION='CHANGEME'`, since the former will be caught by `${KUBESEAL:?}` while the latter will result in `wget` trying to download the `CHANGEME` version. :) Also added an example as a comment to make it clear exactly what the user should set the variable to.

PR #1283, opened 2023-07-27
fix(chart): create servicemonitor only if apiVersion is available

nobbs:
**Description of the change**

This will prevent the chart from failing to install if ServiceMonitor is enabled but the Prometheus Operator is not installed in the cluster. Tiny change adding a check for `monitoring.coreos.com/v1` API capability.

**Benefits**

Installation of Helm chart will silently ignore missing Prometheus Operator / ServiceMonitor CRD even if enabled in values file.

**Possible drawbacks**

None known.

**Applicable issues**

None

**Additional information**

None

### DIFF
--- a/helm/sealed-secrets/templates/servicemonitor.yaml
+++ b/helm/sealed-secrets/templates/servicemonitor.yaml
@@ -1,4 +1,4 @@
-{{- if .Values.metrics.serviceMonitor.enabled }}
+{{- if and .Values.metrics.serviceMonitor.enabled (.Capabilities.APIVersions.Has "monitoring.coreos.com/v1") }}
 apiVersion: monitoring.coreos.com/v1
 kind: ServiceMonitor
 metadata:


helm/sealed-secrets/templates/servicemonitor.yaml
```diff
@@ -1,4 +1,4 @@
-{{- if .Values.metrics.serviceMonitor.enabled }}
+{{- if and .Values.metrics.serviceMonitor.enabled (.Capabilities.APIVersions.Has "monitoring.coreos.com/v1") }}
```
alvneiayu, line 1: hi @nobbs 

first of all, thanks a lot for your contribution.

I only have a concern with this.  This is going to affect the `helm template`. If you don't have the coreos installed in your cluster (current context cluster), the chart is not going to generate the template at all.  Imagine that one user wants to generate the template to apply it in another cluster and they are setting up the serviceMonitor to enable, the template will be generate without (if they don't have the resource included) the serviceMonitor (silently). The user will apply the template and they will see that the serviceMonitor is not installed (if you don't have coreos) but they are setting up the serviceMonitor to enable, it will be quite confused. 

I think that maybe it is a better idea just to show a message (though a helper or something like this) that you don't have the coreos installed. And I think that we need to let the users to generate the template or install Sealed Secrets without letting know if an external resource is installed or not. The users will see the chart crashing and they will see that they need to include it (or they will see a warning message as I am suggesting before).

Thanks a lot again

Álvaro

alvneiayu (changes requested):

PR #1286, opened 2023-08-03
feat: allow changing the default revisionHistoryLimit

tberreis:
**Description of the change**

The deployment has been adapted to allow the user changing the default revisionHistoryLimit.

**Benefits**

You may want to clean up old replicasets so that these don't appear in metrics, OPA violations, ArgoCD UI and so on.
Therefore you want to overwrite the Kubernetes default revision history limit.

**Possible drawbacks**

N/A

**Applicable issues**

N/A

**Additional information**

N/A

### DIFF
--- a/helm/sealed-secrets/values.yaml
+++ b/helm/sealed-secrets/values.yaml
@@ -12,9 +12,6 @@ fullnameOverride: ""
 ## @param namespace Namespace where to deploy the Sealed Secrets controller
 ##
 namespace: ""
-## @param revisionHistoryLimit Number of old history to retain to allow rollback (If not set, default Kubernetes value is set to 10)
-## e.g:
-## revisionHistoryLimit: 1
 
 ## @param extraDeploy [array] Array of extra objects to deploy with the release
 ##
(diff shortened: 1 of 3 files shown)


helm/sealed-secrets/values.yaml
```diff
@@ -12,6 +12,10 @@ fullnameOverride: ""
 ## @param namespace Namespace where to deploy the Sealed Secrets controller
 ##
 namespace: ""
+## @param revisionHistoryLimit Number of old history to retain to allow rollback (If not set, default Kubernetes value is set to 10)
+## e.g:
+## revisionHistoryLimit: 1
```
alemorcuq, line 17:
```suggestion
revisionHistoryLimit: ""
```

I would also move this a bit down, under the "Sealed Secrets Parameters" section.

Could you please add this parameter to the readme too?

alemorcuq (changes requested):
Thanks for your contribution, @tberreis ! Please, check my comment.

helm/sealed-secrets/values.yaml
tberreis: Thank you for your feedback @alemorcuq. Code has been adapted.

alemorcuq (approved):
Thanks!

PR #1294, opened 2023-08-18
Update Linux installation process on README to have a way to dynamically get kubeseal version number

alexjyong:
**Description of the change**

Updating the README.md to show how to get the version of kubeseal dynamically instead of having to hardcode it for the Linux install script

**Benefits**

Users don't have to go to the release page to find the latest version

Install script can be used in CI/CD environments or other environments where kubeseal may be dynamically installed and this will allow the version of the latest to be automatically installed

**Possible drawbacks**

Adds extra detail to the README. I've maintained the original way to do it in the event the user using this doesn't have `jq` and `curl` installed, which is required for this method and cannot/does not want to install it.

**Applicable issues**

N/A

**Additional information**
N/A

### DIFF
--- a/README.md
+++ b/README.md
@@ -385,7 +385,24 @@ tar -xvzf kubeseal-${KUBESEAL_VERSION:?}-linux-amd64.tar.gz kubeseal
 sudo install -m 755 kubeseal /usr/local/bin/kubeseal
 ```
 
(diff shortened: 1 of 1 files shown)


alvneiayu (approved):
LGTM

Issue #223, opened 2019-08-20
Secrets are frequently unsealed


goober:
Hi. We have recently upgraded the sealed-secrets-controller to v0.8.1 from v0.7 and are experiencing that all sealed secrets are getting unsealed frequently without any modifications to the sealed secret entity.

![image](https://user-images.githubusercontent.com/124244/63360388-dbe36b00-c35d-11e9-8000-eb971ac4264e.png)

Is this an expected behavior?

Worth noting is that we are runnnig in an Openshift environment.

mkmik:
before v0.8.x we hadn't events in the first place so you wouldn't have noticed it reported that way in the first place.

goober:
Thanks for the answer, that explains the behavior, and what I could read from the code is how it has worked for a long time, even though it has not been exposed in this way before 0.8. 

Is this how you want it to work or should the unsealed event only be triggered when the sealed secret object has actually changed since the last unseal? I can see arguments for both cases.

mkmik:
I'm not a fan of the noise so a fix would be welcome (although I don't think it's an urgent issue)

alfsch:
We're using ArgoCD to deploy our applications and sealed secrets. The noise of unseal events every 5 seconds triggers refreshes on argocd application controller, what leads to a very huge CPU load. The main problem is, that it raises with every additional sealed secret, which makes it quite unusable for us.

alfsch:
During my investigations about this topic I've seen, that when the status field is activated we get an incredible cpu load in sealed secret operator and argocd's application controller due to frequently unsealing sealed secrets (each secret is unsealed about every 5 seconds).

Sealed Secrets Operator:
![Status Field on](https://user-images.githubusercontent.com/6358461/105729597-b46ad780-5f2d-11eb-9709-6f926532784e.png)
![Status Field off](https://user-images.githubusercontent.com/6358461/105729803-e2e8b280-5f2d-11eb-9268-b78817ba6289.png)

ArgoCD Application Controller
![Status Field on](https://user-images.githubusercontent.com/6358461/105732247-86d35d80-5f30-11eb-9038-700cf16f553c.png)
![Status Field off](https://user-images.githubusercontent.com/6358461/105732288-95217980-5f30-11eb-84d7-73550ce57709.png)

The Screenshots are showing how cpu usage is reduced by switching off the status field.

mkmik:
Does this only happen if the status feature flag is enabled?

alfsch:
[...]

PR #1295, opened 2023-08-20
Fix issue where sealed secrets status is not updated if sealed secret…

### LINKED ISSUE #853: Status shows `no key could decrypt secret` for successful created secret
**Which component**:
kubeseal: v0.17.4
sealed-secret-controller: v0.17.5
kubernetes: 1.22.6

**Describe the bug**
The `status` feature, where you can see if the creation of a secret through the controller was successful or not is partly not correct. In this case the status says `message: no key could decrypt secret (<key>)` but the `logs` and `events` of the controller says `SealedSecret unsealed successfully`.

Screen Shot 2022-06-01 at 18 38 26

**To Reproduce**
- upgrade the controller from v0.12.4 to v0.17.5
- Restart the pod
- compare `status` and `controller log` or `events` entries

**Expected behavior**
When the creation was successful the `status` should show `SealedSecret unsealed successfully` like the logs and the `events`.

**Version of Kubernetes**:

```
Client Version: version.Info{Major:"1", Minor:"24", GitVersion:"v1.24.0", GitCommit:"4ce5a8954017644c5420bae81d72b09b735c21f0", GitTreeState:"clean", BuildDate:"2022-05-03T13:46:05Z", GoVersion:"go1.18.1", Compiler:"gc", Platform:"darwin/amd64"}
Kustomize Version: v4.5.4
Server Version: version.Info{Major:"1", Minor:"22", GitVersion:"v1.22.6", GitCommit:"07959215dd83b4ae6317b33c824f845abd578642", GitTreeState:"clean", BuildDate:"2022-03-30T18:28:25Z", GoVersion:"go1.16.12", Compiler:"gc", Platform:"linux/amd64"}
WARNING: version difference between client (1.24) and server (1.22) exceeds the supported minor version skew of +/-1
```

**Additional context**
I am not sure if i misunderstood the `status` feature 🤔.

mowirth:
**Description of the change**

Removes a check used to prevent unnecessary updates of the sealed secrets status.

**Benefits**

The check is used to prevent unnecessary updates on the sealed secrets status. 

However, in some cases, the sealed secret might not be updated to successfully generate the secret.
For example, if a secret is managed by a different controller, the generation from a sealedSecret would fail; if the controller is removed later, the secret would be unsealed successfully.
In this case, the check prevents a successful update leading to issues when the secret is successfully unsealed (and the successful unseal is reported), but the secret status is not updated since the generation/observedGeneration steps were not updated.

This is also an issue when using management tools such as ArgoCD, since the status would always report an error, while the secret is successfully unsealed.

**Possible drawbacks**

May increase updates of sealed secrets status if not necessary.

**Applicable issues**

- fixes #853 

**Additional information**

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -423,7 +423,8 @@ func (c *Controller) updateSealedSecretStatus(ssecret *ssv1alpha1.SealedSecret, 
 	return nil
 }
 
-func updateSealedSecretsStatusConditions(st *ssv1alpha1.SealedSecretStatus, unsealError error) (updateRequired bool) {
+func updateSealedSecretsStatusConditions(st *ssv1alpha1.SealedSecretStatus, unsealError error) bool {
+	var updateRequired bool
 	cond := func() *ssv1alpha1.SealedSecretCondition {
 		for i := range st.Conditions {
 			if st.Conditions[i].Type == ssv1alpha1.SealedSecretSynced {
@@ -453,7 +454,7 @@ func updateSealedSecretsStatusConditions(st *ssv1alpha1.SealedSecretStatus, unse
 		updateRequired = true
 	}
 
-	return
+	return updateRequired
 }
 
 func isAnnotatedToBeManaged(secret *corev1.Secret) bool {
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -414,13 +414,16 @@ func (c *Controller) updateSealedSecretStatus(ssecret *ssv1alpha1.SealedSecret, 
 	}
 
 	ssecret.Status.ObservedGeneration = ssecret.ObjectMeta.Generation
-	updateSealedSecretsStatusConditions(ssecret.Status, unsealError)
+	updatedRequired := updateSealedSecretsStatusConditions(ssecret.Status, unsealError)
+	if updatedRequired {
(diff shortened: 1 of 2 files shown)


alvneiayu:
hi @mowirth, 
First of all, thanks a lot for your PR.

I have some concerns about this solution because these lines that you are deleting will generate other issue already fixed in the past and also we are going to increase the CPU a lot (https://github.com/bitnami-labs/sealed-secrets/issues/223).

In my opinion, we need to find an alternative way to solve this problem.

Thanks a lot again

Álvaro

mowirth:
Hi @alvneiayu ,

I added a check to only send an update if the status was changed - this way, a status change is still propagated, but only when necessary. 

Best regards and thanks,
Moritz

alvneiayu:
hi @mowirth 

thanks a lot for the new modifications. I am testing it, I will come back ASAP with some feedback.

Álvaro

alvneiayu:
hi @mowirth 

I tested it and it seems that the performance is OK. We have several tests included inside the controller_test.go file to check the Status. Could you take a look if it is possible to include a new test to be sure that future changes (new PRs from other issues) will not generate the same issue resolved?

Thanks a lot again for your PR

Álvaro

mowirth:
@alvneiayu I added some tests to check whether the expected behaviour regarding the update is applied and that an errorneous update with a previously successfully unsealed sealed secret properly updates the condition.

alvneiayu (approved):
LGTM, and thanks a lot @mowirth for your awesome contribution

PR #1315, opened 2023-09-06
fix: explicit service account tokens

keithharvey:
### Issue:
While working on a Civo/K3s cluster, we encountered issues related to the sealed-secrets controller failing to initialize properly. Although we're not sure if this issue is cluster-specific, it aligns with changes made to the TokenRequest API in Kubernetes 1.24 concerning short-lived tokens.

### Solution:
The Helm chart has been updated to explicitly specify a service account token. This is done by manually creating a Secret containing the API access token for the service account and mounting it to the Pod via a projected volume.

### Why This Change Was Necessary:
The change aims to proactively handle Kubernetes 1.24's move to short-lived tokens, especially as it affects service accounts on Civo/K3s clusters. Explicitly specifying the service account token seems to solve the issue and may be beneficial for others facing similar problems.

### Disclaimer:
This issue was encountered on a Civo/K3s cluster. We are not entirely sure if the problem is unique to this specific setup, but the fix works in our case and hopefully can help others.

### Testing:
1. Deploy the updated Helm chart on a Civo/K3s Kubernetes 1.24+ cluster.
2. Verify that the sealed-secrets controller initializes without errors.
3. Examine logs to confirm that issues related to service account tokens are resolved.

### Failure Case:
The failure case generally manifests as [errors in logs](https://stackoverflow.com/questions/69038012/mountvolume-setup-failed-for-volume-kube-api-access-fcz9j-object-default) regarding missing `kube-root-ca` ConfigMap when relying on the default service account token.

### Checklist:
- [x] Helm Chart updated
- [x] Test cases updated
- [x] Documentation updated to include this change

### Caveat
Turning it on for everyone seemed technically simplest, but do we need a more sophisticated way to enable/disable token creation?

### DIFF
--- a/helm/sealed-secrets/templates/_helpers.tpl
+++ b/helm/sealed-secrets/templates/_helpers.tpl
@@ -48,6 +48,15 @@ Create the name of the service account to use
 {{- end -}}
 {{- end -}}
 
+
+{{/*
+Create the name of the service account token to use
+*/}}
+{{- define "sealed-secrets.serviceAccountTokenName" -}}
+    {{- $serviceAccountName := include "sealed-secrets.serviceAccountName" . -}}
+    {{- printf "%s-token" $serviceAccountName | default .Values.serviceAccount.tokenName -}}
+{{- end -}}
+
 {{/*
 Kubernetes standard labels
 */}}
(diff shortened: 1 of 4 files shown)


liggitt:
Rather than revert to API-object based tokens, would it be possible to determine the root cause of the error mentioned? That issue would cause problems for every pod in the cluster by default, so I wouldn't recommend working around it by switching all workloads back to less secure credentials.

keithharvey:
Yeah it was the api server erroring, restarting it alleviated my issue.

Issue #1296, opened 2023-08-21
`kubeseal` adds trailing newline to YAML output


lindhe:
**Which component**:

`kubeseal`

**Describe the bug**

When creating a sealed secret (either a brand new one or amending to an existing one) with `kubeseal -o yaml`, the output has a trailing newline. This is considered bad practice by many, e.g. by Git (and maybe by the POSIX standard too, I'm not sure). Also, the corresponding JSON output is free of trailing newlines.

**To Reproduce**

```shell
kubectl create secret generic foo --dry-run=client -o yaml --from-literal=foo=bar | kubeseal -o yaml
```

**Expected behavior**

I expect there to be no trailing newline.

**Version of Kubernetes**:

n/a

**Additional context**

n/a

lindhe:
I assume this is where it gets added:

https://github.com/bitnami-labs/sealed-secrets/blob/51eb3fb110204303ce4e73538e0e24a9485bc9d6/pkg/kubeseal/kubeseal.go#L353

But I also assume that it's there for a reason. Could we do something like `.trim()` to the object before printing it?

lindhe:
This was resolved as of 0.24.0 (although #1330 was introduced instead).

PR #1331, opened 2023-09-22
fix: remove trailing dashes for multidoc yaml

### LINKED ISSUE #1330: `kubeseal` appends extra document separator `---` when format is YAML
**Which component**:
kubeseal (v0.24.0)

**Describe the bug**
When sealing a secret with `kubeseal`, an extra [YAML document separator](https://yaml.org/spec/1.0/#id2561718) (`---`) is added to the output (since version `v0.24.0`). This causes YAML parsers to fail or parse an extra empty document.

**To Reproduce**

1. Create a simple secret file `secret.yml`
```yaml
apiVersion: v1
kind: Secret
metadata:
  name: test-secret
type: Opaque
stringData:
  foo: bar
```
2. Run `kubeseal --format yaml --secret-file secret.yaml --cert cert.pem`

The resulting YAML document will contain an extra document separator at the end:

```yaml
kind: SealedSecret
[..]
spec:
  [..]
  template:
    [..]
    type: Opaque
---
```

**Expected behavior**
The resulting YAML document MAY contain a document separator **before** each document. The resulting YAML document MUST NOT contain a trailing document seperator.

**Additional context**
This was most likely caused by https://github.com/bitnami-labs/sealed-secrets/pull/1304 , specifically https://github.com/bitnami-labs/sealed-secrets/pull/1304/files#diff-92f00e9d744fba4dee224973d289d4a24568a3b50a98996622e5928e86dfca04R409 .

This could be solved by putting the document separator in front of every document (instead of the end).

theurichde:
Fixes #1330 #1296

**Description of the change**

While introducing multi document support for JSON and YAML, we introduced trailing dashes for YAMLs, which violates the YAML specification. This change will separate JSON and YAML with `\n` for JSON and `---` for YAML, only if there are multiple documents given. It will not append a new line or dashes for the last document

**Benefits**

YAML specification is met, linters will be happy.

**Possible drawbacks**

**Applicable issues**

- fixes #1330 #1296 

**Additional information**

### DIFF
--- a/pkg/kubeseal/kubeseal.go
+++ b/pkg/kubeseal/kubeseal.go
@@ -226,7 +226,7 @@ func Seal(clientConfig ClientConfig, outputFormat string, in io.Reader, out io.W
 		return err
 	}
 
-	for _, secret := range secrets {
+	for pos, secret := range secrets {
 		if len(secret.Data) == 0 && len(secret.StringData) == 0 && !allowEmptyData {
 			return fmt.Errorf("secret.data is empty in input Secret, assuming this is an error and aborting. To work with empty data, --allow-empty-data can be used")
 		}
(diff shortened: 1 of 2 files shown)


tewfik-ghariani:
Or maybe you could simply add the `---` at the beginning of every object for YAML format ^^

theurichde:
> Or maybe you could simply add the `---` at the beginning of every object for YAML format ^^

Yeah, sure. That would be the more pragmatic way. I thought of having it clean from the beginning, with only a separator in between. But having just the dashes upfront every document, would be also fine for me.

What do you think @alemorcuq @agarcia-oss @alvneiayu ?

lindhe:
Having `---` at the beginning of every YAML document is considered good practice by the the third party tool [yamllint](https://github.com/adrienverge/yamllint): https://yamllint.readthedocs.io/en/stable/configuration.html#default-configuration

theurichde:
Let's go the pragmatic way. There is also muuuch less code changes involved.

theurichde:
I will close this PR in favor of #1335

PR #1334, opened 2023-09-27
Updated command for restoring from backups

leonseng:
**Description of the change**

Updated target label for selecting sealed secret controller pod to be deleted, when restoring from backup. 

**Benefits**

Documentation updated to provide correct procedure for restoring from backup

### DIFF
--- a/README.md
+++ b/README.md
@@ -743,7 +743,7 @@ To restore from a backup after some disaster, just put that secrets back before 
 
 ```bash
 kubectl apply -f main.key
-kubectl delete pod -n kube-system -l name=sealed-secrets-controller
+kubectl delete pod -n kube-system -l app.kubernetes.io/name=sealed-secrets
 ```
 
 ### Can I decrypt my secrets offline with a backup key?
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -743,7 +743,7 @@ To restore from a backup after some disaster, just put that secrets back before
 
 ```bash
 kubectl apply -f main.key
-kubectl delete pod -n kube-system -l name=sealed-secrets-controller
+kubectl delete pod -n kube-system -l app.kubernetes.io/name=sealed-secrets
```
alvneiayu, line 746: this commands will work only for the chart, we need to keep the other one for the controller.yaml. Could help us to include the commands for the controller.yaml (keeping the original as we have) and including the new commands for the charts (as you are doing right now), please?
example:

```
* Deploying Sealed Secrets using Helm:
COMMANDS

* Deploying Sealed Secrets using the controller.yaml
COMMANDS
```

alvneiayu (changes requested):

README.md
leonseng: Thanks for picking that up. Fixed.

alvneiayu (approved):
LGTM

PR #1366, opened 2023-11-06
Improve examples with a more legible kubeseal command

agarcia-oss:
**Description of the change**

Kubeseal can generate encrypted secrets in several ways, the one presented in the documentation did not use the latest features of the CLI.


alvneiayu (approved):
LGTM

PR #1369, opened 2023-11-08
Expose controller metrics in an isolated port

### LINKED ISSUE #1368: Expose controller metrics in an isolated port
**Describe the issue**

 Right now, the Sealed Secrets controller is exposing its metrics in the same port as the rest of its endpoints. To allow further customization of the metrics endpoint, it should be exposed on a different port.

**Pros:** The users will be able to configure the metrics (which port to expose the metrics, if you want to expose it or not, etc)

**Cons:** The programs already consuming the metric endpoint must be reconfigured to consume the new port.

**Expected behavior**

The controller exposes the `/metrics` endpoint on a different port than the rest of its endpoints

alvneiayu:
**Description of the change**

The controller exposes the /metrics endpoint on a different port than the rest of its endpoints

**Benefits**

Metrics will be isolated in a different port.

**Possible drawbacks**

The metrics will be exposed in a different port. The users must change the programs consuming this endpoint to show the metrics.

**Applicable issues**

- fixes #1368

### DIFF
--- a/contrib/prometheus-mixin/README.md
+++ b/contrib/prometheus-mixin/README.md
@@ -1,7 +1,7 @@
 # Sealed Secrets Metrics
 
 The Sealed Secrets Controller running in Kubernetes exposes Prometheus
-metrics on `*:8080/metrics`.
+metrics on `*:8081/metrics`.
 
 These metrics enable operators to observe how it is performing. For example 
 how many `SealedSecret` unseals have been attempted and how many errors may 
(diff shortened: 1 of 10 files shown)


alemorcuq (approved):

bootc:
Sorry, but this MR is incomplete and broken. The new `-metrics` service just points at the normal HTTP port (8080) in the container, so scraping metrics from the service results in a 404. The ServiceMonitor has also not been updated and now picks up _both_ Services, tries to scrape both, and both end up with failed scrapes.

PR #1371, opened 2023-11-10
docs: Correct skipRecreate description

tewfik-ghariani:
**Description of the change**

This message is talking about the case in which **skipRecreate** is actually set to `true`

**Benefits**

Users understand correctly its usefulness

**Possible drawbacks**

_None_


agarcia-oss (approved):
Thanks for fixing this!

PR #1392, opened 2023-11-24
Release the newest Sealed Secrets version using the Bitnami container release

### LINKED ISSUE #1393: Consolidation of Sealed Secrets images in DockerHub
The Sealed Secrets controller's [official image](https://hub.docker.com/r/bitnami/sealed-secrets-controller) is hosted in the [Bitnami repository in DockerHub](https://hub.docker.com/u/bitnami). At the same time, Bitnami produces its own [controller and kubeseal image](https://hub.docker.com/r/bitnami/sealed-secrets) within the same repository. We understand this might create unnecessary confusion among Sealed Secrets users, so we have decided to consolidate both images into a single one.

**What will we do?**

Bitnami updates its images upon detecting new vulnerabilities. Additionally, these images are also packaged using a very compact [scratch base image](https://github.com/bitnami/containers/blob/main/bitnami/sealed-secrets/0/debian-11/Dockerfile#L31) and compiled using golang best practices regarding code optimization. These features make Bitnami images a better option than their upstream counterparts. Our plan is to delegate the image generation process to the Bitnami team, while the Sealed Secrets team focuses on the generation of the kubeseal binaries. 

In turn, the Bitnami team will refactor its current asset (`bitnami/sealed-secrets`) into two standalone images: (`bitnami/sealed-secrets-controller` and `bitnami/sealed-secrets-kubeseal`).

**How is this going to affect Sealed Secrets Users?**

This modification is an internal change in the pipeline to produce & keep up-to-date the Sealed Secrets images. For that reason, the change should be seamless for any Sealed Secrets user. Existing and new Sealed Secrets releases will continue to be available in the project’s [Releases](https://github.com/bitnami-labs/sealed-secrets/releases) section and the official Sealed Secrets container name in DockerHub (`bitnami/sealed-secrets-controller`) will remain unchanged.

**When will the change take effect?**

We expect these changes to be ready during Sealed Secrets' next scheduled release on the 15th of January 2024.

alvneiayu:
With these changes, we are going to release the newest Sealed Secrets version using the containers generated by the Bitnami Container Release process using the tag of with our release. Then, we will generate the official release in our project.

Why this decision? Using the Bitnami Release process, the containers are tested in several environment (OpenShift, AKS, etc) running our integration tests. Like this, the project will be covered in several real environments that Sealed Secrets in running by the users.

From the user side, **you will not notice any change at all**. We are publishing the containers in the same registry and our chart and controller will use the same containers.

fixes #1393

### DIFF
--- a/.github/workflows/release.yaml
+++ b/.github/workflows/release.yaml
@@ -9,11 +9,6 @@ on:
 jobs:
   build:
     runs-on: ubuntu-latest
-    env:
-      controller_dockerhub_image_name: docker.io/bitnami/sealed-secrets-controller
-      controller_ghcr_image_name: ghcr.io/bitnami-labs/sealed-secrets-controller
-      kubeseal_dockerhub_image_name: docker.io/bitnami/sealed-secrets-kubeseal
-      kubeseal_ghcr_image_name: ghcr.io/bitnami-labs/sealed-secrets-kubeseal
     steps:
       # Checkout and set env
       - name: Checkout
@@ -22,6 +17,10 @@ jobs:
         run: |
           source $GITHUB_WORKSPACE/versions.env
           echo "GO_VERSION=$GO_VERSION" >> $GITHUB_ENV
+      - name: Configure Git
+        run: |
+          git config user.name "$GITHUB_ACTOR"
+          git config user.email "$GITHUB_ACTOR@users.noreply.github.com"
       - name: Set up Go
         uses: actions/setup-go@v3.3.1
         with:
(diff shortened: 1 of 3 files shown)


agarcia-oss (approved):

alemorcuq (approved):

sathieu:
> From the user side, you will not notice any change at all. We are publishing the containers in the same registry and our chart and controller will use the same containers.

There is : docker image with tag v0.25.0 doesn't exists (while v0.24.5 does).

alvneiayu:
hi @sathieu.

The new images will be published without the v from now. You can find the image 0.25.0 instead of v0.25.0.

Thanks

Álvaro

PR #1403, opened 2023-12-16
Fix metrics endpoints

### LINKED ISSUE #1402: Incomplete and broken move of metrics to isolated port
**Which component**:
controller 0.24.5

**Describe the bug**
#1369 is a major change to how metrics are exposed in sealed-secrets, and the MR only partially updates the Helm chart to account for the changes. Anyone relying on scraping metrics from Sealed Secrets on port 8080 will see breakage, and anyone using the ServiceMonitor bundled with the Helm chart will also get a nasty surprise.

**To Reproduce**
Install Sealed Secrets. In version 0.24.4 and below the metrics are available on port 8080; in 0.24.5 they are only available on port 8081.

Using the Helm chart, the ServiceMonitor detects *both* of the sealed-secrets Service resources and attempts to scrape both. The `sealed-service-metrics` Service definition is broken and metrics are not available via that service. As a result, no metrics are collected at all from either Service.

**Expected behavior**
Metrics should work. Major breaking changes should not be introduced in a patch release.

**Version of Kubernetes**:

```
Client Version: v1.29.0
Server Version: v1.28.4+rke2r1
```

nvalembois:
**Description of the change**

Adapt helm chart to support #1369:

- set metrics service port and targetPort consistent with deployment metrics port
- set metrics servicemonitor port consistent with metrics service port

**Benefits**

ServiceMonitor is no longer broken.

**Possible drawbacks**

None.

**Applicable issues**

- fixes #1402

**Additional information**

Tested succesfully with quay.io/prometheus-operator/prometheus-operator:v0.70.0.

### DIFF
--- a/helm/sealed-secrets/templates/_helpers.tpl
+++ b/helm/sealed-secrets/templates/_helpers.tpl
@@ -60,25 +60,11 @@ app.kubernetes.io/version: {{ .Chart.AppVersion }}
 app.kubernetes.io/part-of: sealed-secrets
 {{- end -}}
 
-{{- define "sealed-secrets.metrics-labels" -}}
-app.kubernetes.io/name: {{ include "sealed-secrets.name" . }}-metrics
-helm.sh/chart: {{ include "sealed-secrets.chart" . }}
-app.kubernetes.io/instance: {{ .Release.Name }}
-app.kubernetes.io/managed-by: {{ .Release.Service }}
-app.kubernetes.io/version: {{ .Chart.AppVersion }}
-app.kubernetes.io/part-of: sealed-secrets
-{{- end -}}
-
 {{/*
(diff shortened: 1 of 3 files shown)


alvneiayu:
hi @nvalembois 

Thanks a lot for your contribution and to fix the metric issue, please could you signoff your changes? the DCO is failing.

Thanks again

Álvaro

alemorcuq:
Hi, @nvalembois.

Thanks for the PR. Since we want to fix this quickly, I've added a commit leaving just the minimum changes to fix the current issue. I don't like the idea of creating the metrics service only if the service monitor is enabled as some users might want access to the metrics without the need of a service monitor.

alemorcuq (approved):

PR #1404, opened 2023-12-18
add commonLabels to customizable values

### LINKED ISSUE #1373: Implement a mechanism for adding labels to all resources
**Which component**:

The Helm chart.

I'd like to label all my deployed resources, e.g. to easily filter on what resources are used by a certain team or what resources are managed by our CD system, etc.

I would like there to be a `commonLabels` field, similar to the `commonAnnotations` field, in [values.yaml](https://github.com/bitnami-labs/sealed-secrets/blob/main/helm/sealed-secrets/values.yaml) that takes an set of labels and merges them into the labels for all object created by the Helm chart.

Example:

```yaml
commonLabels:
  foo: bar
```

I can do it manually, but that's very tedious.

**Additional context**

n/a

mohamed-essam:
**Description of the change**

Helm Chart only update, add new value `commonLabels` that functions the same as `commonAnnotations` by adding labels to `_helpers.tpl` in `sealed-secrets.labels` 

**Benefits**

* Ability to add custom labels to all resources for any organization compliance

**Possible drawbacks**

N/A

**Applicable issues**

- fixes #1373

**Additional information**

My current organization requires adding a custom label to all resources, thus the PR.
The changes in the PR were tested on a local Kubernetes installation with both `commonLabels: {}` and `commonLabels: {x: 'y'}`

### DIFF
--- a/helm/sealed-secrets/README.md
+++ b/helm/sealed-secrets/README.md
@@ -78,6 +78,7 @@ The command removes all the Kubernetes components associated with the chart and 
 | `namespace`         | Namespace where to deploy the Sealed Secrets controller | `""`  |
 | `extraDeploy`       | Array of extra objects to deploy with the release       | `[]`  |
 | `commonAnnotations` | Annotations to add to all deployed resources            | `{}`  |
+| `commonLabels`      | Labels to add to all deployed resources                 | `{}`  |
 
 ### Sealed Secrets Parameters
 
--- a/helm/sealed-secrets/templates/_helpers.tpl
+++ b/helm/sealed-secrets/templates/_helpers.tpl
@@ -58,6 +58,9 @@ app.kubernetes.io/instance: {{ .Release.Name }}
 app.kubernetes.io/managed-by: {{ .Release.Service }}
 app.kubernetes.io/version: {{ .Chart.AppVersion }}
 app.kubernetes.io/part-of: sealed-secrets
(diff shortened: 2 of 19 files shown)


alemorcuq (changes requested):
Thanks for this. Do you mind using the same approach as with `commonAnnotations`? Let's leave that helper just for Kubernetes standard labels

mohamed-essam:
@alemorcuq Done, I just thought it would be cleaner with this approach, I have another suggested approach of splitting `sealed-secrets.labels` into `sealed-secrets.labels` and `sealed-secrets.kube-labels` where `sealed-secrets.kube-labels` is the standard kube labels, and `sealed-secrets.labels` is the combined common labels, what do you think?
I think this would be cleaner, although a little different from what I usually see in charts

alemorcuq (approved):

alemorcuq:
> @alemorcuq Done, I just thought it would be cleaner with this approach, I have another suggested approach of splitting `sealed-secrets.labels` into `sealed-secrets.labels` and `sealed-secrets.kube-labels` where `sealed-secrets.kube-labels` is the standard kube labels, and `sealed-secrets.labels` is the combined common labels, what do you think?
> I think this would be cleaner, although a little different from what I usually see in charts

Thanks. I like your idea, but let's keep it this way to have it more similar to [Bitnami charts](https://github.com/bitnami/charts)

PR #1419, opened 2023-12-27
Support fetching certificate URL via proxy environment variables

itchyny:
**Description of the change**

This change allows users to specify company private URL for certificate URL which requires their proxy to fetch.

**Benefits**

In our company, since the certificate URL requires our proxy, we always use something like `kubeseal --cert <(curl -sSf "https://example.com/our-cert-file.cert")` (we set `https_proxy` environment variable), or sometimes `wget "https://example.com/our-cert-file.cert" && kubeseal --cert "our-cert-file.cert"`. It would be very useful if we can directly specify the private URL to `--cert`.

**Possible drawbacks**

I don't think there're any drawbacks.

**Applicable issues**

- No issues.

**Additional information**


alvneiayu (approved):
LGTM

Issue #596, opened 2021-05-28
--insecure-skip-tls-verify works only for K8s API connections


Fodoj:
I am fetching a cert from the sealed secrets like this:

```
 kubeseal --raw --insecure-skip-tls-verify=true --from-file=foo.secret --cert https://sealed-secrets.my-ingress.com/v1/cert.pem --namespace team-a --name team-a-foo
```

The ingresses certificate is self signed. I expected `--insecure-skip-tls-verify` to ignore tls verification, but looking at the code of sealed-secrets, this flag is never used inside this project and is probably only passed down to K8s Go client. 

I believe it should be an easy fix though - should this flag be used for both use cases? If yes, I can try to submit a quick fix.

PR #1449, opened 2024-01-28
Check `--insecure-skip-tls-verify` when fetching cert directly

manics:
**Description of the change**

Takes the `--insecure-skip-tls-verify` kubeseal command line flag into account when fetching the certificate.

**Benefits**

If you're using self-signed certificates for a Kubernetes ingress this flag can be used to bypass the certificate verification.

**Possible drawbacks**

This allows someone to bypass the certificate verification which is not recommended, but given that this flag is already supported for k8s server connections it seems reasonable to support it for fetching the certificate outside of the k8s API too.

**Applicable issues**

- fixes https://github.com/bitnami-labs/sealed-secrets/issues/596

**Additional information**

### DIFF
--- a/pkg/kubeseal/kubeseal.go
+++ b/pkg/kubeseal/kubeseal.go
@@ -5,6 +5,7 @@ import (
 	"context"
 	"crypto/rand"
 	"crypto/rsa"
+	"crypto/tls"
 	"encoding/base64"
 	"encoding/json"
 	"errors"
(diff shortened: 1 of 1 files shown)


alemorcuq:
Thanks for your PR, @manics. We will review it later this week. In the meantime, please fix the failed pipeline on DCO.

manics:
This is more complicated than I thought. The `--insecure-skip-tls-verify` is defined by the K8s config, but you can't access the flag directly, you need to construct a K8s config object. I'll close this for now while I think about a solution.

Issue #1448, opened 2024-01-25
Document system:authenticated group usage in GKE


alemorcuq:
We are binding the `system:authenticated` group to the `service-proxier` role so that `kubeseal` can easily communicate with the `controller` to get the public key and encrypt your secrets. This role only has permissions to get the `sealed-secrets` service and to make `POST` and `GET` requests to that service.

In GKE, the use of `system:authenticated` [is discouraged](https://cloud.google.com/kubernetes-engine/docs/best-practices/rbac) because anyone with a Google account is assigned to that group. Here's a recent [KubeCon talk on this topic](https://www.youtube.com/watch?v=PbZbojx4kVM) that mentions our use case.

While nothing sensitive is exposed to users on that group, we should document this in our GKE documention. We should also an option in the chart to allow users to disabled the creation of the `service-proxier` service.

keithjpaulson:
use of system:authenticated is more than discouraged; from google docs
`"To help secure your clusters against mass malware attacks that exploit cluster-admin access misconfigurations, GKE clusters running version 1.28 and later won't allow you to bind the cluster-admin ClusterRole to the system:anonymous user or to the system:unauthenticated or system:authenticated groups."`
I'd guess this will cause issues on GKE if the service-proxier isn't disabled, or group isn't changed?

alemorcuq:
@keithjpaulson:

From your quote, it says that you won't be able to bind the `cluster-admin ClusterRole` to `system:authenticated`. Doing this effectively gives full admin rights to your cluster to anyone with a Google account, that's why Google is disabling it.

We are not doing that. We are binding a custom `Role` that only have permissions to do GET and POST requests to a specific service created by us.

keithjpaulson:
@alemorcuq Correct, I read that part wrong -- they will just flag any use of system:authenticated as potentially insecure in console popups and emails that make administrators like me (who haven't touched sealed secrets) panic a bit.  Apologies.

cormacrelf:
On GKE  1.31.6-gke.1064001, GKE Warden now completely rejects the use of `system:authenticated`. No longer able to install sealed-secrets. Here's the error from Flux's helm controller.

[...]

PR #1451, opened 2024-01-29
[helm] add rbac.proxier config

alice-sawatzky:
**Description of the change**

Add options to the Helm chart to configure the `proxier` Role, and its RoleBinding.

**Benefits**

This allows the user to override the default RoleBinding that grants `system:authenticated` the `proxier` role. `system:authenticated` is not a very safe option in a lot of clusters, so it's important to expose this option.

**Possible drawbacks**

This was written to preserve existing behavior by default, so as to not break compatibility. This isn't ideal as the existing behavior is insecure on GKE (and possibly other contexts as well), but since it doesn't directly expose anything damaging (but rather is a defense-in-depth measure), it's probably OK to just highlight the importance of changing this in the documentation.

**Applicable issues**

- partially addresses #1448

**Additional information**

### DIFF
--- a/helm/sealed-secrets/templates/role-binding.yaml
+++ b/helm/sealed-secrets/templates/role-binding.yaml
@@ -40,14 +40,7 @@ roleRef:
   kind: Role
   name: {{ printf "%s-service-proxier" (include "sealed-secrets.fullname" .) }}
 subjects:
-  {{- range .Values.rbac.serviceProxier.subjects }}
-  - apiGroup: {{ .apiGroup | default "rbac.authorization.k8s.io" | quote }}
-    kind: {{ quote .kind }}
-    name: {{ quote .name }}
-    {{- with .namespace }}
-    namespace: {{ quote . }}
-    {{- end }}
-  {{- end }}
+  {{- include "sealed-secrets.render" (dict "value" .Values.rbac.serviceProxier.subjects "context" $) | nindent 2 }}
 ---
 {{ end }}
 {{ if and (and .Values.rbac.create .Values.rbac.namespacedRoles) (not $.Values.rbac.clusterRole) }}
--- a/helm/sealed-secrets/templates/role-binding.yaml
+++ b/helm/sealed-secrets/templates/role-binding.yaml
@@ -41,9 +41,9 @@ roleRef:
   name: {{ printf "%s-service-proxier" (include "sealed-secrets.fullname" .) }}
 subjects:
   {{- range .Values.rbac.proxier.subjects }}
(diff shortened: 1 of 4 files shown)


alemorcuq:
Thanks for your PR, @alice-sawatzky. We will review it later this week.

helm/sealed-secrets/templates/role-binding.yaml
```diff
@@ -38,11 +40,16 @@ roleRef:
   kind: Role
   name: {{ printf "%s-service-proxier" (include "sealed-secrets.fullname" .) }}
 subjects:
-- apiGroup: rbac.authorization.k8s.io
-  kind: Group
-  name: system:authenticated
-{{ end }}
+  {{- range .Values.rbac.proxier.subjects }}
```
alemorcuq, line 43: I would use the `sealed-secrets.render` function to render the subjects. Like this:
```
{{- include "sealed-secrets.render" (dict "value" .Values.rbac.proxier.subjects "context" $) | nindent 2 }}
```

helm/sealed-secrets/values.yaml
```diff
@@ -404,6 +404,22 @@ rbac:
   ## @param rbac.pspEnabled PodSecurityPolicy
   ##
   pspEnabled: false
+  ## "Proxier" RBAC Role configuration
+  ##
+  proxier:
```
alemorcuq, line 409: Let's change this to `serviceProxier` instead.

helm/sealed-secrets/values.yaml
```diff
@@ -404,6 +404,22 @@ rbac:
...
+    ## @param create Specifies whether to create the "proxier" role, to allow external users to access the SealedSecret API
+    ##
+    create: true
+    ## @param bind Specifies whether to create a RoleBinding for the "proxier" role
+    ##
+    bind: true
```
alemorcuq, line 415: I'm not 100% sure about this. Are you thinking of some scenario where one would want to create the role but no rolebinding for it?

alemorcuq (changes requested):
Thanks for the PR, @alice-sawatzky. Please, check my comments below.

helm/sealed-secrets/values.yaml
alice-sawatzky: say if it's used as a dependency chart, a situation could arise where it's easier to create the rolebinding in the parent chart, rather than passing values to the subchart. or maybe the subjects to bind aren't known in this context and some other helm release handles the actual binding. I don't directly need this use case, just seemed plausibly useful since delegating admin functions can get complex

alemorcuq (approved):

PR #1454, opened 2024-02-04
Offline sealedsecret validation

### LINKED ISSUE #1444: Offline validation using signatures
9 out of 10 people have problems sealing secrets correctly. Correct namespace, correct cluster and so on.
That statement is obviously an overreaction. However that said I'd like to validate SealedSecrets before they are applied to clusters.

Now I am aware that there is the `--validate` flag but that requires the ability to talk to the correct cluster. That is fine for human-to-cluster but not for validations on ci pipelines.
Speaking best practices a ci pipeline should never be able to talk to a cluster directly which kinda defeats the point of using that way of validation.

I did not exactly think that through but wouldn't we be able to add signatures to the generated SealedSecrets? If the ci or whatever knows the public key the SealedSecret could be validated if its encrypted correctly for the intended cluster and scope while not having the risk of any exposures nor the ability to decrypt it.

mohamed-essam:
**Description of the change**

1. Add a unique controller UUID that is generated on controller startup and saved in a configmap.
1. Add the ability to validate sealedsecrets offline through adding validation data preceding each sealedsecret in the form of `<base64ValidationDataJson>;<base64ActualSecret>`.
1. Add `kubeseal --validate-offline`

The JSON contains:
* Scope
* Secret Namespace
* Secret Name
* Controller UUID

**Benefits**

The ability to validate sealed secret values through offline CI.

**Possible drawbacks**

None that come to mind.

**Applicable issues**

- fixes #1444 

**Additional information**

* All changes tested on a local minikube installation.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -298,7 +298,7 @@ jobs:
 
     - name: Testing environment setup
       run: |
-        helm install sealed-secrets -n kube-system --set fullnameOverride=sealed-secrets-controller --set image.registry=$controller_registry --set image.repository=$controller_repository --set image.tag=$controller_tag --set image.pullPolicy=Never helm/sealed-secrets
+        helm install sealed-secrets -n kube-system --set enableOfflineValidation=true --set fullnameOverride=sealed-secrets-controller --set image.registry=$controller_registry --set image.repository=$controller_repository --set image.tag=$controller_tag --set image.pullPolicy=Never helm/sealed-secrets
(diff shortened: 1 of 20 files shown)


alemorcuq:
Thanks for this PR, @mohamed-essam. I appreciate the effort you put on this but, as I said on https://github.com/bitnami-labs/sealed-secrets/issues/1444#issuecomment-1934014092, we don't think this is the way forward.

PR #1457, opened 2024-02-12
[helm] Check for ServiceMonitor apiVersion capabilities

wsn-dev:
**Description of the change**

In an Kubernetes as a Service Environment the values.yaml for the sealed-secrets are predefined.
This means also that the Monitoring is enabled by default.
Enabling the ServiceMonitor and not have the kube-prometheus-stack in place leads to issues installing this chart.
The kube-prometheus-stack releates on some pre defined secrets thats why the sealed-secrets deployment is needed before the kube-prometheus-stack.
Also enabling and disabling this leads to some commits in an gitops approach way.
Thats why i would like to have this capabilities check in place. 

**Benefits**

The ServiceMonitor is not applyed unless the approappropriate crd is installed.

**Possible drawbacks**

helm template will not render the ServiceMonitor object if it does not know about this api capabilities or the --validate flag is not provided.

### DIFF
--- a/helm/sealed-secrets/Chart.yaml
+++ b/helm/sealed-secrets/Chart.yaml
@@ -14,6 +14,6 @@ maintainers:
     url: https://github.com/bitnami-labs/sealed-secrets
 name: sealed-secrets
 type: application
-version: 2.14.2
+version: 2.14.3
 sources:
   - https://github.com/bitnami-labs/sealed-secrets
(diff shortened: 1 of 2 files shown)


agarcia-oss:
Hi @wsn-dev thanks for the contribution, but we cannot merge the changes. 

These same changes were [proposed in the past](https://github.com/bitnami-labs/sealed-secrets/pull/1283#discussion_r1295594715) and we think they will cause helm template issues as you commented in the PR.

Thanks!

PR #1471, opened 2024-02-25
fix: missing common annotations in the helm chart

### LINKED ISSUE #1372: `commonAnnotations` are not implemented for all resources
The documentation for the `commonAnnotations` value says "Annotations to add to all deployed resources":

https://github.com/bitnami-labs/sealed-secrets/blob/5b0c26ea09fb2a32634eb1c935c5c28e3a44914c/helm/sealed-secrets/values.yaml#L19

Yet, not all resources have annotations at all, let alone the common annotations:

https://github.com/bitnami-labs/sealed-secrets/blob/5b0c26ea09fb2a32634eb1c935c5c28e3a44914c/helm/sealed-secrets/templates/cluster-role-binding.yaml#L4-L9

I suggest we implement `commonAnnotations` to work as documented. Depending on the circumstances, it might be good to implement it via a template in [_helpers.tpl](https://github.com/bitnami-labs/sealed-secrets/blob/main/helm/sealed-secrets/templates/_helpers.tpl) and reference that instead.

cyrgim:
**Description of the change**

Add common annotations to all resources in the helm chart as per the [documentation](https://github.com/bitnami-labs/sealed-secrets/blob/main/helm/sealed-secrets/values.yaml#L19):
`## @param commonAnnotations [object] Annotations to add to all deployed resources`

**Benefits**

Add common annotations to all resources in the helm chart as per the documentation

**Applicable issues**

- fixes #1372

### DIFF
--- a/helm/sealed-secrets/templates/configmap-dashboards.yaml
+++ b/helm/sealed-secrets/templates/configmap-dashboards.yaml
@@ -18,6 +18,9 @@ metadata:
     {{- if $.Values.metrics.dashboards.annotations }}
     {{- include "sealed-secrets.render" ( dict "value" $.Values.metrics.dashboards.annotations "context" $) | nindent 4 }}
     {{- end }}
+    {{- if $.Values.commonAnnotations }}
+    {{- include "sealed-secrets.render" ( dict "value" $.Values.commonAnnotations "context" $ ) | nindent 4 }}
+    {{- end }}
 data:
   {{ base $path }}: |-
 {{ $.Files.Get $path | indent 4 }}
--- a/helm/sealed-secrets/templates/ingress.yaml
+++ b/helm/sealed-secrets/templates/ingress.yaml
@@ -8,12 +8,13 @@ metadata:
     {{- if .Values.commonLabels }}
     {{- include "sealed-secrets.render" (dict "value" .Values.commonLabels "context" $) | nindent 4 }}
     {{- end }}
-  {{- if .Values.ingress.annotations }}
   annotations:
     {{- if .Values.ingress.annotations }}
     {{- include "sealed-secrets.render" ( dict "value" .Values.ingress.annotations "context" $) | nindent 4 }}
     {{- end }}
-  {{- end }}
+    {{- if .Values.commonAnnotations }}
+    {{- include "sealed-secrets.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+    {{- end }}
 spec:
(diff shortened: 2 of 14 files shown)


alemorcuq (changes requested):
Thanks! Could you make sure all common annotations use `sealed-secrets.render` instead of `toYaml`?

helm/sealed-secrets/templates/configmap-dashboards.yaml
```diff
@@ -18,6 +18,9 @@ metadata:
     {{- if $.Values.metrics.dashboards.annotations }}
     {{- include "sealed-secrets.render" ( dict "value" $.Values.metrics.dashboards.annotations "context" $) | nindent 4 }}
     {{- end }}
+    {{- if .Values.commonAnnotations }}
+    {{- include "sealed-secrets.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+    {{- end }}
```
alemorcuq, line 23:
```suggestion
    {{- if $.Values.commonAnnotations }}
    {{- include "sealed-secrets.render" ( dict "value" $.Values.commonAnnotations "context" $ ) | nindent 4 }}
    {{- end }}
```

helm/sealed-secrets/templates/ingress.yaml
```diff
@@ -13,6 +13,9 @@ metadata:
     {{- if .Values.ingress.annotations }}
     {{- include "sealed-secrets.render" ( dict "value" .Values.ingress.annotations "context" $) | nindent 4 }}
     {{- end }}
+    {{- if .Values.commonAnnotations }}
+    {{- include "sealed-secrets.render" ( dict "value" .Values.commonAnnotations "context" $ ) | nindent 4 }}
+    {{- end }}
```
alemorcuq, line 18: This is inside a `.Values.ingress.annotations` if clause, I think it should be out of it.

alemorcuq (changes requested):
Sorry, I missed these two in my first review. Looks good otherwise.

helm/sealed-secrets/templates/configmap-dashboards.yaml
cyrgim: oh yes, missed it.

helm/sealed-secrets/templates/ingress.yaml
cyrgim: Now removed the first clause for consistency with the other manifests, as the `.Values.ingress.annotation` was checked 2 times.

alemorcuq (approved):
Thanks!

Issue #1282, opened 2023-07-26
Controller crashed: `Observed a panic: "assignment to entry in nil map"`


lindhe:
**Which component**:

controller version 0.23.0

**Describe the bug**

The only thing I know is that the controller crashed and the log looked like this:

<details><summary><code>kubectl -n kube-system logs sealed-secrets-controller-854f5fbffc-9g5xw</code></summary>
<p>

```
controller version: 0.23.0
2023/07/26 10:44:41 Starting sealed-secrets controller version: 0.23.0
Searching for existing private keys
----- sealed-secrets-keygkj4b
HTTP server serving on :8080
Updating trident/backend-ontap-nas
Updating argocd/cluster-config-repo
Event(v1.ObjectReference{Kind:"SealedSecret", Namespace:"trident", Name:"backend-ontap-nas", UID:"8012bc39-e03c-492e-8e62-b9c42d18c74a", APIVersion:"bitnami.com/v1alpha1", ResourceVersion:"31182844", FieldPath:""}): type: 'Normal' reason: 'Unsealed' SealedSecret unsealed successfully
Updating argocd/argocd-notifications-secret
Event(v1.ObjectReference{Kind:"SealedSecret", Namespace:"argocd", Name:"cluster-config-repo", UID:"c2f90054-82e2-42a5-bf07-b886967b396a", APIVersion:"bitnami.com/v1alpha1", ResourceVersion:"37273431", FieldPath:""}): type: 'Normal' reason: 'Unsealed' SealedSecret unsealed successfully
E0726 10:44:41.883408       1 runtime.go:79] Observed a panic: "assignment to entry in nil map" (assignment to entry in nil map)
goroutine 71 [running]:
k8s.io/apimachinery/pkg/util/runtime.logPanic({0x17ff680?, 0x1c87950})
        /home/runner/go/pkg/mod/k8s.io/apimachinery@v0.27.3/pkg/util/runtime/runtime.go:75 +0x99
k8s.io/apimachinery/pkg/util/runtime.HandleCrash({0x0, 0x0, 0xc0003ce020?})
        /home/runner/go/pkg/mod/k8s.io/apimachinery@v0.27.3/pkg/util/runtime/runtime.go:49 +0x75
panic({0x17ff680, 0x1c87950})
        /opt/hostedtoolcache/go/1.20.3/x64/src/runtime/panic.go:884 +0x213
[...]

PR #1474, opened 2024-02-29
fix: panic when patching empty secret

alemorcuq:
When patching an empty secret, i.e. one with no data field, the controller will panic as it's trying to assign to a nil map. Same will happen with the labels.

- Always initialize the Data and Labels maps if they are empty
- Add a test to cover this use case

fixes: #1282


agarcia-oss (approved):
Great fix, @alemorcuq 👏

PR #1519, opened 2024-04-29
fix: show field name in error message when base64 decoding fails

mxey:
**Description of the change**

unsealing errors usually show the field name that caused the error - except for base64 decoding. This change fixes that.

**Benefits**

clearer error message

**Possible drawbacks**

none

**Applicable issues**

**Additional information**

### DIFF
--- a/pkg/apis/sealedsecrets/v1alpha1/sealedsecret_expansion.go
+++ b/pkg/apis/sealedsecrets/v1alpha1/sealedsecret_expansion.go
@@ -277,7 +277,8 @@ func (s *SealedSecret) Unseal(codecs runtimeserializer.CodecFactory, privKeys ma
 		for key, value := range s.Spec.EncryptedData {
 			valueBytes, err := base64.StdEncoding.DecodeString(value)
 			if err != nil {
-				return nil, err
+				errs = append(errs, multierror.Tag(key, err))
+				continue
 			}
 			plaintext, err := crypto.HybridDecrypt(rand.Reader, privKeys, valueBytes, label)
 			if err != nil {
--- a/pkg/apis/sealedsecrets/v1alpha1/sealedsecret_test.go
+++ b/pkg/apis/sealedsecrets/v1alpha1/sealedsecret_test.go
@@ -298,7 +298,7 @@ func TestSealRoundTripWithMisMatchClusterWide(t *testing.T) {
 
 	_, err := ssecret.Unseal(codecs, keys)
 	if err == nil {
-		t.Fatalf("Unseal did not return expected error: %v", err)
+		t.Fatal("Expecting error: got nil instead")
 	}
 }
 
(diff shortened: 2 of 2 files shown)


pkg/apis/sealedsecrets/v1alpha1/sealedsecret_expansion.go
```diff
@@ -277,7 +277,7 @@ func (s *SealedSecret) Unseal(codecs runtimeserializer.CodecFactory, privKeys ma
 		for key, value := range s.Spec.EncryptedData {
 			valueBytes, err := base64.StdEncoding.DecodeString(value)
 			if err != nil {
-				return nil, err
+				errs = append(errs, multierror.Tag(key, err))
```
alemorcuq, line 280: If there's an error here we should continue to the next item.

```suggestion
				errs = append(errs, multierror.Tag(key, err))
				continue
```

pkg/apis/sealedsecrets/v1alpha1/sealedsecret_test.go
```diff
@@ -572,6 +572,33 @@ func TestRejectBothEncryptedDataAndDeprecatedV1Data(t *testing.T) {
...
+	codecs := serializer.NewCodecFactory(scheme)
+	_, keys := generateTestKey(t, testRand(), 2048)
+
+	_, err := sealedSecret.Unseal(codecs, keys)
+	if err == nil {
+		t.Fatalf("Unseal did not return expected error: %v", err)
+	}
+
+	if !strings.Contains(err.Error(), "foo") {
+		t.Errorf("Unseal error did not contain field name %q: %v", "foo", err)
```
alemorcuq, line 598: For consistency with the other tests:

```suggestion
		t.Errorf("Expecting error: %v to contain field %q", err, "foo")
```

pkg/apis/sealedsecrets/v1alpha1/sealedsecret_test.go
```diff
@@ -572,6 +572,33 @@ func TestRejectBothEncryptedDataAndDeprecatedV1Data(t *testing.T) {
...
+		},
+	}
+
+	scheme := runtime.NewScheme()
+	codecs := serializer.NewCodecFactory(scheme)
+	_, keys := generateTestKey(t, testRand(), 2048)
+
+	_, err := sealedSecret.Unseal(codecs, keys)
+	if err == nil {
+		t.Fatalf("Unseal did not return expected error: %v", err)
```
alemorcuq, line 594: For consistency with the other tests:

```suggestion
		t.Fatalf("Expecting error: got nil instead")
```

alemorcuq (changes requested):
Thanks for the PR. I've left some comments.

pkg/apis/sealedsecrets/v1alpha1/sealedsecret_expansion.go
mxey: Good catch. I added another assertion to ensure that invalid base64 does not trigger a decrypt error.

pkg/apis/sealedsecrets/v1alpha1/sealedsecret_test.go
mxey: changed

pkg/apis/sealedsecrets/v1alpha1/sealedsecret_test.go
mxey: changed

alemorcuq (approved):

PR #1528, opened 2024-05-17
[helm] Set `GOMAXPROCS` and `GOMEMLIMIT` environment variables

jnoordsij:
**Description of the change**

Set `GOMAXPROCS` and `GOMEMLIMIT` environment variables based on container resources.

Inspired by https://github.com/traefik/traefik-helm-chart/pull/1029.

**Benefits**

This should reduce potential CPU throttling and OOMKills on containers.

**Possible drawbacks**

This creates an empty `env` key for those not setting resource values. This is only a little ugly, but should not be harmful. Alternatively, we could add some conditional wrapper around the whole `env` block to only make it appear if a value is set, but that will be more complicated if additional env would be added in the future.

**Applicable issues**

N/A

**Additional information**

The [`resourceFieldRef`](https://kubernetes.io/docs/concepts/workloads/pods/downward-api/#downwardapi-resourceFieldRef) is a very specific Kubernetes directive that is created specifically for passing resource-related values, which rounds up the CPU value to the nearest whole number (e.g. 250m to 1) and passes the memory as a numeric value; so `64Mi` would result in the environment variable being set to `67108864`. This by design makes it completely compatible with Go's API.

An example is documented within Kubernetes documentation itself: https://kubernetes.io/docs/tasks/inject-data-application/environment-variable-expose-pod-information/#use-container-fields-as-values-for-environment-variables.

### DIFF
--- a/helm/sealed-secrets/templates/deployment.yaml
+++ b/helm/sealed-secrets/templates/deployment.yaml
@@ -106,7 +106,7 @@ spec:
             {{- end }}
             - --privatekey-annotations
             - {{ trimSuffix "," $privatekeyAnnotations | quote }}
-            {{- end }}            
+            {{- end }}
             {{- if $.Values.privateKeyLabels }}
             {{- $privateKeyLabels := ""}}
             {{- range $k, $v := $.Values.privateKeyLabels }}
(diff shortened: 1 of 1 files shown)


alemorcuq (approved):
Great, thank you!

PR #1536, opened 2024-05-23
Fix code generation

alemorcuq:
**Description of the change**
Code generation wasn't working. This PR fixes it and updates the generated code:

- Use latest version of `controller-gen` to generate the manifests for the CRD
- Use `kube_codegen.sh` to generate the client code. The `generate_groups.sh` script we were using previously has been deprecated and removed.
- Code generation is now done from a `hack` directory following the [Kubernetes sample-controller example](https://github.com/kubernetes/sample-controller/blob/master/hack/update-codegen.sh).
- `make manifests` has been run to update the manifests.
- `make generate` has been run to update the generated code.


alvneiayu (approved):
LGTM

agarcia-oss (approved):

agarcia-oss (approved):

PR #1542, opened 2024-06-07
Add sprig function library for templating

### LINKED ISSUE #1193: Sprig support in templates
**Which component**:
Controller.

Derive password hash (bcrypt) from password (e.g. for htpasswd).

Add sprig functions to templates.

I do not know any alternatives in sealed-secrets. An alternative could be to use external-secrets (https://github.com/external-secrets/external-secrets/blob/d605a5b450abb8072090ef115396f8b618aec2ec/pkg/template/v2/template.go#L138).

**Additional context**
No.

### LINKED ISSUE #1534: Templating support for escaping JSON/YAML values
**Which component**: controller

We use yaml for our application configuration files. These config files end up containing some secrets (such as db passwords), so we use `template.data` (as [documented here][ss templating example]) to inject these secrets.

The problem is that these passwords sometimes contain [indicator characters] such as `"` and `'` which have special meaning to YAML and cannot be trivially injected into a YAML string without escaping ([the documented example][ss templating example] would break too).

Here's the [relevant code in sealed secrets], it just uses go's `text/template` under the hood. Unfortunately, AFAICT, `text/template` does not have a proper way of doing this. Notably, [js/JSEscape does not work].

Would y'all be open to making the [Sprig function library][Sprig] available here? It has a [toJson function] that would work well for generating both JSON and YAML. This is what ArgoCD provides for templating, see [ArgoCD - Go Template] for details.

I'd be happy to send in a PR adding Sprig if you're open to it.

- We could avoid generating passwords with YAML [indicator characters].
- Separate our application configuration from our secrets, as [recommended here][split config from secrets].

[ss templating example]: https://github.com/bitnami-labs/sealed-secrets/blob/v0.26.2/docs/examples/config-template/sealedsecret.yaml#L25
[indicator characters]: https://yaml.org/spec/1.2.2/#53-indicator-characters
[relevant code in sealed secrets]: https://github.com/bitnami-labs/sealed-secrets/blob/v0.26.2/pkg/apis/sealedsecrets/v1alpha1/sealedsecret_expansion.go#L297
[js/JSEscape does not work]: https://github.com/golang/go/issues/37634
[Sprig]: https://masterminds.github.io/sprig/
[toJson function]: https://masterminds.github.io/sprig/defaults.html#tojson-musttojson
[split config from secrets]: https://github.com/bitnami-labs/sealed-secrets/tree/v0.26.2/docs/examples/config-template
[ArgoCD - Go Template]: https://argo-cd.readthedocs.io/en/stable/operator-manual/applicationset/GoTemplate/

jfly:
This fixes https://github.com/bitnami-labs/sealed-secrets/issues/1534 and https://github.com/bitnami-labs/sealed-secrets/issues/1193.

**Description of the change**

Add sprig function library for templating

**Benefits**

As described in https://github.com/bitnami-labs/sealed-secrets/issues/1534, this lets people safely generate more types of configuration (such as yaml files) and safely and correctly escape passwords.

**Possible drawbacks**

I have not removed `env`, `expandenv`, and `getHostByName`, [as argocd does](https://github.com/argoproj/argo-cd/blob/v2.11.3/applicationset/utils/utils.go#L31-L34). I supposed there's some security concern here? Happy to filter them out if folks like.

**Applicable issues**

- fixes #1534
- fixes #1193

**Additional information**

Thanks for the consideration!


alemorcuq (changes requested):
Thanks @jfly ! Could you please run `go mod tidy`? I'm pretty sure `sprig` should not be an indirect dependency.

jfly:
Oh yikes. Sorry, I am a complete go noob. sprig should definitely not be an indirect dependency. Fixed!

docs/examples/config-template/sealedsecret.yaml
```diff
@@ -22,15 +22,15 @@ spec:
                                "port": "22",
                                "env": "SOME_ENV",
                                "user": "myuser",
-                               "password": "{{ index . "server1" }}",
+                               "password": {{ toJson .server1 }},
```
jfly, line 25: I snuck in an unrelated change here (`index . "server1"` to `.server1`). Happy to back it out if folks prefer. I prefer it because it avoids parens:

```
"password": "{{ toJson (index . "server1") }}",
```

(but, total go noob, not sure if there are other considerations)

alemorcuq (approved):

alemorcuq:
Thanks @jfly !

PR #1547, opened 2024-06-13
Remove code setting name/namespace overriding secret templates

### LINKED ISSUE #1543: template.metadata.name ignored in sealed secret
**Which component**:
sealed secret controller: docker.io/bitnami/sealed-secrets-controller:0.26.3

**Describe the bug**

When creating a sealed secret with a setting of spec.template.metadata.name different from the name of the sealed secret, a secret is created with name equal to metadata.name of the sealed secret. The metadata.name from the template section is ignored. 

**To Reproduce**

Create a sealed secret. 
Add an override for metadata.name in the template section. Example: 

```
kind: SealedSecret
metadata:
  name: passwordsecret-ss
spec:
  encryptedData:
    ARTIFACTORY_PASSWORD: YOUR_ENCRYPTED_SECRET_HERE
  template:
    metadata:
      name: passwordsecret
```
Now a secret named 'passwordsecret-ss' is created, not the expected secret 'passwordsecret'.

**Expected behavior**
In the above example, a secret with name 'passwordsecret' should be created. 

**Version of Kubernetes**:
kubectl version:
```
Client Version: v1.29.1
Server Version: v1.27.4+k3s1
WARNING: version difference between client (1.29) and server (1.27) exceeds the supported minor version skew of +/-1
```

**Additional context**

tarikdem:
**Description of the change**

Currently, when specifying a template in a `SealedSecret` resource, `metadata.name` and `metadata.namespace` are ignored and the name and namespace from the source metadata is applied.

**Benefits**

Possibility to override the name and namespace of the generated `Secret` resource.

**Possible drawbacks**

The code that's overriding `metadata.name` and `metadata.namespace` is added before the templating functionality so I might be missing potential implications of this change. One thing that might be impactful is the fact that overriding the namespace might **enable somebody to apply a secret to a namespace they don't have access to**. If that is the case, the CRD should be updated to remove the `namespace` field from `spec.template.metadata` altogether.

**Applicable issues**

- fixes #1543 

**Additional information**

### DIFF
--- a/pkg/apis/sealedsecrets/v1alpha1/sealedsecret_expansion.go
+++ b/pkg/apis/sealedsecrets/v1alpha1/sealedsecret_expansion.go
@@ -324,10 +324,6 @@ func (s *SealedSecret) Unseal(codecs runtimeserializer.CodecFactory, privKeys ma
 		return nil, fmt.Errorf("using deprecated 'data' field, use 'encryptedData' or flip the feature flag")
 	}
 
-	// Ensure these are set to what we expect
-	secret.SetNamespace(smeta.GetNamespace())
-	secret.SetName(smeta.GetName())
-
 	gvk := s.GetObjectKind().GroupVersionKind()
 	if anno, ok := s.Spec.Template.Annotations[SealedSecretSkipSetOwnerReferencesAnnotation]; !ok || anno != "true" {
 		// Refer back to owning SealedSecret


tarikdem:
Closing this as I found the answer in the docs:

> Note the SealedSecret and Secret must have the same namespace and name. This is a feature to prevent other users on the same cluster from re-using your sealed secrets. See the [Scopes](https://github.com/bitnami-labs/sealed-secrets?tab=readme-ov-file#scopes) section for more info.

PR #1579, opened 2024-08-14
Label "app.kubernetes.io/instance" in the condition_info Prometheus metric

tewfik-ghariani:
**Description of the change**

A new label called `ss_app_kubernetes_io_instance` is added to the **sealed_secrets_controller_condition_info** Prometheus metric

This label shows the `app.kubernetes.io/instance` label value of the SealedSecret resource itself, if it exists

e.g
```
kind: SealedSecret
metadata:
  labels:
    app.kubernetes.io/instance: grafana-qa
```

Here is an example of the newly added metric label:
Screenshot 2024-08-14 at 15 53 43

I checked the `/metrics` endpoint and the metric label is considered empty if it did not exist, thus it is not breaking anything

**Benefits**

The scraped information of the Sealed Secret resources include an indicator of the instance name / release name associated to that specific resource

Users relying on Grafana may conduct more complex queries to join and map the instance name with other queries

**Possible drawbacks**

_None_

**Applicable issues**

_None_

**Additional information**
The usage of the `app.kubernetes.io/instance` label is the standard as it is part of the k8s recommended labels: https://kubernetes.io/docs/concepts/overview/working-with-objects/common-labels/

### DIFF
--- a/pkg/controller/metrics.go
+++ b/pkg/controller/metrics.go
@@ -17,6 +17,7 @@ const (
 	labelNamespace = "namespace"
 	labelName      = "name"
 	labelCondition = "condition"
+	labelInstance  = "ss_app_kubernetes_io_instance"
 )
 
 var conditionStatusToGaugeValue = map[v1.ConditionStatus]float64{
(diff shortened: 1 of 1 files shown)


tewfik-ghariani:
Hi, is it possible to review this PR? cc @agarcia-oss ^^

tewfik-ghariani:
not stale

tewfik-ghariani:
Re-open?

PR #1647, opened 2024-11-27
Added SHA256 digest support to the helm deployment template 

devopstoday11:
**Description of the change**

Add SHA256 digest support to the deployment template and add the "digest" parameter to values.yaml 

**Benefits**

It is easy to pull images with specific SHA digest values when the need arises. 

**Possible drawbacks**

We used it in our production environment for a while and haven't seen any drawbacks yet. 

**Additional information**

The deployment template update to support the SHA digest helped us explicitly specify the SHA digest for particular architecture-based images when facing issues in the cluster that are irrelevant to sealed secrets and image pull. Having the SHA digest support helped us not have a roadblock to deploying and using sealed secrets smoothly. So, I am sharing the update we did and used with the community.

### DIFF
--- a/helm/sealed-secrets/values.yaml
+++ b/helm/sealed-secrets/values.yaml
@@ -33,6 +33,7 @@ commonLabels: {}
 ## @param image.registry Sealed Secrets image registry
 ## @param image.repository Sealed Secrets image repository
 ## @param image.tag Sealed Secrets image tag (immutable tags are recommended)
+## @param image.digest Sealed Secrets image SHA256 Digest value 
 ## @param image.pullPolicy Sealed Secrets image pull policy
 ## @param image.pullSecrets [array]  Sealed Secrets image pull secrets
 ##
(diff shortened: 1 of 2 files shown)


helm/sealed-secrets/values.yaml
```diff
@@ -33,13 +33,15 @@ commonLabels: {}
...
 ## @param image.tag Sealed Secrets image tag (immutable tags are recommended)
+## @param image.digest Sealed Secrets image SHA256 Digest value 
 ## @param image.pullPolicy Sealed Secrets image pull policy
 ## @param image.pullSecrets [array]  Sealed Secrets image pull secrets
 ##
 image:
   registry: docker.io
   repository: bitnami/sealed-secrets-controller
   tag: 0.27.2
+  #digest: ""
```
alvneiayu, line 44: Please uncomment this, no need to comment it. Just define it with the empty ""

alvneiayu (changes requested):
could you include it also in the README tables, please?

PR #1655, opened 2024-12-06
fix: explicitly set resourceFieldRef.divisor

Jaydee94:
**Description of the change**

In the helm chart the divisor is not set and kubernetes mutates the Deployments to "0" which is the same as "1".

**Benefits**

When deploying with ArgoCD the application will not be displayed as "OutOfSync"

**Possible drawbacks**

---


Jaydee94:
Hey, can you consider to accept this change to the helm chart?

This would help us so that our Argocd does not show out of sync.

Best Regards :)

agarcia-oss (approved):

Issue #1639, opened 2024-11-20
Bring-your-own-certificate - controller does not fetch latest key imported


sybernatus:
**Which component**:
The name (and version) of the affected component (controller or kubeseal)
- kubeseal: v0.27.1
- controller: v0.26.3

**Describe the bug**
We followed the documentation [bring-your-own-certificate](https://github.com/bitnami-labs/sealed-secrets/blob/main/docs/bring-your-own-certificates.md)

While importing our own key on the controller under the `test-key` secret name, the new key is taken into account by the controller once we restarted it:
```shell
time=2024-11-20T13:52:11.023Z level=INFO msg="registered private key" secretname=test-key
time=2024-11-20T13:52:11.024Z level=INFO msg="registered private key" secretname=sealed-secrets-keymqbn8
time=2024-11-20T13:52:11.024Z level=INFO msg="registered private key" secretname=sealed-secrets-keyqm6bb
```

Even if this is the newest key (`kubectl get secret`):
```shell
sealed-secrets-keymqbn8   kubernetes.io/tls          2      10d
sealed-secrets-keyqm6bb   kubernetes.io/tls          2      10d
test-key                  kubernetes.io/tls          2      13m
```

The controller continue to retrieve an old public key using `--fetch-cert` argument. It seems to be randomly taken at each controller restart.
This is causing re-encrypting our secrets because there is no way to tell kubeseal to always use the latest key to re-encrypt the sealed secrets.

**To Reproduce**
Steps to reproduce the behaviour:
- Follow the procedure [bring-your-own-certificate](https://github.com/bitnami-labs/sealed-secrets/blob/main/docs/bring-your-own-certificates.md) until `Deleting the controller Pod is needed to pick the new keys`
- launch `kubeseal --fetch-cert` and compare it to the public key in the newly created secret.

**Expected behavior**
- the controller should retrieve only the latest public certificate (as it is the most relevant certificate to encrypt secret)

**Version of Kubernetes**:

- Output of `kubectl version`:

```
v1.28.10
```

**Additional context**

Today our workaround is:
- to get public key from our secret deployed
- to get the name of a sealed secret on the cluster & the secret name inside
- to get the secret associated using the name
- to seal this secret using the public key

alvneiayu:
hi @sybernatus 

[...]

PR #1681, opened 2025-01-22
Fix register a key using secret creationTimestamp instead of certificate validity timestamp

sybernatus:
**Description of the change**

Register a key and order it based on its secret creation timestamp instead of its cert starting validity (NotBefore attribute)

**Benefits**

In case we are bringing our own certificate, if the certificate is created before sealed secret installation but applied after the installation, still the latest certificate used will be the certificate installed by sealed secret during its initialization instead of the latest secret certificate created.

**Possible drawbacks**

**Applicable issues**

- fixes # https://github.com/bitnami-labs/sealed-secrets/issues/1639

**Additional information**


alvneiayu:
hi @sybernatus 

First of all, thanks a lot for your PR.

This PR will change the way to order the certificates for all the people with Sealed Secrets deployed. This impact is really high if we include it as default.

I suggest to include it with a flag to configure what is the way that the users want to order the certificates. If we see in the future that everybody is using the flag, we can deprecated it and set it by default but I think that we need to be conservative here.
I though something like --order-by-creation-timestamp.

For this case, we need to include a test and documentation.

Thanks a lot

Álvaro

sybernatus:
I will try to include this flag in this pr

sybernatus:
Up please 🙂

agarcia-oss (approved):
LGTM!

alvneiayu (approved):
LGTM, thanks a lot for your incredible work @sybernatus

PR #1703, opened 2025-03-14
Fix potential controller sensitive data exposure by sprig template functions

RaveNoX:
**Description of the change**

Fixing potential controller environment exposure by removing some sprig template functions:
- `env`, `expandenv` - controller environment variables exposure (which can contain sensitive data)
- `getHostByName` - add possibility to resolve dns entries from templates, which is not secure in some cases

This potential exposure is added by PR #1542 where security risks was described in "possible drawbacks" section.

Also, sprig func map is now initialized due module initialization and reusing for rendering, which may improve performance.

In fact is just port of the [argocd](https://github.com/argoproj/argo-cd/blob/v2.11.3/applicationset/utils/utils.go#L31-L34) solution, which described in initial PR #1542

**Benefits**

- Fix potential controller sensitive data exposure
- Improve rendering performance

**Possible drawbacks**

No one

**Applicable issues**

**Additional information**


alvneiayu:
thanks a lot for your work @RaveNoX

PR #1737, opened 2025-05-28
add watch for secrets

farodin91:
**Description of the change**

Watch for new key secrets.

**Benefits**

We manage our secrets external their we have to restart sealed secrets always a new cert is created.

**Possible drawbacks**

**Applicable issues**

- fixes #

**Additional information**


alvneiayu:
well, I have a question about it. Could you provide me more information about your scenario. In the past, include a new watcher increases the CPU and memory of the product and it is important to know it and why we want to include it. Which tool are you using and the complete scenario, please?

farodin91:
We create certificates externally for multiple cluster and rotate them periodically. We had issues the new certificates weren't loaded.

PR #1867, opened 2026-01-14
use my namespace as key namespace

### LINKED ISSUE #1862: Controller watches for seal key secrets in all namespaces
**Which component**:
controller

**Describe the bug**
After upgrading our deployment to `v0.34.0`, it seems the controller is watching secrets in all namespaces to register as keys. In older releases, it would only watch for secrets in the namespace the controller is deployed in. This is problematic for our setup as we use 2 sealed secret controllers in different namespaces, both with their own managed set of unseal keys. New sealed secrets are now encrypted with the wrong seal key.

**To Reproduce**

1. Create 2 namespaces
2. Create a (different) secret in both namespaces and label them with `sealedsecrets.bitnami.com/sealed-secrets-key` so they're picked up the controller as seal keys
3. Deploy sealed-secrets in 1 of the namespaces
4. Notice that both secrets are registered as seal keys

**Expected behavior**
As stated in the [documentation](https://github.com/bitnami-labs/sealed-secrets?tab=readme-ov-file#manual-key-management-advanced), the controller should only look for seal key secrets in its own namespace.

**Version of Kubernetes**:

```
Client Version: 4.17.38
Server Version: 4.18.30
Kubernetes Version: v1.31.14
```

**Additional context**
After some digging around, I think it is related to this PR: https://github.com/bitnami-labs/sealed-secrets/pull/1831. In that PR, the logic to manage seal keys is refactored and there is a change in the way secrets are retrieved. In the old logic (`pkg/controller/main.go:128`), there is a filter on secrets which takes the namespace into consideration. The new logic (`pkg/controller/controller.go:138`) seems to be missing the namespace predicate.

After rolling back to the previous version (`v0.33.1`), only the secrets in the controller namespace are considered seal keys.

farodin91:
Fixes #1862

### DIFF
--- a/pkg/controller/controller_test.go
+++ b/pkg/controller/controller_test.go
@@ -120,12 +120,13 @@ func TestConvert2SealedSecretPassThrough(t *testing.T) {
 
 func TestDefaultConfigDoesNotSkipRecreate(t *testing.T) {
 	ns := "some-namespace"
+	keyNs := "some-key-namespace"
 	var tweakopts func(*metav1.ListOptions)
-	clientset := fake.NewSimpleClientset()
+	clientset := fake.NewClientset()
 	ssc := ssfake.NewSimpleClientset()
 	keyRegistry := testKeyRegister(t, context.Background(), clientset, ns)
 
-	got, err := prepareController(clientset, ns, tweakopts, &Flags{SkipRecreate: false}, ssc, keyRegistry)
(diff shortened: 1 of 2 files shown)


rickbrouwer (approved):

farodin91:
@alvneiayu Would you like to review?

PR #1871, opened 2026-01-30
Enable encryptedData to be omitted by setting key to null in template

### LINKED ISSUE #1870: Selectively omit encryptedData keys when using templates
**Which component**:
controller

I would like to selectively omit encryptedData keys from the final secret when those keys are only needed as inputs to templates.

Set the matching key to `null` in `spec.template.data` to indicate it should be omitted from the final secret.

* Adding an additional field to the spec with a list of keys to omit (or include?).
* Adding an all-or-nothing field to include or exclude the entirety of encryptedData in the final secret

**Additional context**
The motivating use case is a tool that expects a directory of configuration files, where some, but not all, of the fields in those files are sensitive. The tool in question will refuse to run if a file in that directory cannot be parsed.

The ideal case is to use a SealedSecret with templates to encrypt only the sensitive data, leave the rest of the configuration in plaintext, and mount the resulting secret as a directory in a pod. Currently, this results in the individual sensitive fields being interpreted as configuration files and being rejected. This feature would permit this use case by omitting each of the encrypted fields from the final secret.

meln5674:
**Description of the change**

Make `SealedSecret.spec.template.data` fields nullable. A null value indicates that the key should be omitted from the final secret.

Change is backwards compatible.

**Benefits**

Templated Secrets can be used in cases where the set of keys is significant, i.e. extraneous keys are not acceptable.

**Possible drawbacks**

None known.

**Applicable issues**

- fixes https://github.com/bitnami-labs/sealed-secrets/issues/1870

**Additional information**

### DIFF
--- a/helm/sealed-secrets/crds/bitnami.com_sealedsecrets.yaml
+++ b/helm/sealed-secrets/crds/bitnami.com_sealedsecrets.yaml
@@ -2,7 +2,7 @@ apiVersion: apiextensions.k8s.io/v1
 kind: CustomResourceDefinition
 metadata:
   annotations:
-    controller-gen.kubebuilder.io/version: v0.15.0
+    controller-gen.kubebuilder.io/version: v0.20.1
   name: sealedsecrets.bitnami.com
 spec:
   group: bitnami.com
@@ -69,7 +69,10 @@ spec:
                   data:
                     additionalProperties:
                       type: string
-                    description: Keys that should be templated using decrypted data.
+                    description: |-
+                      Keys that should be templated using decrypted data.
+                      Set a key with the same name as in `encryptedData`
+                      to `null` to omit the unsealed data from the final secret.
                     nullable: true
                     type: object
                   immutable:
--- a/pkg/apis/sealedsecrets/v1alpha1/sealedsecret_expansion.go
+++ b/pkg/apis/sealedsecrets/v1alpha1/sealedsecret_expansion.go
@@ -337,15 +337,19 @@ func (s *SealedSecret) Unseal(codecs runtimeserializer.CodecFactory, privKeys ma
 		// a real secret value.
 		// See https://github.com/bitnami-labs/sealed-secrets/issues/1607
 		for key, value := range s.Spec.Template.Data {
-			if _, exists := data[key]; !exists {
-				data[key] = value
+			if _, exists := data[key]; !exists && value != nil {
+				data[key] = *value
 			}
 		}
 
 		for key, value := range s.Spec.Template.Data {
 			var plaintext bytes.Buffer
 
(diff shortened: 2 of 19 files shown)


meln5674:
@alvneiayu It would appear the import and tooling versions have diverged.

3 of the 4 lint findings are deprecated functions called in files I did not modify, but instead the functions themselves are deprecated in the new `make generate` output, which suggests a version of k8s.io/code-generator other than the one listed in the go.mod was used to generate the current versions. The version annotation on the generated CRD went from 0.15.0 to 0.20.0, which is further evidence of version drift. I also notice the version of `sigs.k8s.io/controller-tools/cmd/controller-gen` in the Makefile is not pinned to a specific version, and will not have reproducible outputs.

I should also mention, `make generate` on its own fails as it tries to use a package (`cmd/validation-gen`) not actually present in that version of that module. I had to use `make lint CODEGEN_PKG="$GOPATH/pkg/mod/k8s.io/code-generator@v0.35.1"` to get it to run at all.

According to the git history, the version of that module the last time the zz_deepcopy.go file was changed was 0.30.1, but resetting the go.mod version to that results in other errors.

I have fixed the one lint finding that was introduced by one of my changes and also ran `make manifests` which I neglected to do the first time, but beyond that, let me know how you would like me to proceed.

(As a piece of totally unsolicited advice, from one operator maintainer to another, putting needed tool versions in the go.mod via `go get -tool`, available since go 1.24, makes problems like this go away, as does running `make generate` and `make manifests` in CI and verifying the outputs have not changed)

meln5674:
@alvneiayu I lost track of this, but after looking back into this, I identified the issue and submitted a fix [here](https://github.com/bitnami/sealed-secrets/pull/2020). I believe after this it should work, as it passes both code generation and tests locally.

alvneiayu:
thanks @meln5674 

I will evaluate it again, again thanks for your contribution and for the fixes.

Álvaro

alvneiayu:
Hi @meln5674 

Heads up: #2019 just merged and changes Spec.Template.Data compatibility. After rebasing, these map[string]string literals will fail to compile against your new map[string]*string type:

- sealedsecret_test.go — TestValidateEncryptedDataIgnoresTemplate (2 spots)
- controller_test.go — TestAttemptUnsealIgnoresTemplate (1 spot)

Fix with your someStr() helper, e.g.:

ssecret.Spec.Template.Data = map[string]*string{
    "probe": someStr(`{{ fail "attacker-controlled failure" }}`),
}

No logic conflict — just a type fixup.

meln5674:
Rebased and updated tests, which are passing locally.

alvneiayu (approved):

PR #1876, opened 2026-02-11
Bump Go to 1.26.0

jnoordsij:
**Description of the change**

Update to latest Go version (1.26.0). See also https://go.dev/blog/go1.26.

**Benefits**

**Possible drawbacks**

**Applicable issues**

- fixes #

**Additional information**


agarcia-oss (commented):
We tend to be a bit more conservative with the golang updates. Normally we do not upgrade to a new golang minor version right after release.
In this case, I think we better stick to 1.25.x for a bit.

jnoordsij:
All fine, feel free to keep this open till later. Just for what it's worth, Go releases by design are already somewhat more batteproof on stable release than other software (see https://go.dev/wiki/Go-Release-Cycle#august--february-week-2-release-issued).

PR #1905, opened 2026-03-02
fix: add mutex locking to KeyRegistry to prevent data races

### LINKED ISSUE #1904: Data race in KeyRegistry - mutex only held by getCert()
***Which component:***
Controller (sealed-secrets-controller), all versions                                                                                                                   
                                                                                                                              
***Describe the bug***
KeyRegistry embeds a sync.Mutex but only getCert() actually holds it. Several other methods read or write the shared keys map and mostRecentKey pointer without synchronization:

| Method                  | Lock held? | Access           |
| ------------------ | ---- |----------------------------------|
| getCert()                | Yes | read mostRecentKey                       |
| registerNewKey()  | No   | write keys map + mostRecentKey │
| generateKey()        | No   | calls registerNewKey()                  │
| latestPrivateKey() | No   | readMostRecentKey                       |
| initKeyRenewal      | No.  | read keys + mostRecentKey         |

Because the background key-rotation goroutine calls generateKey() → registerNewKey() concurrently with HTTP handlers calling getCert() and latestPrivateKey(), this is a data race on the keys map and the mostRecentKey pointer.

To Reproduce
1. Deploy sealed-secrets-controller with a short key renewal  interval (e.g. --key-renew-period=1m)
2. Run go test -race ./pkg/controller/... or send concurrent requests to /v1/cert.pem and /v1/rotate while a key rotation is in progress
3. The Go race detector reports a data race on KeyRegistry.keys and KeyRegistry.mostRecentKey

***Expected behavior***
All reads and writes to KeyRegistry.keys and KeyRegistry.mostRecentKey should be protected by the mutex. registerNewKey(), generateKey(), latestPrivateKey(), and any direct field access in initKeyRenewal() should hold the lock (or be provably single-threaded) when touching shared state.

smuda:
**Description of the change**

KeyRegistry embedded sync.Mutex but only getCert() held it. Background key rotation goroutines write keys/mostRecentKey while HTTP handlers read them concurrently, causing data races. Add mutex acquisition to registerNewKey, latestPrivateKey, and new accessor methods (privateKeys, keyLen, mostRecentKeyTime). Replace direct field access in controller and main with thread-safe accessors.

**Benefits**

No data race.

**Possible drawbacks**

It's possible to end up with lower throughput, which is the reason I changed mutex to RWMutex.
The embedded sync.Mutex was promoted, meaning external code could call kr.Lock() directly. The named mu field removes that.

**Applicable issues**

- fixes #1904

### DIFF
--- a/pkg/controller/controller.go
+++ b/pkg/controller/controller.go
@@ -2,7 +2,6 @@ package controller
 
 import (
 	"context"
-	"crypto/rsa"
 	"encoding/json"
 	"errors"
 	"fmt"
@@ -579,7 +578,10 @@ func (c *Controller) Rotate(content []byte) ([]byte, error) {
 		if err != nil {
 			return nil, fmt.Errorf("error decrypting secret. %v", err)
 		}
-		latestPrivKey := c.keyRegistry.latestPrivateKey()
+		latestPrivKey, err := c.keyRegistry.latestPrivateKey()
+		if err != nil {
+			return nil, fmt.Errorf("error getting latest private key. %v", err)
+		}
 		resealedSecret, err := ssv1alpha1.NewSealedSecret(scheme.Codecs, &latestPrivKey.PublicKey, secret)
 		if err != nil {
 			return nil, fmt.Errorf("error creating new sealed secret. %v", err)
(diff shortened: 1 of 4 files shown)


smuda:
@alvneiayu Of course I can reopen the PR! I just thought that you deemed it uninteresting but didn't want to say so. I do have a problem with managing my PRs and don't want to lose track of them, being unresponsive to the maintainers.

alvneiayu:
hi @smuda 

Just to say, really amazing job. Thanks a lot for your contribution and for the investigation behind this contribution.

Álvaro

alvneiayu (approved):

PR #1967, opened 2026-05-21
Fix oci push

alvneiayu:


agarcia-oss (approved):

MarV1908 (commented):
I think this needs to be reverted in order to fix the OCI artifact push on helm registry.

PR #1978, opened 2026-05-29
Allow setting hostUsers on deployment

jnoordsij:
**Description of the change**

As of Kubernetes 1.36, the User Namespaces feature has graduated to stable; see also https://kubernetes.io/docs/concepts/workloads/pods/user-namespaces/.

To leverage this on pods, one needs to set the hostUsers flag on a deployment. This MR adds the ability to add such a flag to this chart. See also https://kubernetes.io/docs/tasks/configure-pod-container/user-namespaces/.

**Benefits**

Allowing deployment to run with user namespaces, increasing (theoretical) security level.

**Possible drawbacks**

None

**Applicable issues**

n/a

**Additional information**

n/a

### DIFF
--- a/helm/sealed-secrets/templates/deployment.yaml
+++ b/helm/sealed-secrets/templates/deployment.yaml
@@ -54,6 +54,9 @@ spec:
       securityContext: {{- omit .Values.podSecurityContext "enabled" | toYaml | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ include "sealed-secrets.serviceAccountName" . }}
+      {{- if kindIs "bool" .Values.hostUsers }}
+      hostUsers: {{ .Values.hostUsers }}
+      {{- end }}
       {{- if .Values.hostNetwork }}
       hostNetwork: true
       {{- end }}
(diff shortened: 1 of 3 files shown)


helm/sealed-secrets/templates/deployment.yaml
```diff
@@ -54,6 +54,9 @@ spec:
       securityContext: {{- omit .Values.podSecurityContext "enabled" | toYaml | nindent 8 }}
       {{- end }}
       serviceAccountName: {{ include "sealed-secrets.serviceAccountName" . }}
+      {{- if kindIs "bool" .Values.hostUsers }}
+      hostUsers: {{ .Values.hostUsers }}
```
alvneiayu, line 58: is not included in the values.yaml. Please include it also in the README.

alvneiayu (commented):
First of all, thanks a lot for your contribution.

I reviewed the PR, could you take it a look the comment, please?

Thanks again

Álvaro

helm/sealed-secrets/templates/deployment.yaml
jnoordsij: I've added it to both places; let me know if you have any other ideas about where to put it and how.

jnoordsij:
Thanks for the review; I've adjusted to include a proposed (default) value in the values file and add the flag to the README.

jnoordsij:
not stale

alvneiayu (approved):

PR #2017, opened 2026-08-05
kubeseal: apply --namespace when unsealing secrets without one

david0:
**Description of the change**

With --recovery-unseal, a sealed secret file that doesn't declare a namespace can now be decrypted by supplying --namespace. The flag value is used to rebuild the encryption label, so users no longer have to edit the file to add a namespace before recovering it.

**Benefits**

This is helpful because the output of `helm template` does not include a namespace by default, so being able to pass --namespace avoids having to edit the rendered manifests before recovery.

**Possible drawbacks**

It could be that the user specifies a wrong namespace or forgets to use this parameter altogether, but thats not worse than before.

**Applicable issues**

**Additional information**


alvneiayu (approved):

alvneiayu (commented):
Integration tests are broken

PR #2018, opened 2026-08-10
fix(controller): start HTTP early for large additional-namespaces lists

### LINKED ISSUE #1959: Startup is slow with large number of additional namespaces
We have observed that with a large number of namespaces given to `--additional-namespaces`, the time of the controller is quite long. Currently it takes ~1s per namespace to start up. We have ~300 namespaces given to `--additional-namespaces` so startup takes ~5min.

There are a number of problems that arise because of the large startup:
- The HTTP server is only started after all namespace watches have been started. This means that liveness probes will fail during this time even though the controller is making progress.
- Startup probes can help alleviate the problem above. However, in our case we dynamically add namespaces to `--additional-namespaces` so we need to restart the controller to update the namespaces list.
- Due to the large startup time the controller lags in responsiveness to `SealedSecret` and `Secret` change events.

A few improvements come to mind:
- Improve the startup time of watches per namespace.
- Move start up of the HTTP server (at least for probes) earlier to before the namespace watch creation loop. Liveness probes can succeed, and Readiness probes can fail during this time. After the watches are started, the readiness probe can return successfully.

Related:

- #697

VedantMadane:
**Description of the change**

With a large `--additional-namespaces` list the controller spent a long time in a serial loop before it ever bound the HTTP listener. Liveness probes that hit `/healthz` therefore failed even while startup was making progress.

This change:

1. Starts the HTTP and metrics servers right after the home-namespace controller begins so `/healthz` is available during bootstrap.
2. Adds `/readyz` which returns 503 until additional-namespace informers have been started then 200.
3. Points the chart readiness probe at `/readyz` (liveness stays on `/healthz`).
4. Starts additional-namespace validation and informer setup with a concurrency limit of 16 so wall-clock time is not one API RTT per namespace.

**Benefits**

- Pods with many additional namespaces keep liveness green while informers start.
- Readiness stays false until those informers are started so traffic waits for a useful controller.
- Faster bootstrap when the list is large.

**Possible drawbacks**

- Chart readiness now uses `/readyz`. Custom probes that expected readiness on `/healthz` should switch path or use `customReadinessProbe`.
- Bootstrap concurrency is capped at 16. Extremely slow API servers may still take a while but no longer block probes.

**Applicable issues**

- fixes #1959

**Additional information**

Unit coverage for `/readyz` is in `TestHttpReadyz`. `go test ./pkg/controller` passes locally.

### DIFF
--- a/helm/sealed-secrets/templates/deployment.yaml
+++ b/helm/sealed-secrets/templates/deployment.yaml
@@ -216,7 +216,7 @@ spec:
           {{- if .Values.readinessProbe.enabled }}
           readinessProbe: {{- include "sealed-secrets.render" (dict "value" (omit .Values.readinessProbe "enabled") "context" $) | nindent 12 }}
             httpGet:
-              path: /readyz
+              path: /healthz
               port: http
           {{- else if .Values.customReadinessProbe }}
           readinessProbe: {{- include "sealed-secrets.render" (dict "value" .Values.customReadinessProbe "context" $) | nindent 12 }}
--- a/helm/sealed-secrets/templates/deployment.yaml
+++ b/helm/sealed-secrets/templates/deployment.yaml
@@ -216,7 +216,7 @@ spec:
           {{- if .Values.readinessProbe.enabled }}
           readinessProbe: {{- include "sealed-secrets.render" (dict "value" (omit .Values.readinessProbe "enabled") "context" $) | nindent 12 }}
             httpGet:
-              path: /healthz
+              path: /readyz
               port: http
           {{- else if .Values.customReadinessProbe }}
           readinessProbe: {{- include "sealed-secrets.render" (dict "value" .Values.customReadinessProbe "context" $) | nindent 12 }}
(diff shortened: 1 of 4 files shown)


helm/sealed-secrets/templates/deployment.yaml
```diff
@@ -216,7 +216,7 @@ spec:
           {{- if .Values.readinessProbe.enabled }}
           readinessProbe: {{- include "sealed-secrets.render" (dict "value" (omit .Values.readinessProbe "enabled") "context" $) | nindent 12 }}
             httpGet:
-              path: /healthz
+              path: /readyz
```
alvneiayu: Please, this is unrelated to the addition-namespace lists. Keep it as healthz, please

alvneiayu (commented):
Thanks a lot for your contribution. I included a comment.

Thanks again

Álvaro

alvneiayu (dismissed):
LGTM

alvneiayu (commented):
Linter errors 🥲

alvneiayu (approved):
LGTM

